PR #22, opened 2024-10-02
feat(document-model-libs): remove wrapped editor to prevent loss of state

ryanwolhuter:
Dynamically creating the editor component we use in the story like this:

`const WrappedEditor = wrapEditor(Editor);`

Causes the story to destroy and rebuild the component every time the args change. This makes the component state reset when actions are dispatched, causing the storybook behavior not match the behavior in connect.

We can remove this function and achieve the same result by updating the definition of the editor component story type and the args we pass to it.

There was also an existing type error for the RWA editor because it needed other custom args. I have updated the createDocumentStory function to allow passing these other args. I have defined them as specific types because these are likely to be used by other editors.

### DIFF
--- a/packages/document-model-libs/editors/rwa/rwa.stories.tsx
+++ b/packages/document-model-libs/editors/rwa/rwa.stories.tsx
@@ -1,11 +1,23 @@
-import { createDocumentStory } from 'document-model-libs/utils';
-import { reducer, utils } from '../../document-models/real-world-assets';
+import {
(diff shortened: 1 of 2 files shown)


gpuente (approved):
LGTM!

For now you have to trigger the release manually for this package, using the  `Release Package Manual` github action:

Screenshot 2024-10-02 at 11 34 23 AM

PR #368, opened 2024-10-04
feat(design-system): rwa editor props and modals logic

gpuente:
### The EditorContext

This context takes in the state from the document model and makes a mirror of the state to be kept internally. This state is kept in sync with the document model state, but this synchronization can be paused so as to avoid changing data that is actively being edited by the user.

The editor state from the document model is kept in a ref, so that callbacks returned by the context that depend on it can use the latest value without needing to include the state in the dependency array.

Along with the document state, the context also expects a function that I have chosen to call a "dispatcher". 

`export type EditorDispatcher = <TEditorAction extends EditorAction>(
    action: TEditorAction,
) => ActionOutputFor<TEditorAction>;
`

This function is similar to the reducer, except it only accepts an action as an argument. This action has a `type` which maps to the type of a reducer action, but the payload is the form inputs for a given data type. This dispatcher calls the document-model code which calls the reducer's `dispatch` function after doing the appropriate transformations on the data. If this `dispatch` call is successful, the updated item is returned to the "dispatcher", which in turn returns this value to the ui so that it can update optimistically while the state synchronizes.

This context also now manages the table state via a reducer. This state indicates which table, table item and operation are selected by the user.

### The Editor component

I have encapsulated all of the editor logic into a single component which can be imported by document-model-libs.

We want to be able to test the behavior of the RWA editor in its entirety in the storybook so that we can tell when renders etc. are happening because of the components vs the document model itself.

I have removed the separate stories for the various tables and replaced them with one story for the whole editor. 

These stories have a `simulateBackgroundUpdates` arg that can be toggled for testing the editor's behavior in the situation where the document model state is changing in the background.

### Tables

Previously there were 5 different table components. Now the ideal case is to use the `TableWithForm` component, which handles everything and only needs the `tableName` prop.

`<TableWithForm tableName={tableNames.SPV} />`

For example.

In the places where more custom behavior is needed, this component takes props to update the rows. The `GroupTransactionsTable` is the only table that does not use this component yet, because it also has filtering logic/components which still need to be made general. I intend to do this soon, so that these filtering components can be used by all the tables.

Converting the document model data into the table data is handled by the `handleTableData` hook. This takes the `tableName` and finds the appropriate items in the state and maps them. 

### Forms

The inputs for the forms that mutate the document model are unfortunately not an exact match to the data structures in the model. These are also different from the shape of the table data. This requires different typings for the values passed to `react-hook-form`. I have now set up type maps that use the `tableName` to find the appropriate type, which alleviates the need for generics.

The form logic including the invocation of `useForm`, getting appropriate default values for edit vs create, and generating the form input components are now handled in these hooks:

`useTableForm` - returns all the props to render the form
`useSubmit` - calls react-hook-form and returns the functions to submit data
`useFormInputs` - returns input components based on the `tableName` of the form

The `useTableForm` hook receives an `operation` prop of create, view or edit. This is intentionally not always the same as the `operation` of the `EditorContext` itself, because we also need to use this hook in the modals. These modals are shown from the forms, so the `EditorContext` might be in a state of `edit` for editing an asset, but then the user opens a modal from that form which needs to call the hook with the `create` operation.

### Generics

Since there were so many different tables with different fields and types of data, there were a ton of generics all over the place. This was made even worse by the tables needing to show different fields / data than what comes directly from the document model.

With this new state approach, `TableName` type is used to index all of the other data types. This means that all we need to do is pass the `tableName` argument, and everything else can be inferred from there. The vast majority of generic components no longer need to be generic.

### Modals

We had a problem where modals were being closed after submitting their data. This happened because the modals were children of the form components and their data changes on data submission. This is already solved by the new state handling, but for good measure I also added a ModalManager which works in the same way as in Connect. This makes the open/close state of these modals independent of the parent components.

With the new generalized state system we no longer need individual modals for the different tables/forms. We only need one for delete and one for create.


acaldas (approved):
Copying my comment from the original PR: https://github.com/powerhouse-inc/design-system/pull/380#pullrequestreview-2347921928

As we discussed, it's hard for me to fully appreciate the changes as I don't have enough context on the specifics of the rwa editor.
I really like the separation of concerns from the business logic to the UI. Looking forward to see how the insights flow upstream to a general editor component/container.

`EditorContext` seems to be an improved version of the component in Connect that calls the specific document model editors.
https://github.com/powerhouse-inc/connect/blob/main/src/components/editors.tsx#L81

It uses the `useDocumentDispatch` hook to clone the document state for the UI.
https://github.com/powerhouse-inc/connect/blob/b2800d5eabd7af61150e6896e4f6eef64784c955/src/utils/document-model.ts#L43

packages/design-system/src/rwa/components/modal/modal-manager.tsx
```diff
@@ -0,0 +1,70 @@
+import { ModalPropsMapping, modals, ModalType } from '@/rwa';
```
acaldas, line 1: In the future, we should think about Connect providing the editor component a way to show a modal. It would keep the UX more consistent as modals can de done in a million different ways.

PR #370, opened 2024-10-07
feat(scalars): setup scalars project

gpuente:
## Description

- setup scalars project


.eslintignore
```diff
@@ -2,4 +2,6 @@ node_modules
 dist
 build
 storybook-static
-.nx
\ No newline at end of file
+.nx
+vite.config.ts
```
acaldas: Is this needed? If it's an esm issue we should set `"type": "module"` in package.json or rename the vite config to `.mts`.

.github/workflows/check-pr-scalars.yml
```diff
@@ -0,0 +1,77 @@
...
+        types: [opened, synchronize]
+
+env:
+  NX_ISOLATE_PLUGINS: false
+
+jobs:
+    lint:
+        name: Lint
+        runs-on: ubuntu-latest
+        container: node:22
```
acaldas, line 22: Adding a comment here but it's not for this PR, I've created a separate issue: https://github.com/powerhouse-inc/powerhouse/issues/371

Each job is running pnpm install in it's own docker instance. And I guess now this runs in all the packages of the monorepo right? We need to be careful in the CI actions to do as less work as possible.

Pnpm has a way to cache the dependencies as we have used in some actions: https://pnpm.io/continuous-integration#github-actions
Here we use `container: node.22` so each jobs runs in its own docker container AFAIK. This way the pnpm cache is not shared between jobs and `nx` is not able to optimize the jobs with cache.

I think we should still use docker actions, as we have add some issues with node APIs not being available in some jobs.

I don't know if there is an alternative approach to this, but using a cache action would already take us far: https://github.com/actions/cache

One idea here is to have a `setup` job that installs and caches the dependencies, using a checksum of `pnpm-lock` as cache key and caching `pnpm store path`.

Then all the other jobs run after it in parallel and load from the cache.

Maybe we could also share the nx cache?

.github/workflows/check-pr-scalars.yml
```diff
@@ -0,0 +1,77 @@
...
+            - uses: actions/checkout@v4
+
+            - name: Install PNPM
+              uses: pnpm/action-setup@v3
+              with:
+                  version: latest-8
+                  run_install: false
+
+            - name: Install Dependencies
+              run: pnpm install --no-frozen-lockfile
```
acaldas: I think it makes sense to use `--frozen-lockfile` as it's enabled by default in CI. Any reason not to?
It's important to enforce the lockfiles to be updated.

packages/scalars/.eslintrc
```diff
@@ -0,0 +1,54 @@
...
+    ],
+    "parser": "@typescript-eslint/parser",
+    "parserOptions": {
+        "ecmaVersion": "latest",
+        "sourceType": "module",
+        "tsconfigRootDir": "."
+    },
+    "plugins": ["@typescript-eslint"],
+    "rules": {
+        "@typescript-eslint/explicit-function-return-type": "off",
```
acaldas, line 21: Would be cool to create our own rule list so we can just extend it in each project.

packages/scalars/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    },
+    "include": [
+        "src/**/*",
+        "test/**/*",
+        "*.config.ts",
+        "*.config.js",
+        "reset.d.ts",
+        "modules.d.ts",
+    ],
+    "exclude": ["vite.config.ts", "vite.config.ts"]
```
acaldas: Same here, would be good to have linting on `vite.config`

packages/scalars/src/scalars/EmailAddress.ts
```diff
@@ -0,0 +1,45 @@
...
+        throw new GraphQLError(`Value is not string: ${JSON.stringify(value)}`);
+    }
+
+    const result = emailSchema.safeParse(value);
+
+    if (result.success) return result.data;
+    throw new GraphQLError(result.error.message);
+};
+
+export const GraphQLEmailAddressConfig: GraphQLScalarTypeConfig<
```
acaldas: I think we should not use specific names for the exports, it makes loading the scalars dynamically harder.

Each scalar could export:
- `schema` (currently `emailSchema`)
- `config` (currently `GraphQLEmailAddressConfig`)
- `scalar`(currently `GraphQLEmailAddress`)
- `typedef` (currently `EmailAddressTypeDefinition`)

Would be nice to define a `ScalarModule` type for this.

packages/scalars/src/scalars/index.ts
```diff
@@ -0,0 +1,7 @@
+import { GraphQLEmailAddress } from './EmailAddress';
+
+export * from './EmailAddress';
```
acaldas: Here it would be just `export * as EmailAddress from './EmailAddress"`.

packages/scalars/src/typeDefs.ts
```diff
@@ -0,0 +1,3 @@
+export const EmailAddressTypeDefinition = 'scalar EmailAddress';
```
acaldas: If the type definition is exported by each scalar then I don't think this file is needed.

packages/scalars/README.md
```diff
@@ -0,0 +1,78 @@
...
+    - Navigate to the `src/scalars` directory.
+    - Create a new file for your scalar, e.g., `MyScalar.ts`.
+
+2. **Define the Scalar**:
+
+    - Implement the scalar using the `GraphQLScalarType` from `graphql` library or custom logic.
+    - Ensure you export the scalar properly.
+
+    ```typescript
+    import { GraphQLScalarType, Kind } from 'graphql';
```
acaldas: Let's update this with the required exports after the other comments.
Let's use named exports only. (No export default)

packages/scalars/vite.config.ts
```diff
@@ -0,0 +1,46 @@
+import { defineConfig } from 'vite';
+import dts from 'vite-plugin-dts';
+import generateFile from 'vite-plugin-generate-file';
+
+const entry = {
+    index: 'src/index.ts',
```
acaldas: Would be nice to dynamically generate an entry for each scalar and also add it as an export in the generated package.json files.
Then it would be possible to do `import { schema, config } from "@powerhousedao/scalars/email-address".

packages/scalars/src/scalars/index.ts
acaldas: In the future we could create a `nx generate scalar EmailAddress` that creates the boilerplate and updates this file.

acaldas (changes requested):
Very nice! I added a few comments but let's try to discuss the handover of this project on today's call.

One thing that would be helpful, and help to test our setup, would be to show how to adapt a scalar from the `graphql-scalars` lib: https://the-guild.dev/graphql/scalars/docs/scalars/account-number
So we could have a AccountNumber scalar that maps the exports from that lib to ours, and adds a zod schema.

gpuente:
Hey @acaldas!

I've addressed the comments you left on this PR.

I added an auto-generated "exports" field to the `package.json` when the scalars project is built. Additionally, I created an Nx generator for the creation of new scalars: `npx nx generate scalar-generator:new-scalar`

tools/scalar-generator/project.json
```diff
@@ -0,0 +1,51 @@
...
+            }
+        },
+        "lint": {
+            "executor": "@nx/eslint:lint"
+        },
+        "test": {
+            "executor": "@nx/jest:jest",
+            "outputs": ["{workspaceRoot}/coverage/{projectRoot}"],
+            "options": {
+                "jestConfig": "tools/scalar-generator/jest.config.ts"
```
acaldas, line 47: I don't see this `jest.config.ts` file, was it not pushed?
Any reason for jest or can it use vitest instead?

tools/scalar-generator/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "scalar-generator",
+    "version": "0.0.1",
+    "dependencies": {
+        "@nx/devkit": "20.0.0",
+        "tslib": "^2.3.0"
+    },
+    "type": "commonjs",
```
acaldas: Let's use `"type": "module"` for all the packages.

packages/scalars/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "typescript": "^5.3.3",
+    "vite": "^5.0.11",
+    "vite-plugin-dts": "^3.7.1",
+    "vite-plugin-generate-file": "^0.1.1",
+    "vitest": "^2.0.5"
+  },
+  "dependencies": {
+    "graphql": "^16.9.0",
+    "zod": "^3.23.8"
+  },
```
acaldas, line 39: Graphql and zod should be peerDependencies instead, and excluded from the bundle.

They also need to be added as devDependencies to work locally.

packages/scalars/src/scalars/EmailAddress.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type ScalarType = {
+    input: string;
+    output: string;
+};
+
+export const typedef = 'scalar EmailAddress';
+
+export const schema = z.string().email();
+
```
acaldas, line 19: Screenshot 2024-10-10 at 13 21 03

Each scalar should also export a TS type in string form. Maybe `export const type`?

packages/scalars/src/scalars/index.ts
```diff
@@ -0,0 +1,20 @@
+// namespace imports -- DO NOT REMOVE OR EDIT THIS COMMENT
```
acaldas, line 1: Let's keep it like this but this could also be generated when bundling.

tools/scalar-generator/project.json
gpuente: this was added automatically by nx, I've removed this, we are not testing the generators for now

tools/scalar-generator/package.json
gpuente: ready!

packages/scalars/package.json
gpuente: done!

packages/scalars/src/scalars/EmailAddress.ts
gpuente: added!

packages/scalars/src/scalars/index.ts
gpuente: I'm using this comments as reference for the generator, so it can update the index file properly

packages/scalars/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "files": [
+    "dist"
+  ],
+  "types": "dist/types/src/index.d.ts",
+  "main": "dist/cjs/src/index.js",
+  "module": "dist/es/src/index.js",
+  "type": "module",
+  "scripts": {
+    "check-types": "tsc",
+    "lint": "eslint . && npm run check-types",
```
acaldas: Let's add `"postlint": "npm run check-types"` instead of having it here. This way it's still possible to pass arguments to eslint

packages/scalars/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "lint:fix": "eslint --fix . && npm run check-types",
```
acaldas: Same here

acaldas (approved):

PR #373, opened 2024-10-09
Use unified configs

ryanwolhuter:


acaldas:
Bliss. Can you update the design snapshots so the test job passes?

acaldas (approved):

PR #387, opened 2024-10-10
Smaller bundler and improved support for tree shaking

acaldas:
Unzipped bundle went from 3mb to 300kb.

- Removed dependencies from the bundle so that they are only added on an app if the components that need them are used.
- Replaced connect loader with mp4 video.
- Icons can be loaded individually


packages/design-system/scripts/create-icon-components.ts
```diff
@@ -1,15 +1,25 @@
...
+  readdir,
+  readFileSync,
+  writeFileSync,
+} from "node:fs";
+import { join } from "node:path";
 
-const iconsDir = path.join("public/icons");
-const outputFilePath = path.join("src/assets/icon-components.tsx");
+const iconsDir = join("public/icons");
+const outputDirPath = join("src/assets/icon-components");
```
acaldas, line 12: Each icon is written to a separate file to allow lazy loading a single icon

packages/design-system/scripts/create-icon-components.ts
```diff
@@ -22,26 +32,45 @@ fs.readdir(iconsDir, (err, files) => {
...
-    const svgData = fs.readFileSync(filePath, "utf8");
-    const svgDataWithProps = svgData.replace("<svg", "<svg {...props}");
-    outputContent += `export function ${componentName}(props: Props) {\n`;
-    outputContent += `    return (\n${svgDataWithProps}\n    );\n`;
-    outputContent += `}\n\n`;
+    const svgData = readFileSync(filePath, "utf8");
+    const svgDataWithProps = svgData
+      .replace("<svg", "<svg {...props}")
+      // turn props into camelCase
+      .replace(
```
acaldas, line 44: Some props where getting lost when converting from svg to jsx.

packages/design-system/src/connect/components/search-bar/search-bar.test.tsx
```diff
@@ -25,15 +30,19 @@ const filterItems: ConnectSearchBarProps["filterItems"] = [
...
         filterItems={filterItems}
         filterLabel="File type"
         onChange={() => {}}
         placeholder="Search Files"
         value="test"
-      />,
+      />
+    );
+
+    await waitForElementToBeRemoved(() =>
```
acaldas, line 44: Snapshots that have icons need to way for them be loaded. Maybe there is a more general way to do this in storybook?

packages/design-system/src/connect/utils/index.ts
```diff
@@ -1,4 +1,3 @@
 export * from "./address";
 export * from "./get-document-icon-src";
 export * from "./get-folder-status";
-export * from "./mocks";
```
acaldas, line 4: Removed mocks from barrel files so they don't get included in the bundle.

packages/design-system/src/powerhouse/components/icon/icon.tsx
```diff
@@ -1,20 +1,61 @@
...
 };
 
+function IconErrorFallback(props: Props) {
+  return <div style={{ width: props.width, height: props.height }} />;
+}
+
+function loadIcon(name: IconName): ElementType {
+  try {
+    return lazy<ComponentType<Props>>(
+      () => import(`@/assets/icon-components/${name}.tsx`),
```
acaldas, line 28: This will lazy load each individual icon.

packages/design-system/src/powerhouse/components/icon/icon.tsx
```diff
@@ -1,20 +1,61 @@
...
-  const IconComponent = iconComponents[name];
-  return <IconComponent {...props} style={_style} />;
+
+  const IconComponent = useMemo(() => loadIcon(name), [name]);
+
+  return (
+    // displays div with the same size while icon
+    // loads to avoid UI displacement
+    <Suspense
+      fallback={<div data-testid="icon-fallback" style={dimensions} />}
```
acaldas, line 54: While the icon is loading, a div with the same size is displayed so it takes up the same space in the UI.

packages/design-system/vite.config.ts
```diff
@@ -26,7 +46,41 @@ export default defineConfig(() => {
     build: {
       sourcemap: true,
       rollupOptions: {
-        external: [...Object.keys(pkg.peerDependencies), "react/jsx-runtime"],
+        output: {
+          entryFileNames: "[name].[hash].js",
+          chunkFileNames: "[name].[hash].js",
+          assetFileNames: "[name].[hash][extname]",
+          preserveModules: true,
```
acaldas, line 53: Preserves each file as a separate chunk.

packages/design-system/vite.config.ts
```diff
@@ -26,7 +46,41 @@ export default defineConfig(() => {
...
+        },
+        external: (id) => {
```
acaldas, line 55: Excludes all dependencies from the bundle and throws an error if a devDependency is accessed from the bundle.

gpuente (approved):

PR #432, opened 2024-10-25
feat(scalars): initial implementation of Toggle 

alejandrocabriales:


packages/design-system/src/scalars/components/toggle/toggle.tsx
```diff
@@ -1,28 +1,53 @@
...
+      onChange = () => {},
+      ...props
+    },
+    ref,
+  ) => (
+    <SwitchPrimitives.Root
+      required={required}
+      checked={checked}
+      role="switch"
+      name="switch"
```
YasielCabrera: the name should be a prop instead a hardcoded value

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/toggle/toggle.tsx
alejandrocabriales: you are right, done

YasielCabrera (approved):

nesdc88 (approved):
LGTM

PR #437, opened 2024-10-28
feat(scalars): add Radio & RadioGroup components (with stories & tests)

nesdc88:
## Ticket
https://trello.com/c/WSUj6qol/714-4-enum-radiobutton-component

## Description
Add Radio & RadioGroup components (with stories & tests)

## What solved
- [X] Add Radio component
- [X] Add RadioGroup component


packages/design-system/src/scalars/components/radio-group/radio/radio.stories.tsx
```diff
@@ -0,0 +1,263 @@
...
+    label: { control: "text" },
+    value: { control: "text" },
+  },
+  args: {
+    label: "Default",
+    value: "default",
+  },
+};
+
+export const DefaultChecked: Story = {
```
alejandrocabriales, line 39: It's possible to moving the shared argTypes (like label, value, checked, description, and readOnly) to the meta level instead of defining them in each story? This would simplify the code and reduce repetition. What do you think?

packages/design-system/src/scalars/components/radio-group/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useId } from "react";
+import * as RadioGroupPrimitive from "@radix-ui/react-radio-group";
+import { FormLabel } from "@/scalars/components/form-label";
+import { cn } from "@/scalars/lib/utils";
+
+export interface RadioGroupProps
+  extends React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Root> {
+  "aria-label"?: string;
```
YasielCabrera: if ` RadioGroupPrimitive.Root` doesn't include this, we should inherit from the HTML input element and make this out the box. There's this param, but could be dozen more like event handlers, etc and we should not handle each one of them via props this way unless there's no other way to do it

packages/design-system/src/scalars/components/radio-group/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,79 @@
...
+  children: React.ReactElement;
+  className?: string;
+  defaultValue?: string;
+  description?: string;
+  hasError?: boolean;
```
YasielCabrera: We should add an `error` props (it's a requirement), from that prop you can define if there's errors or not

packages/design-system/src/scalars/components/radio-group/radio/radio.stories.tsx
nesdc88: For me this makes sense in the case of "label" and "value" which are common in all variants, but the other controls are used depending on the variant. WYF guys? @alejandrocabriales @YasielCabrera

packages/design-system/src/scalars/components/radio-group/radio/radio.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { cn } from "@/scalars/lib/utils";
+
+export interface RadioProps
+  extends React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Item> {
+  className?: string;
+  description?: string;
+  hasError?: boolean;
+  id?: string;
+  label: string;
+  readOnly?: boolean;
```
YasielCabrera: we should handle readonly from design as disabled, as it is the correct state in the HTML standard (readonly exist and it is valid but it is not the correct state for this one)

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/radio-group/radio/radio.stories.tsx
YasielCabrera: IMHO all the possible params (args) that a component could have should be in the meta argTypes instead in specific stories. Note that this is just the arg type declaration to let storybook know what controls they should add in the UI no the actual value of the component.

nesdc88: argTypes moved to the top (meta) level

packages/design-system/src/scalars/components/radio-group/radio-group/radio-group.tsx
nesdc88: Removed ( RadioGroupPrimitive.Root has aria-label support)

packages/design-system/src/scalars/components/radio-group/radio-group/radio-group.tsx
nesdc88: Done

packages/design-system/src/scalars/components/radio-group/radio/radio.tsx
nesdc88: Done

packages/design-system/src/scalars/components/radio-group/radio/radio.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  it("should match snapshot", () => {
+    const { asFragment } = render(
+      <RadioGroup>
+        <Radio label="Test Label" value="test" />
+      </RadioGroup>,
+    );
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it("renders with label and value", () => {
```
YasielCabrera: by convention, each test should start with the word "should". Eg: "should render with label and value"

packages/design-system/src/scalars/components/radio-group/radio/radio.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  it("generates unique id when not provided", () => {
+    render(
+      <RadioGroup>
+        <Radio label="Test Label" value="test" />
+      </RadioGroup>,
+    );
+    const radio = screen.getByRole("radio");
+    expect(radio.id).toMatch(/^.*-radio$/); // Check if id ends with -radio
+  });
+});
```
YasielCabrera, line 104: a few tests proposals:
- should change the value when a different radio is selected
- should allow no to have any radio selected
- should select only one radio at the time
- should call on change when any radio is clicked
- should select the radio when the label is clicked

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/radio-group/radio/radio.test.tsx
nesdc88: Done

packages/design-system/src/scalars/components/radio-group/radio/radio.test.tsx
nesdc88: More test cases added

YasielCabrera (approved):

PR #443, opened 2024-10-28
feat(document-model-libs): add initial state editor

ryanwolhuter:
Users want to be able to define and edit the initial state of document models. This is especially important for backwards compatibility.

What makes this annoying is that the initial state must conform to the shape of the graphql schema, and this code had to be written by hand.

I have added here JSON editors that start with an empty object, but have a "sync with schema" button which will update the json object such that it conforms to the schema with the minimal values. This function also is aware of existing json code, and will not overwrite values that are already correct and conforming, it will just add missing ones.

This editor also has syntax highlighting and basic linting.

I wanted to wire up the existing editor's validation logic, but I discovered that it is actually not working. In the implementation of the editor, we ignore the validator that is passed in completely:

```tsxasync function generateSchema(code: string, name: string) {
    // using callbacks instead of await due to rollup error
    codegen(code).then((result) => {
      import(/* @vite-ignore */ result).then((validators) => {
        const schemaName = `${name}StateSchema`;
        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
        const validator = validators[schemaName];
        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
        
        /* validation schema is set here */
        
        setValidationSchema(validator);
        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
        
        /* and passed to the onGenerate function here */
        
        onGenerate({ documentName: name, schema: code, validator });
      });
    });
  }
```

` ``tsx
<EditorSchema
          height={200}
          name={name}
          
          {/* but in the usage of onGenerate in the component, the validator is not used */
          
          onGenerate={(schema) => {
            setSchemaState((_state) => ({
              ..._state,
              ...schema,
            }));
            if (schema.schema !== specification?.state[scope].schema) {
              setStateSchema(schema.schema, scope);
            }
          }}
          scope={scope}
          theme={theme}
          value={specification?.state[scope].schema}
        />
```
        
        Considering that shipping the zod package adds so much to our bundle size, I think it would be worthwhile to pursue a more lightweight solution.
        
        I believe that JSON schemas present one such solution: https://json-schema.org/
        
        There are several libraries that convert graphql schemas to json schemas. we could then use something like this:  https://ajv.js.org/ to generate the  `Diagnostic` objects used by codemirror and eslint https://codemirror.net/docs/ref/#lint

### DIFF
--- a/packages/document-model-libs/editors/document-model-2/components/index.ts
+++ b/packages/document-model-libs/editors/document-model-2/components/index.ts
@@ -3,3 +3,4 @@ export * from "./label";
 export * from "./graphql-editor";
 export * from "./module-form";
 export * from "./input";
+export * from "./json-editor";
--- a/packages/document-model-libs/editors/document-model-2/components/json-editor.tsx
+++ b/packages/document-model-libs/editors/document-model-2/components/json-editor.tsx
@@ -1,19 +1,13 @@
 import { Annotation, EditorState } from "@codemirror/state";
 import { EditorView, ViewUpdate } from "@codemirror/view";
 import { oneDark } from "@codemirror/theme-one-dark";
-import { graphql, updateSchema } from "cm6-graphql";
 import { useEffect, useRef } from "react";
-import { GraphQLSchema, parse } from "graphql";
 import { basicSetup } from "codemirror";
-import {
-  validateGraphQlDocuments,
-  createDefaultRules,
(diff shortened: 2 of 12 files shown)


gpuente (approved):
LGTM!

PR #448, opened 2024-10-29
feat(scalars): add styles for toogle and fix the darkmode

alejandrocabriales:
## Ticket
https://trello.com/c/uY0oAAr1/713-2-boolean-toggle-component

## What solved
- [X] Default status without a label at the right.
- [X] Default status with a label at the right. 
- [X] Off status without a label. 
- [X] Off status with a label.
- [X] Off status with a label. 
- [X] Read only On status without a label.
- [X] Read only On status with a label at the right.
- [X] Read only On status with a label at the right. 
- [X] Read only Off status with a label at the right. 
- [X] Read only Off status without a label.

### DIFF
--- a/packages/design-system/src/globals.css
+++ b/packages/design-system/src/globals.css
@@ -247,17 +247,3 @@ h6,
 p {
   color: var(--gray-900);
 }
-
-/* Light Theme */
-.light {
-  --background-color: #ffffff; /* Background */
-  --text-color: #333333; /* Text */
-  --primary-color: #1d4ed8; /* Primary color */
-}
-
-/* Dark Theme */
-.dark {
-  --background-color: #1a1a2e; /* Background */
-  --text-color: #f5f5f5; /* Text */
-  --primary-color: #3b82f6; /* Primary color */
-}
--- a/packages/design-system/src/globals.css
+++ b/packages/design-system/src/globals.css
@@ -16,9 +16,6 @@
     -ms-overflow-style: none; /* IE and Edge */
     scrollbar-width: none; /* Firefox */
   }
-  body {
-		@apply dark:bg-black  min-h-screen relative;
-	}
   /* Modal */
   .modal-shadow {
     box-shadow:
--- a/packages/design-system/src/globals.css
+++ b/packages/design-system/src/globals.css
@@ -16,7 +16,9 @@
     -ms-overflow-style: none; /* IE and Edge */
     scrollbar-width: none; /* Firefox */
   }
-
+  body {
+		@apply dark:bg-black  min-h-screen relative;
+	}
   /* Modal */
   .modal-shadow {
     box-shadow:
@@ -244,4 +246,18 @@ h5,
 h6,
 p {
   color: var(--gray-900);
+}
+
+/* Light Theme */
+.light {
+  --background-color: #ffffff; /* Background */
+  --text-color: #333333; /* Text */
+  --primary-color: #1d4ed8; /* Primary color */
+}
+
+/* Dark Theme */
+.dark {
+  --background-color: #1a1a2e; /* Background */
+  --text-color: #f5f5f5; /* Text */
+  --primary-color: #3b82f6; /* Primary color */
 }
--- a/packages/design-system/src/scalars/components/toggle/toggle-field.test.tsx
+++ b/packages/design-system/src/scalars/components/toggle/toggle-field.test.tsx
@@ -1,0 +1,75 @@
+import React from "react";
+import { render, screen, fireEvent } from "@testing-library/react";
+import { describe, it, vi } from "vitest";
+import { ToggleField } from "./toggle-field";
+
+describe("Toggle Component", () => {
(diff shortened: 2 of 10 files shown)


packages/design-system/src/globals.css
```diff
@@ -16,7 +16,9 @@
     -ms-overflow-style: none; /* IE and Edge */
     scrollbar-width: none; /* Firefox */
   }
-
+  body {
+		@apply dark:bg-black  min-h-screen relative;
+	}
```
YasielCabrera: we shouldn't apply this styles to the library, if we do this wherever the library is used it is going to change the page background. I recommend to do this in the storybook related config files instead

packages/design-system/src/scalars/components/toggle/toggle-field.test.tsx
```diff
@@ -0,0 +1,74 @@
+import { render, screen, fireEvent } from "@testing-library/react";
+import { describe, it, vi } from "vitest";
+import { ToggleField } from "./toggle-field";
+
+describe("Toggle Component", () => {
+  const mockOnChange = vi.fn();
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
```
YasielCabrera: If I'm not wrong, this is in the test setups so this is not needed here

packages/design-system/src/scalars/components/toggle/toggle-field.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  it("should render checked status with a label on the left", () => {
+    render(<ToggleField label="Test Label" checked={true} />);
+    expect(screen.getByText("Test Label")).toBeInTheDocument();
+  });
+
+  it("should not render the label when not provided", () => {
+    render(<ToggleField />);
+    expect(screen.queryByText("Test Label")).not.toBeInTheDocument();
+  });
```
YasielCabrera, line 36: these two tests are the same

packages/design-system/src/scalars/components/toggle/toggle.test.tsx
```diff
@@ -1,75 +1,97 @@
...
+  const defaultProps = {
+    onChange: vi.fn(),
+    disabled: false,
+    checked: false,
+    required: false,
+  };
+
   beforeEach(() => {
     vi.clearAllMocks();
   });
```
YasielCabrera: same that above, this shouldn't be needed here

packages/design-system/src/scalars/components/toggle/toggle.test.tsx
```diff
@@ -1,75 +1,97 @@
...
 
-  it("should match snapshot", () => {
-    const { asFragment } = render(<ToggleField />);
-    expect(asFragment()).toMatchSnapshot();
-  });
+  describe("Functionality", () => {
```
YasielCabrera: the description for this component should be more descriptive or direct, "functionality" doesn't lead to what actually this component is

packages/design-system/src/scalars/components/toggle/toggle.test.tsx
```diff
@@ -1,75 +1,97 @@
...
+    });
   });
 
-  it("renders with custom className", () => {
-    // this is a custom class name for testing purposes
-    // eslint-disable-next-line tailwindcss/no-custom-classname
-    render(<ToggleField className="custom-class" />);
-    const toggle = screen.getByTestId("custom-class");
-    expect(toggle).toHaveClass("custom-class");
+  describe("Rendering", () => {
```
YasielCabrera: same than above

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/toggle/toggle-field.test.tsx
alejandrocabriales: you are rigth, done

packages/design-system/src/scalars/components/toggle/toggle-field.test.tsx
alejandrocabriales: not, they test different case

packages/design-system/src/globals.css
alejandrocabriales: you are right , done

packages/design-system/src/scalars/components/toggle/toggle.test.tsx
alejandrocabriales: you are right, done

packages/design-system/src/scalars/components/toggle/toggle.test.tsx
alejandrocabriales: you are rigth, done

packages/design-system/src/scalars/components/toggle/toggle.test.tsx
alejandrocabriales: you are rigth, done

YasielCabrera (approved):

PR #462, opened 2024-10-30
fix(codegen): from json file

froid1911:


packages/codegen/src/codegen/utils.ts
```diff
@@ -13,7 +13,7 @@ export async function loadDocumentModel(
       documentModel = file.state.global;
     } else if (path.endsWith(".json")) {
       const data = fs.readFileSync(path, "utf-8");
-      const document = JSON.parse(data) as DocumentModelState;
```
acaldas, line 16: This doesn't break the zip loading?

froid1911: only if the zip file ends with .json, see line 14

acaldas:
Nice, there just seems to an issue with the commas.

Screenshot 2024-10-30 at 15 57 31

packages/codegen/src/codegen/utils.ts
acaldas: Nice, there is an eslint error due to data being any.

acaldas: Ah I think this was done to support generating from the spec.json file. I think we need a check that gets the document state on both cases.

PR #466, opened 2024-10-30
feat(codegen): changed bundle to esm

acaldas:
Bundles codegen as esm package.
Also updated document-model dependency.

### DIFF
--- a/packages/codegen/src/codegen/.hygen/templates/package.json
+++ b/packages/codegen/src/codegen/.hygen/templates/package.json
@@ -1,0 +1,3 @@
+{
+  "type": "commonjs"
+}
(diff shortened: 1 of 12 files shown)


packages/codegen/src/codegen/.hygen/templates/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "commonjs"
```
acaldas, line 2: This file is needed so node knows that template files are commonjs instead of ESM

packages/codegen/tsup.config.ts
```diff
@@ -17,24 +22,43 @@ function buildDtsEntries(exports: Record<string, string>) {
...
-    entry: {
-      ...buildDtsEntries(pkg.exports),
-      cli: "src/cli.ts",
+  {
+    entry: ["src/**/.hygen/templates/**/*"],
+    outDir: "dist/codegen/.hygen/templates",
+    splitting: false,
+    sourcemap: true,
+    clean: false,
+    format: "cjs", // hygen does not support esm
```
acaldas, line 51: Hygen expects template files to be commonjs

PR #478, opened 2024-11-02
feat(scalars): update the opinionated validation in the toggle-field component

alejandrocabriales:
## Description
- Should validate that the **Label** is only a string.
- Should the **required** field be set with true/false.
- Should display the proper status when the field is required.
- Should the **optional** field be set with true/false/ not set (null).
- Should allow setting the **Value** config with true/false (or undefined if the field is optional)
- Should allow setting the **Default** status true/false to the required field (or undefined to the optional field)
- Should validate that the Read Only (not set???) field is not editable.
- Should display Error/Warning messages.

### DIFF
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -88,12 +88,17 @@
     "rollup-plugin-visualizer": "^5.12.0",
     "storybook": "^8.3.4",
     "storybook-addon-pseudo-states": "^4.0.2",
-    "tsx": "^4.19.1"
+    "@hookform/resolvers": "^3.9.1",
+     "tsx": "^4.19.1",
+    "zod": "^3.23.8"
+
   },
   "peerDependencies": {
     "document-model": "^2.1.0",
     "react": "^18.3.1",
-    "react-dom": "^18.3.1"
+    "react-dom": "^18.3.1",
+    "@hookform/resolvers": "^3.9.1",
+    "zod": "^3.23.8"
   },
   "dependencies": {
     "@internationalized/date": "^3.5.6",
@@ -125,6 +130,10 @@
     "tailwindcss-animate": "^1.0.7",
     "usehooks-ts": "^3.1.0",
     "viem": "^2.16.1",
-    "wagmi": "^2.12.17"
+    "wagmi": "^2.12.17",
+    "form": "^0.2.5",
+    "add": "^2.0.6",
+    "@radix-ui/react-slot": "^1.1.0",
+    "@radix-ui/react-form": "^0.1.0"
   }
 }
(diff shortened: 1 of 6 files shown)


packages/design-system/package.json
```diff
@@ -125,6 +130,10 @@
     "tailwindcss-animate": "^1.0.7",
     "usehooks-ts": "^3.1.0",
     "viem": "^2.16.1",
-    "wagmi": "^2.12.17"
+    "wagmi": "^2.12.17",
+    "form": "^0.2.5",
+    "add": "^2.0.6",
+    "@radix-ui/react-slot": "^1.1.0",
+    "@radix-ui/react-form": "^0.1.0"
```
YasielCabrera, line 137: Do we need all those new dependencies? I'm noting that `zod` is duplicated here

packages/design-system/src/scalars/components/fragments/toggle-field/schema.ts
```diff
@@ -0,0 +1,26 @@
+import { z } from "zod";
+
+export const formSchema = z
+  .object({
+    checked: z.boolean().optional(),
+    required: z.boolean().optional(),
```
YasielCabrera, line 6: we should validate the `checked` property only, the `required` property would be acting as a modifier of the validation of the toggle value

packages/design-system/src/scalars/components/fragments/toggle-field/toggle-field.tsx
```diff
@@ -20,11 +23,25 @@ const ToggleField: React.FC<ToggleFieldProps> = ({
...
+  const validateField = () => {
+    return [
+      ...errors,
+      ...validateRequiredField(value ?? defaultValue, required),
+    ];
+  };
+
+  const errorMerge = validateField();
+  const showError = touched && !!errorMerge.length;
+  const showWarning = touched && !!warnings.length;
```
YasielCabrera, line 44: we should reuse this logic in a hook, so we're able to reuse it in each field, otherwise, the same logic will be implemented in each Field component

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/fragments/toggle-field/schema.ts
alejandrocabriales: you are right, this will close , because we move to different path for validation

packages/design-system/src/scalars/components/fragments/toggle-field/toggle-field.tsx
alejandrocabriales: you are right, this will close , because we move to different path for validation

packages/design-system/package.json
alejandrocabriales: you are right, this will close , because we move to different path for validation

PR #484, opened 2024-11-05
feat(document-model-libs): QA updates 2

ryanwolhuter:
New QA updates:

- Remove input on type for JSON editor for now, the debounce is being weird
- add delete buttons for modules and operations
- add inputs for reducer exceptions
- add inputs for operation descriptions
- add tabs layout for global and local

### DIFF
--- a/packages/document-model-libs/.storybook/main.ts
+++ b/packages/document-model-libs/.storybook/main.ts
@@ -1,10 +1,8 @@
-import { getConfig } from "@powerhousedao/codegen";
 import type { StorybookConfig } from "@storybook/react-vite";
 
-const { editorsDir } = getConfig();
 
 const config: StorybookConfig = {
-  stories: [`../${editorsDir}/**/*.stories.tsx`],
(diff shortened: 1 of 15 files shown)


packages/document-model-libs/.storybook/main.ts
```diff
@@ -1,10 +1,8 @@
-import { getConfig } from "@powerhousedao/codegen";
 import type { StorybookConfig } from "@storybook/react-vite";
 
-const { editorsDir } = getConfig();
```
acaldas, line 4: This doesn't work?

acaldas (approved):

packages/document-model-libs/.storybook/main.ts
ryanwolhuter: yeah at least on my machine, says there's a problem with cjs vs mjs 😢 it only breaks storybook. I think its a problem with the storybook runtime and ts config

PR #499, opened 2024-11-08
feat(codegen): generate actions exceptions

gpuente:
## Description:

- Added code generation for exceptions

Code generation sample:

```typescript
export type ErrorCode =
    | 'InvalidAddressFormat'
    | 'UnknownNetwork'
    | 'DuplicatedName'
    | 'NameTooLong'
    | 'TitleTooLong'
    | 'InvalidName'
    | 'InvalidTitle';

export interface ReducerError {
    errorCode: ErrorCode;
}

export class InvalidAddressFormat extends Error implements ReducerError {
    errorCode = 'InvalidAddressFormat' as ErrorCode;
    constructor(message = 'InvalidAddressFormat') {
        super(message);
    }
}

export class UnknownNetwork extends Error implements ReducerError {
    errorCode = 'UnknownNetwork' as ErrorCode;
    constructor(message = 'UnknownNetwork') {
        super(message);
    }
}

export class DuplicatedName extends Error implements ReducerError {
    errorCode = 'DuplicatedName' as ErrorCode;
    constructor(message = 'DuplicatedName') {
        super(message);
    }
}

export class NameTooLong extends Error implements ReducerError {
    errorCode = 'NameTooLong' as ErrorCode;
    constructor(message = 'NameTooLong') {
        super(message);
    }
}

export class TitleTooLong extends Error implements ReducerError {
    errorCode = 'TitleTooLong' as ErrorCode;
    constructor(message = 'TitleTooLong') {
        super(message);
    }
}

export class InvalidName extends Error implements ReducerError {
    errorCode = 'InvalidName' as ErrorCode;
    constructor(message = 'InvalidName') {
        super(message);
    }
}

export class InvalidTitle extends Error implements ReducerError {
    errorCode = 'InvalidTitle' as ErrorCode;
    constructor(message = 'InvalidTitle') {
        super(message);
    }
}


export const errors = {
    SetWallet: {
        InvalidAddressFormat,
        UnknownNetwork,
    },
    AddActiveSigner: {
        DuplicatedName,
        NameTooLong,
        TitleTooLong,
        InvalidName,
        InvalidTitle,
    },
};


```


packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-document-model-module/customReducers.esm.t
```diff
@@ -9,6 +9,7 @@ unless_exists: true
 */
 
 import { <%= h.changeCase.pascal(documentType) %><%= h.changeCase.pascal(module) %>Operations } from '../../gen/<%= module %>/operations';
+import { default as Exceptions } from '../../gen/<%= module %>/exceptions';
```
acaldas: I wouldn't import the errors here yet since we are not doing anything with them.

Right now, the Operation type has a error?: string; property, it might be helpful in the future to have a way to infer the errors from each action type. But not something for now.

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-document-model-module/exceptions.esm.t
```diff
@@ -0,0 +1,30 @@
...
+    constructor(message = '<%= h.changeCase.pascal(error.name) %>') {
+        super(message);
+    }
+}
+
+<% }); _%>
+<% }); _%>
+
+
+export default {
```
acaldas: Let's avoid default exports and use named exports instead.

Exporting the error classes is interesting but I think we should also export each error an object and type/interface for each error, since classes are for runtime only.

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-document-model-module/exceptions.esm.t
acaldas, line 1: For consistency, we should keep the `error` naming instead of `exceptions`.

gpuente: fix: [fe3acc8](https://github.com/powerhouse-inc/powerhouse/pull/499/commits/fe3acc8cc9ed6c6420ba92eb644fd38a01fbcbb5)

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-document-model-module/exceptions.esm.t
gpuente: fix: [fe3acc8](https://github.com/powerhouse-inc/powerhouse/pull/499/commits/fe3acc8cc9ed6c6420ba92eb644fd38a01fbcbb5)

PR #508, opened 2024-11-11
fix(document-model-libs): avoid unknown scalars after re-open document

gpuente:


gpuente:
This PR is not longer needed

PR #512, opened 2024-11-12
feat(scalars): add the validations for int and bigInt

alejandrocabriales:
## Ticket

## Description
- Should allow setting **minValue:** number as a minimum allowed value.
- Should show the proper message when the minValue is not met.
- Should allow setting **maxValue:** number as a maximum allowed value.
- Should show the proper message when the maxValue has been exceeded.
-  Should allow setting **allowNegative:** Boolean.
- Should allow negative numbers if the allowNegative config is true.
- Should allow setting **isBigInt:** Boolean — Whether large numbers beyond Int are allowed.

### DIFF
--- a/packages/design-system/src/scalars/components/number-field/numberFieldValidations.ts
+++ b/packages/design-system/src/scalars/components/number-field/numberFieldValidations.ts
@@ -13,18 +13,22 @@ export const validateIsBigInt =
   (props: NumberFieldProps) => (value: string) => {
     const isLargeNumber = Math.abs(Number(value)) > Number.MAX_SAFE_INTEGER;
     return isLargeNumber && !props.isBigInt
-      ? "Value is too large for standard integer, set isBigInt to true"
+      ? "Value is too large for standard integer"
       : true;
   };
 
 export const validatePrecision =
   (props: NumberFieldProps) => (value: string) => {
-    if (props.precision === 0) return true;
     if (props.precision === undefined) return true;
+
     const decimalPart = value.toString().split(".")[1];
-    return decimalPart && decimalPart.length <= (props.precision ?? 0)
+    if (props.precision === 0) {
+      return !decimalPart ? true : "Value must be an integer";
+    }
+
+    return decimalPart && decimalPart.length <= props.precision
       ? true
-      : `Value must have ${props.precision} decimal places`;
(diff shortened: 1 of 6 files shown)


packages/design-system/src/scalars/components/number-field/numberFieldValidations.ts
```diff
@@ -0,0 +1,45 @@
...
+      : Number(value) > 0
+        ? true
+        : "Value must be a positive value";
+  };
+
+export const validateIsBigInt =
+  (props: NumberFieldProps) => (value: string) => {
+    const isLargeNumber = Math.abs(Number(value)) > Number.MAX_SAFE_INTEGER;
+    return isLargeNumber && !props.isBigInt
+      ? "Value is too large for standard integer, set isBigInt to true"
```
YasielCabrera: You should omit this part: "set isBigInt to true". This is a message to be presented to the users no developers, and users can not change the code

packages/design-system/src/scalars/components/number-field/numberFieldValidations.ts
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+export const validatePrecision =
+  (props: NumberFieldProps) => (value: string) => {
+    if (props.precision === 0) return true;
+    if (props.precision === undefined) return true;
+    const decimalPart = value.toString().split(".")[1];
+    return decimalPart && decimalPart.length <= (props.precision ?? 0)
+      ? true
+      : `Value must have ${props.precision} decimal places`;
```
YasielCabrera: If precision is 0 and receives a value with 1 decimal place, it should fail, right?

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/number-field/numberFieldValidations.ts
alejandrocabriales: You are right, done

packages/design-system/src/scalars/components/number-field/numberFieldValidations.ts
alejandrocabriales: good catch , done

YasielCabrera (approved):

PR #515, opened 2024-11-12
feat(scalars): validate float cases and fix some stories

alejandrocabriales:
## Ticket

## Description
- Should display the proper status when the required state is required.
- Should allow setting the **Value:String** config with true/false (or null if the field is optional)
- Should allow setting the **Default** status true/false to the required field (or null to the optional field) 
- Should allow setting **showErrorOnBlur:** true / false.
- Should show the error message only when the field loses focus if showErrorOnBlur is true.
- Should allow setting the **showErrorOnChange:** true / false.
- Should display the error message as the user types if showErrorOnChange is true.
- Should allow **precision:** number.
- Should allow the number of decimal places specified in precision config.
- Should allow setting **trailingZeros:** boolean.
- Should allow setting **minValue:** number as a minimum allowed value.
- Should show the proper message when the minValue is not met.
- Should allow setting **maxValue:** number as a maximum allowed value.
- Should show the proper message when the maxValue has been exceeded.
- Should allow setting **decimalRequired:** Boolean — Whether a decimal point is required.

### DIFF
--- a/packages/design-system/src/scalars/components/fragments/toggle-field/toggle.stories.tsx
+++ b/packages/design-system/src/scalars/components/fragments/toggle-field/toggle.stories.tsx
@@ -70,9 +70,16 @@ const meta: Meta<typeof ToggleField> = {
 export default meta;
 type Story = StoryObj<typeof meta>;
 
-export const Default: Story = {
+export const DefaultChecked: Story = {
   args: {
     value: true,
+  },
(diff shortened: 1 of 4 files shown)


nesdc88 (approved):
LGTM

YasielCabrera (commented):
lgtm

PR #516, opened 2024-11-12
feat(scalars): add url field

YasielCabrera:
## Ticket
https://trello.com/c/9PSFnLlU/688-5-urlfield-url-urlyoutube-urlgithub-urldiscord-urldiscourse

## Description
Implemented Url Field base

### DIFF
--- a/packages/design-system/src/scalars/components/url-field/url-field.tsx
+++ b/packages/design-system/src/scalars/components/url-field/url-field.tsx
@@ -1,0 +1,94 @@
+import React, { useId } from "react";
+import {
+  Input,
+  FormGroup,
+  withFieldValidation,
+  FormLabel,
+  FormDescription,
+  FormMessageList,
+} from "../fragments";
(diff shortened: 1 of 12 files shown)


packages/design-system/src/scalars/components/url-field/url-field.tsx
```diff
@@ -0,0 +1,94 @@
...
+            readOnly
+            // tailwind does not support dynamic classes
+            style={{ width: `${(protocol.length + 1) * 8}px` }}
+            className={`w-auto rounded-r-none border-r-0 focus:z-10`}
+          />
+        )}
+        <Input
+          id={id}
+          type="url"
+          role="textbox"
```
alejandrocabriales: Perhaps we don’t need `role="textbox"` here? An <input type="url"> is already accessible without it.

YasielCabrera: I remove it, thanks!

alejandrocabriales:
LGTM

PR #520, opened 2024-11-13
feat: design system/date picker fixes

ryanwolhuter:
- remove milliseconds from date input
- extract entry time label to component

there was a bug because of how the entry time label was getting its value from the form — using `watch` in the function that returns the component causes the whole component to rerender on any change of the value. this caused the input to lose focus and the date picker to close immediately on input.

to solve this we move this label component to its own component, which can then use the `useWatch` hook instead of `watch`. This means that the rendering is contained in the label component itself, and therefore the value is not changing and rerendering the component.


acaldas (approved):
Nice!

PR #521, opened 2024-11-13
fix(scalars): fix message for validations and add better descriptions

alejandrocabriales:
## Ticket

## Description
-  Should allow the number of decimal places specified in precision config.


YasielCabrera (approved):
lgtm

PR #529, opened 2024-11-18
chore(scalars): adjust ToggleField stories to new variant proposal

alejandrocabriales:
## Ticket
https://trello.com/c/5oApGmej/761-standardize-storybook-documentation-for-component-consistency

## Description
Adjust ToggleField stories to new variant proposal

### DIFF
--- a/packages/design-system/src/scalars/components/fragments/toggle-field/toggle.stories.tsx
+++ b/packages/design-system/src/scalars/components/fragments/toggle-field/toggle.stories.tsx
@@ -13,12 +13,8 @@ const meta: Meta<typeof ToggleField> = {
   decorators: [withForm],
   argTypes: {
     ...getDefaultArgTypes({
-      enabledArgTypes: {
-        value: true,
-        defaultValue: true,
-      },
       valueControlType: "boolean",
-      valueType: "true",
+      valueType: "boolean",
     }),
     ...getValidationArgTypes({
       enabledArgTypes: {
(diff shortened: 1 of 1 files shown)


packages/design-system/src/scalars/components/fragments/toggle-field/toggle.stories.tsx
```diff
@@ -1,64 +1,30 @@
...
-    },
-    onChange: {
-      action: "checked changed",
-      description: "Callback when the checked state changes",
-      table: {
-        type: { summary: "(checked: boolean) => void" },
+    ...getDefaultArgTypes({
+      enabledArgTypes: {
+        value: true,
+        defaultValue: true,
```
YasielCabrera: value and default value are enabled by default. But you should add `valueControlType: "boolean"` and `valueType: "boolean"`

packages/design-system/src/scalars/components/fragments/toggle-field/toggle.stories.tsx
```diff
@@ -1,64 +1,30 @@
...
-      description: "Name attribute for the toggle input",
-    },
+      valueControlType: "boolean",
+      valueType: "true",
+    }),
+    ...getValidationArgTypes({
+      enabledArgTypes: {
+        customValidator: false,
+      },
+    }),
```
YasielCabrera: the toggle field doesn't have support of `customValidator`? If now, you should add it, it is a requirement

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/fragments/toggle-field/toggle.stories.tsx
alejandrocabriales: you are right, done

YasielCabrera (approved):
lgtm

PR #536, opened 2024-11-19
feat(scalars): add final style for number input

alejandrocabriales:
## Ticket

## Description
-  Should update the styles with the Final design.


packages/design-system/src/globals.css
```diff
@@ -102,6 +102,9 @@
   --color-orange-700: 32 100% 60%;
   --color-orange-800: 33 100% 55%;
   --color-orange-900: 32 100% 50%;
+  --color-charcoal-300: 221 17% 67%;
+  --color-charcoal-700: 219 17% 34%;
+  --color-charcoal-900: 220 17% 17%;
```
YasielCabrera: we should add the entire charcoal palette instead of just three colors

packages/design-system/src/globals.css
```diff
@@ -171,6 +174,9 @@
   --orange-700: hsl(var(--color-orange-700));
   --orange-800: hsl(var(--color-orange-800));
   --orange-900: hsl(var(--color-orange-900));
+  --charcoal-300: hsl(var(--color-charcoal-300));
+  --charcoal-700: hsl(var(--color-charcoal-700));
+  --charcoal-900: hsl(var(--color-charcoal-900));
```
YasielCabrera: this looks to be repeated

packages/design-system/src/scalars/components/fragments/checkbox-field/checkbox-field.tsx
```diff
@@ -48,6 +48,7 @@ const CheckboxRaw: React.FC<CheckboxFieldProps> = ({
           hasError={hasError}
           description={description}
           className={cn(!disabled && "group-hover:cursor-pointer")}
+          data-testid="label-testid"
```
YasielCabrera: why do we need to add `data-testid` here? You can get it in the test by text or by role as it was before

packages/design-system/src/scalars/components/fragments/toggle-field/toggle-field.tsx
```diff
@@ -49,6 +49,7 @@ const ToggleRaw: React.FC<ToggleFieldProps> = ({
             disabled={disabled}
             required={required}
             description={description}
+            data-testid="label-testid"
```
YasielCabrera: similar comment that before, why do we need to add `data-testid` here to make it work? My concern is that this attribute is going to be present in the build bundle too

packages/design-system/src/shared-tailwind.config.ts
```diff
@@ -91,6 +91,9 @@ const theme = {
     "orange-700": "hsl(var(--color-orange-700) / <alpha-value>)",
     "orange-800": "hsl(var(--color-orange-800) / <alpha-value>)",
     "orange-900": "hsl(var(--color-orange-900) / <alpha-value>)",
+    "charcoal-300": "hsl(var(--color-charcoal-300) / <alpha-value>)",
+    "charcoal-700": "hsl(var(--color-charcoal-700) / <alpha-value>)",
+    "charcoal-900": "hsl(var(--color-charcoal-900) / <alpha-value>)",
```
YasielCabrera: similar comment than in global.css, we should add the entire charcoal palette here too

packages/design-system/src/scalars/components/boolean-field/boolean-field.test.tsx
```diff
@@ -28,25 +28,30 @@ describe("BooleanField", () => {
...
-  it("should pass all props to ToggleField when isToggle is true", () => {
-    const { getByRole } = renderWithForm(
-      <BooleanField name="test" {...commonProps} isToggle={true} />,
+  it("should pass all props to ToggleField when isToggle is true", async () => {
+    const { getByRole, findByTestId } = renderWithForm(
+      <BooleanField
+        name="test"
+        {...commonProps}
+        isToggle={true}
+        label="Test Label"
```
YasielCabrera: this label is already part of the `commonProps`

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/fragments/checkbox-field/checkbox-field.tsx
alejandrocabriales: You are rigth , We can use byText, the role was unnecesary in that case, so the option getText its better option
thanks for the feedback

packages/design-system/src/scalars/components/fragments/toggle-field/toggle-field.tsx
alejandrocabriales: You are rigth , We can use byText, the role was unnecesary in that case, so the option getText its better option
thanks for the feedback

packages/design-system/src/globals.css
alejandrocabriales: you are right, done

packages/design-system/src/shared-tailwind.config.ts
alejandrocabriales: you are right done

packages/design-system/src/scalars/components/boolean-field/boolean-field.test.tsx
alejandrocabriales: yo are right

YasielCabrera (approved):
lgtm

PR #537, opened 2024-11-20
feat(scalars): add amount currency variants

alejandrocabriales:
## Ticket

## Description
- Default status with currency. 
- Active status with currency
- Default - Select hover with currency
- Default - Select open with currency 
- Default - Select focus with currency isPopoverOpen 
- Default - Read only with currency
- Default - Warning
 - Default - Error
 - Default - Multi-error
 - Description status
- Mandatory status
- Should set **precision** as a Number.
- Should set **minValue** as a Number.
- Should set **maxValue** as a Number.
- Should set **allowNegative** as a Boolean.


nesdc88 (approved):
LGTM

PR #544, opened 2024-11-20
fix(document-model): do not generate a new timestamp for operations

gpuente:
## Description:

- Prevent replace operation timestamp


ryanwolhuter (approved):
nice! will need a way to mock the date in the test, but otherwise looks good 👍

PR #554, opened 2024-11-23
feat(scalars): improve types, logic and behavior in Enums related components & Textarea

nesdc88:
## Ticket
https://trello.com/c/zZlPyvEY/754-final-design-4-enum

## Description
- Should add "Auto" to EnumField variants & implement the behavior
- Should use only "options" in the EnumField (remove disabledOptions and optionLabels)
- Should remove the icon (and behavior) of deleting a single selected option in the Multi-select
- Should change the dropdown trigger icon only in the case of the Combobox (Searchable Select)
- Should add "optionsCheckmark" prop to SelectField & implement the behaviour
- Logic improvements


alejandrocabriales (approved):
lgtm

PR #562, opened 2024-11-26
feat(scalars): add base on amount-field validations

alejandrocabriales:
## Ticket

## Description
Should set **trailingZeros** as a Boolean.
 Should set **allowedConcurrecies** as an Array of strings.
 Should set **allowedTokens** as an Array of strings.


YasielCabrera (changes requested):
The value of the amount field doesn't match with the requirements

YasielCabrera (approved):
lgtm

PR #564, opened 2024-11-26
feat(scalars): add an example form with EnumField

nesdc88:
## Ticket
https://trello.com/c/L7DEdtjO/748-4-enumfield-single-multi-select

## Description
- Should add the component in the form in the Storybook


alejandrocabriales (approved):
lgtm

PR #566, opened 2024-11-26
fix(scalars): fix autoexpand behavior when multiline is false in TextareaField

nesdc88:
## Ticket

## Description
- If autoexpand is true and multiline is false it should show a textarea but can't allow multiple lines/enter/line breaks even on pasted content


packages/design-system/src/scalars/components/fragments/textarea-field/textarea-field.tsx
```diff
@@ -84,29 +86,51 @@ const TextareaFieldRaw = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
...
+      if (autoExpand) {
+        adjustHeight();
+      }
+    };
+
+    const handleKeyDown = (e: React.KeyboardEvent<HTMLTextAreaElement>) => {
+      // Prevent Enter key if multiline is false
+      if (!multiline && e.key === "Enter") {
+        e.preventDefault();
+        e.stopPropagation();
```
YasielCabrera: it should not stop the propagation in case an user wants a custom event

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/fragments/textarea-field/textarea-field.tsx
nesdc88: Deleted

YasielCabrera (approved):
lgtm

PR #569, opened 2024-11-26
fix(scalars): field internal state sync

YasielCabrera:
## Ticket
https://trello.com/c/Z1QZdy5Z

## Description
- Fix field internal sync with external value  when `value` and `onChange` is provided
- `name` control is now readonly to avoid storybook issues
- Fixed error when errors or warnings are added to the components in the storybook


nesdc88 (approved):
LGTM

PR #571, opened 2024-11-27
fix(scalars): fix the colors and remove the autoFocus in the docs

alejandrocabriales:
## Ticket

## Description
 - The gap between the field and the label should be 11px. **Current Output:** The gap is 8px (.5rem). 
 - The autoFocus prop should not be visible in the component. **Current Output:** The prop is visible as the first element.
 - NumberField. Dark mode. The description should have the color Gray/500. The description is displayed with color Gray/600


YasielCabrera (approved):
lgtm

PR #580, opened 2024-11-27
fix(scalars): remove unncesary validations and fix stories

alejandrocabriales:
## Ticket

## Description
- NumberField. Inserting a number and deleting it. If the field is optional, no error message should be displayed. After deleting the number an error message is shown "Value must be a positive value".

-  NumberField. Inserting the value 0. **Expected Output:** The error message should differ from the positive value.  An error message is displayed "Value must be a positive value" even when the minValue set is 0. 

-  NumberField. numericType control. A list with the numeric types should be displayed. An input field is displayed.


YasielCabrera (approved):
lgtm

PR #585, opened 2024-11-28
fix(scalars): add arrow step functionality and tests

alejandrocabriales:
## Ticket

## Description
- The steps representation should be the same as shown in the design.  The steps are represented with different styles than those defined in the design.


packages/design-system/src/scalars/components/amount-field/amount-field.tsx
```diff
@@ -167,6 +167,7 @@ const AmountFieldRaw: FC<AmountFieldProps> = ({
               disabled={disabled}
               onChange={handleOnChangeSelect}
               options={options}
+              className="flex h-10"
```
YasielCabrera: the right height should be ensured in the `SelectField` component

packages/design-system/src/scalars/components/number-field/number-field.tsx
```diff
@@ -26,7 +27,7 @@ export const NumberFieldRaw = forwardRef<HTMLInputElement, NumberFieldProps>(
       description,
       value,
       defaultValue,
-      onChange,
+      onChange = () => {},
```
YasielCabrera: `onChange` should be optional without a default value, Adding an "empty function" is going to still be a function and could change how the validation is done making it less efficient

packages/design-system/src/scalars/components/number-field/number-field.tsx
```diff
@@ -66,7 +68,23 @@ export const NumberFieldRaw = forwardRef<HTMLInputElement, NumberFieldProps>(
...
+      const adjustment = (operation === "increment" ? 1 : -1) * (step || 1);
+      let newValue = currentValue + adjustment;
 
+      if (maxValue !== undefined && newValue > maxValue) {
+        newValue = maxValue;
+      } else if (minValue !== undefined && newValue < minValue) {
+        newValue = minValue;
+      }
+
+      onChange(newValue as unknown as React.ChangeEvent<HTMLInputElement>);
```
YasielCabrera: You should create an event, no casting it, even when TS accept it because you're forcing it type, it may have error in other places where the onChange is expecting a real event

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/amount-field/amount-field.tsx
alejandrocabriales: you are right, done

packages/design-system/src/scalars/components/number-field/number-field.tsx
alejandrocabriales: done

packages/design-system/src/scalars/components/number-field/number-field.tsx
alejandrocabriales: you are right, done

packages/design-system/src/scalars/components/number-field/number-field.test.tsx
```diff
@@ -106,4 +114,107 @@ describe("NumberField Component", () => {
...
+
+    expect(mockOnChange).toHaveBeenCalledWith(
+      expect.objectContaining({
+        target: {
+          value: "10",
+          name: "Label",
+        },
+      } as { target: { value: string } }),
+    );
+  });
```
YasielCabrera: when the value reaches the limits it shouldn't allow to increment/decrement but it also shouldn't invoke the `onChange` because the value doesn't change

packages/design-system/src/scalars/components/number-field/number-field.test.tsx
```diff
@@ -106,4 +114,107 @@ describe("NumberField Component", () => {
...
+
+    expect(mockOnChange).toHaveBeenCalledWith(
+      expect.objectContaining({
+        target: {
+          value: "0",
+          name: "Label",
+        },
+      } as { target: { value: string } }),
+    );
+  });
```
YasielCabrera: same comment above applies here

packages/design-system/src/scalars/components/number-field/number-field.tsx
```diff
@@ -66,7 +68,46 @@ export const NumberFieldRaw = forwardRef<HTMLInputElement, NumberFieldProps>(
       trailingZeros,
       precision,
     );
+    const handleChange = (
+      e: React.MouseEvent<HTMLButtonElement>,
+      operation: "increment" | "decrement",
+    ) => {
+      e.preventDefault();
+      const currentValue = Number(displayValue || defaultValue || 0);
```
YasielCabrera: You should use nullish operator instead (`??`) because if `displayValue` is `0` it is going to be taken as `false` and then if `defaultValue` is different from `0`, that's the value to be evaluated here, which is an issue

packages/design-system/src/scalars/components/number-field/number-field.tsx
```diff
@@ -66,7 +68,46 @@ export const NumberFieldRaw = forwardRef<HTMLInputElement, NumberFieldProps>(
...
+      const adjustment = (operation === "increment" ? 1 : -1) * (step || 1);
+      let newValue = currentValue + adjustment;
 
+      if (maxValue !== undefined && newValue > maxValue) {
+        newValue = maxValue;
+      } else if (minValue !== undefined && newValue < minValue) {
+        newValue = minValue;
+      }
```
YasielCabrera: Following my comment about this in the tests: if the new value is outside the limits, the function should keep the previous value and not continue. Maybe you can enforce this by disabling the step buttons when the next value is not going to be in range. That would also apply when the next value is above/below the safe integer values (unless the type is bigint because if that happens you're going to fall in number overflow errors

packages/design-system/src/scalars/components/number-field/number-field.tsx
```diff
@@ -66,7 +68,46 @@ export const NumberFieldRaw = forwardRef<HTMLInputElement, NumberFieldProps>(
...
+        nativeEvent: new Event("input"),
+        preventDefault: () => {},
+        stopPropagation: () => {},
+        persist: () => {},
+        isDefaultPrevented: () => false,
+        isPropagationStopped: () => false,
+        timeStamp: Date.now(),
+        type: "input",
+      };
+      onChange?.(changeEvent);
```
YasielCabrera: This wouldn't be the right way to trigger the `onChange`, it is an object but not an instance of `Event` or any of its subclasses, and the methods are mocked and will not work if a user needs for example, the `stopPropagation`. Maybe something like this:
- This example is maybe what you need to do here: https://jetrockets.com/blog/onchange-trigger-example-with-react
- The MDN docs about how to do it: https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events#triggering_built-in_events

packages/design-system/src/scalars/components/number-field/number-field.tsx
```diff
@@ -80,30 +121,52 @@ export const NumberFieldRaw = forwardRef<HTMLInputElement, NumberFieldProps>(
...
+            onKeyDown={blockInvalidChar}
+            value={displayValue}
+            defaultValue={defaultValue}
+            onChange={onChange}
+            step={step}
+            onPaste={blockInvalidPaste}
+            ref={ref}
+            {...props}
+          />
+          {step && (
```
YasielCabrera: We should "hide" the steps only when they're `0` IMO. By default they should be enabled and the default value should be `1`. This is how number inputs behave by default in HTML. Take a look to this example: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_type_number

packages/design-system/src/scalars/components/number-field/number-field.tsx
```diff
@@ -80,30 +121,52 @@ export const NumberFieldRaw = forwardRef<HTMLInputElement, NumberFieldProps>(
...
+            min={minValue}
+            max={maxValue}
+            aria-valuemin={minValue}
+            aria-valuemax={maxValue}
+            aria-invalid={!!errors?.length}
+            onKeyDown={blockInvalidChar}
+            value={displayValue}
+            defaultValue={defaultValue}
+            onChange={onChange}
+            step={step}
```
YasielCabrera: If the native steps are disabled you shouldn't pass it down to the input

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/number-field/number-field.test.tsx
alejandrocabriales: you are right done

packages/design-system/src/scalars/components/number-field/number-field.test.tsx
alejandrocabriales: done

packages/design-system/src/scalars/components/number-field/number-field.tsx
alejandrocabriales: you righ, done

packages/design-system/src/scalars/components/number-field/number-field.tsx
alejandrocabriales: done

packages/design-system/src/scalars/components/number-field/number-field.tsx
alejandrocabriales: done

packages/design-system/src/scalars/components/number-field/number-field.tsx
alejandrocabriales: done

packages/design-system/src/scalars/components/number-field/number-field.tsx
alejandrocabriales: done

packages/design-system/src/scalars/components/number-field/number-field.tsx
```diff
@@ -67,6 +70,42 @@ export const NumberFieldRaw = forwardRef<HTMLInputElement, NumberFieldProps>(
...
+      if (
+        (maxValue !== undefined && newValue > maxValue) ||
+        (minValue !== undefined && newValue < minValue)
+      ) {
+        return;
+      }
+      if (newValue === currentValue) {
+        return;
+      }
+      const nativeEvent = new Event("input", {
```
YasielCabrera: `input` and `change` events are different. In  this case we're trying to trigger a "change" event instead

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/number-field/number-field.tsx
alejandrocabriales: you are right, done

YasielCabrera (approved):
lgtm

PR #589, opened 2024-11-29
feat(monorepo): try new preserve local deps protocol

ryanwolhuter:
the release on the monorepo was failing because of lockfile issues again.

these problems boil down to our usage of the workspace protocol, and how nx interacts with that when doing builds.

we found this comment in a thread about this problem:

https://github.com/nrwl/nx/issues/20978#issuecomment-2464559054

There is a still-to-be-documented config that has been introduced to nx that promises to fix these issues.

"generatorOptions": {
     "preserveLocalDependencyProtocols": true
}

This should allow us to automate our releases such that we no longer need to do redundant releases of all of the packages in order to release packages individually.

More importantly, it will stop the build system from removing all of the workspace:* from the package json files and committing that to github.


acaldas (approved):
🤞

PR #592, opened 2024-11-30
chore(scalars): use auto EOL prettier config in eslint to prevent issues on windows

YasielCabrera:
Auto EOL config of prettier to prevent EOL issues on windows


nesdc88 (approved):
LGTM

PR #601, opened 2024-12-02
feat(scalars): add number components transform

alejandrocabriales:
## Ticket

## Description
- NumberField. precision = 2, trailingZeros = false. Inserting a number with two zeros (e.g. 34.00)* The zeros should not be preserved, 34 should be displayed. The full number is displayed, and the zeros are displayed with trailingZeros in true or false. The same result is displayed.

- NumberField. props. Setting a prop (e.g MaxValue). Clear up the setting should be allowed. After the prop is set, it is not possible to clear up the config. Refreshing the Storybook doesn't work either.


YasielCabrera (approved):
lgtm

PR #605, opened 2024-12-02
feat(scalars): change scalars import path

YasielCabrera:
## Ticket
https://trello.com/c/RJXvos0e/766-fix-design-system-library-build-connect-demo

## Description
Change the scalars import to the root path of the `design-system` library


nesdc88 (approved):
LGTM

PR #606, opened 2024-12-02
fix(scalars): warning arg failing on stories

YasielCabrera:
## Ticket

## Description
Prevent the stories from failing when the `warning` arg is set


nesdc88 (approved):
LGTM

PR #617, opened 2024-12-04
fix(scalars): add correct validations in bigInt and validate per type

alejandrocabriales:
## Ticket

## Description
- Selecting the NegativeInt type and inserting the number 9.  The field should allow only values less than 0. 

- NumberField. isBigInt = true. Inserting text. **Expected Output:** An error message should be displayed. **Current Output:** The error message is displayed but when click out the field, the text becomes NaN.


nesdc88 (approved):
LGTM

PR #618, opened 2024-12-04
feat(monorepo): fix build problems

ryanwolhuter:
Our monorepo build system assumes that the build tools we are using are of the same version and are compatible with each other. 

The existing package.json files of several of the packages had their own installations of tools like vite and storybook.

These versions were not in sync, which meant that in some places and contexts the build would fail. This also made life difficult because we would need to maintain these deps in several places instead of just one. 

There was also a build warning for several of the packages that said our "exports" were defined in the wrong order. This was causing our editors and build tools to get confused and think that exports did not exist when they in fact do. For reference, the correct order is: types, require, import.

In this PR, we fix these issues by:

- having one installation of build tools, i.e. typescript, vite, and storybook

- using shared tsconfig options, so that we don't accidentally transpile packages in ways that are not compatible with each other

- using typescript's recommended method for handling monorepos, project references https://www.typescriptlang.org/docs/handbook/project-references.html

- fixing the order of definitions in the exports field of package.json files. This whole thing is a point of contention in the javascript community right now because it is very confusing.

- preferring to include files in tsconfig files instead of excluding ones we don't want, which prevents weird conflicts like for example with storybook and vite.


acaldas (approved):
Nice!

Was there an issue with the peerDependencies? They are not added to the bundle so users of the package need to have it installed.
I guess in the monorepo nx takes care of it but if someone installs one of packages it might be problematic?

ryanwolhuter:
> Nice!
> 
> Was there an issue with the peerDependencies? They are not added to the bundle so users of the package need to have it installed. I guess in the monorepo nx takes care of it but if someone installs one of packages it might be problematic?

Thanks ☺️ 

Yeah there are a couple of problems with how the deps were being done. 

- Dev deps in peer deps — this doesn't make sense, because the dev deps are already not included in the bundle. Putting dev deps in peer deps just confuses the build tools and doesn't give any benefit
- Peer deps from the workspace — this also doesn't make sense, because we want the packages to work independently. Expecting the user to have the other packages installed goes against this and confuses the build tools

Really the only deps we have that make sense to put in peer deps are the react- packages, since we can safely assume that users will know they need to have react installed to use our UI library code.

In terms of how our packages are bundled, I think we are actually going about optimization in the wrong way. We are trying to optimize bundle size for library code, which is actually counter productive because it causes our library code to include code that the consumer doesn't need in the bundle chunks.

We are chunking library code and shipping that in packages, when really what we want is to make all of our library code as tree-shakeable as possible so that consumers can import only what they need and then do their own chunking optimization.

The main reason we are having bundle size and tree shaking problems is our use of dangerous barrel files.

https://tkdodo.eu/blog/please-stop-using-barrel-files
https://dev.to/thepassle/a-practical-guide-against-barrel-files-for-library-authors-118c

I actually did a little experiment and removed the barrel file exports from just one of the directories in the design-system, and it already made a big difference. 

The current chunking-based optimization is also the reason we are seeing warnings like this:

`Export "createReducer" of module "src/document/utils/base.ts" was reexported through module "src/document/utils/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
  Either change the import in "src/document-model/gen/reducer.ts" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.`
  
  This is contributing to the inconsistent behavior of some deployments we've been seeing

acaldas:
I agree with everything, and nx might be fixing some of this already. For node it's a lot easier as we can just remove all dependencies from the bundle. For the browser we need to balance bundling the necessary dependency code while avoiding duplicated code, where two packages have the same dependency. There might also be the issue where two dependencies have conflicting versions of the same dependency.

The peerDependencies that were removed here where `document-model`, `@powerhousedao/scalars` and `graphql`.
`document-model` is the most core lib, so pretty much all packages depend on it. The scalars are also important to deduplicate as all packages should have access to the same set of scalars.
Multiple versions of graphql cause an exception which has come up sometimes in the past, here is an example: https://stackoverflow.com/questions/48769019/prisma-graphql-error-on-signup-mutation/49116060#49116060.

I think that should clearly separate which packages are meant for an end user to install himself and which ones are only building blocks for user-facing packages.
The non user-facing libs shouldn't even need a bundling step and just export typescript.
Ideally there would be only a single bundling step when building Connect/Switchboard/ph-cli.

About the barrel files, I agree they should be avoided as it is very easy to break tree-shaking. However, we also need to think about making it easy for the dev to import stuff from our packages, for instance, having to do an import per Design system component vs just importing them all on the same line.
I've seen some ways to indicate to the bundler that a module doesn't have side effects and can be removed safely if not used:
Vite custom method: https://x.com/_developit/status/1842225028009447896
Package.json `sideEffects` field: https://dev.to/markliu2013/deep-dive-into-sideeffects-configuration-14me

PR #630, opened 2024-12-05
fix(scalars): pattern control in storybook stories

YasielCabrera:
## Ticket

## Description
Fix pattern control in the storybook stories. Now the regex as string value is correctly transformed in actual `RegExp` objects making it work correctly in the stories. Also preventing invalid regex from breaking the story


alejandrocabriales:
LGTM

PR #631, opened 2024-12-05
feat(scalars): add the correct value in the onSubmit

alejandrocabriales:
## Ticket

## Description
NumberField. Submiting the value. The output should not be a string. The output is a string "".


YasielCabrera (approved):
lgtm

PR #633, opened 2024-12-05
chore(scalars): upgrade storybook/addon-docs

YasielCabrera:
Upgrade `@storybook/addon-docs` to the equivalent version of storybook used in the package


alejandrocabriales (approved):
LGTM

PR #649, opened 2024-12-06
feat(document-model-libs): chatroom document model

gpuente:
## Description:

- Added chatroom document model
- Added chatroom editor


acaldas (approved):

packages/document-model-libs/editors/chat-room-editor/chat-room-editor.stories.tsx
```diff
@@ -0,0 +1,13 @@
...
+import * as ChatRoomModule from "../../document-models/chat-room";
+
+const { meta, CreateDocumentStory: ChatRoom } = createDocumentStory(
+  Editor,
+  ChatRoomModule.reducer,
+  ChatRoomModule.utils.createDocument(),
+);
+export { ChatRoom };
+
+export default { ...meta, title: "Chat Room Editor" };
```
ryanwolhuter: When fixing the typescript build issues, I found that we were not actually type checking these story files correctly. I think it is still possible to export the default like this, but there is a config somewhere that would need to be done that I can't seem to find.

For now all that needs to change is adding ` as Meta<typeof Editor>;` to the export default, so like 

`export default { ...meta, title: "Chat Room Editor" } as Meta<typeof Editor>;` 

this lets typescript know that the default export which is not named is in fact using these storybook nested types.

packages/document-model-libs/editors/chat-room-editor/components/Message/Message.tsx
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable react/jsx-no-bind */
```
ryanwolhuter: these are supposed to just be warnings, but if they are annoying you then let's just disable them in the eslint config. they are just stylistic anyway.

packages/document-model-libs/editors/chat-room-editor/editor.tsx
```diff
@@ -0,0 +1,135 @@
+/* eslint-disable react/jsx-no-bind */
+import { EditorProps } from "document-model/document";
+import {
+  ChatRoomState,
+  ChatRoomAction,
+  ChatRoomLocalState,
+  ReactionType,
+  actions,
+} from "../../document-models/chat-room";
```
ryanwolhuter, line 9: with the typescript fix, you should now be able to just import these from `document-models/chat-room` instead of needing the relative path. This is most likely what your editor wants to do already

packages/document-model-libs/editors/chat-room-editor/index.ts
```diff
@@ -0,0 +1,25 @@
+import { ExtendedEditor, EditorContextProps } from "document-model-libs";
+import Editor from "./editor";
+import {
+  ChatRoomState,
+  ChatRoomAction,
+  ChatRoomLocalState,
+} from "../../document-models/chat-room";
```
ryanwolhuter, line 7: same for here about the import

ryanwolhuter (approved):

PR #657, opened 2024-12-10
fix(scalars): add the correct value type and send correct value in onSubmit

alejandrocabriales:
## Ticket
https://trello.com/c/0myyOw9N/771-test-the-components-already-created-in-the-document-editor

## Description
 The value in the NumberField should be number or bigInt when the value it's submitted


nesdc88 (approved):
LGTM

PR #665, opened 2024-12-10
fix(scalars): add the correct value type number or bigInt

alejandrocabriales:
## Ticket

## Description

- NumberField. numericType = NonPositiveFloat. Inserting e.g. - 0.90. Click Submit.  A float number should be allowed. The number is redounded to an integer and error message is shown: "Value must be a float".

- NumberField. numericType = NegativeInt. Inserting e.g. - 87. Click Submit.  The output should be a number.  The current output is a String. 

- NumberField. Removing the value. The placeholder should be shown. **Current Output:** The field displays value 0. even if the placeholder has been set.

- NumberField. Steps = 2. Using the hotkeys (arrows) . The values should increase/decrease with the steps provided.  **Current Output:** The values increase/decrease in steps of 1.

-  NumberField. precision = 2. Inserting e.g 0.9. **Expected Output:** The value should be allowed.An error message is shown: "Value must be a integer". 

- NumberField. Inserting values. **Expected Output:** The 0 should be removed and the value should be added.  The values are inserted after the 0. The only way to remove the 0 is adding a few numbers and move ack to the first digit and remove the 0.


packages/design-system/src/scalars/components/number-field/number-field.test.tsx
```diff
@@ -310,4 +310,153 @@ describe("NumberField", () => {
...
     expect(mockOnChange).not.toHaveBeenCalled();
   });
+
+  //New test for the issues
+  it("should handle NonPositiveFloat numeric type correctly", async () => {
+    const mockOnSubmit = vi.fn();
+    const user = userEvent.setup();
+
+    renderWithForm(
+      <Form onSubmit={mockOnSubmit}>
```
YasielCabrera, line 320: the `renderWithForm` method already wrap the components with a form, use `render` instead to wrap it yourself

packages/design-system/src/scalars/components/number-field/number-field.test.tsx
```diff
@@ -310,4 +310,153 @@ describe("NumberField", () => {
...
+        {({ formState: { isSubmitting } }) => (
+          <div className="flex w-[400px] flex-col gap-4">
```
YasielCabrera, line 322: in the tests you don't need to add styles to the form, just keep it simple

packages/design-system/src/scalars/components/types.ts
```diff
@@ -54,7 +54,8 @@ export type NumericType =
   | "PositiveFloat" // Positive float values
   | "NonNegativeFloat" // Non-negative float values (>= 0.0)
   | "NonPositiveFloat" // Non-positive float values (<= 0.0)
-  | "BigInt";
+  | "BigInt"
+  | undefined;
```
YasielCabrera: `undefined` should now be a numeric type, instead, you can add it as optional whenever you need it. Eg: `something: NumericType | undefined`

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/number-field/number-field.test.tsx
alejandrocabriales: you are right , done

packages/design-system/src/scalars/components/number-field/number-field.test.tsx
alejandrocabriales: done

packages/design-system/src/scalars/components/types.ts
alejandrocabriales: you are right, done

YasielCabrera (approved):
lgtm

PR #667, opened 2024-12-10
fix(scalars): avoid to convert bigInt to decimal notions

alejandrocabriales:
## Ticket

## Description
- Avoid to convert the number to decimal notations


nesdc88 (approved):
LGTM

PR #668, opened 2024-12-10
fix(scalars): allow a int pass as float and fix send bigInt when correct prop its set

alejandrocabriales:
## Ticked

## Description
- Allow to send a bigInt 
-  Allow to pass a int as float

### DIFF
--- a/packages/design-system/src/scalars/components/number-field/number-field-validations.ts
+++ b/packages/design-system/src/scalars/components/number-field/number-field-validations.ts
@@ -19,8 +19,10 @@ export const validateIsBigInt =
 export const isInteger = (value: unknown): boolean =>
   Number.isInteger(Number(value));
 
-export const isFloat = (value: unknown): boolean =>
-  !Number.isInteger(Number(value));
+export const isFloat = (value: unknown): boolean => {
+  const number = Number(value);
+  return !isNaN(number) && isFinite(number);
+};
 
 export const validateNumericType =
   ({ allowNegative = false, numericType }: NumberFieldProps) =>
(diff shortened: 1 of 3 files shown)


packages/design-system/src/scalars/components/number-field/number-field-validations.ts
```diff
@@ -31,67 +31,70 @@ export const validateNumericType =
...
-        if (!isFloat(value)) return "Value must be a float";
+        if (!isFloat(value) && !isInt)
+          return "Value must be a non-negative float or integer";
         if (!allowNegative && parseFloat(value as string) < 0)
-          return "Value must be a non-negative float";
+          return "Value must be a non-negative float or integer";
         if (parseFloat(value as string) >= 0) return true;
-        return "Value must be a non-negative float";
+        return "Value must be a non-negative float or integer";
       }
```
YasielCabrera: This looks to be the same of `PositiveFloat`, you can avoid duplication to put the two cases together

packages/design-system/src/scalars/components/number-field/number-field-validations.ts
```diff
@@ -31,67 +31,70 @@ export const validateNumericType =
...
+        if (!isInt) return "Value must be a non-positive integer";
         if (!allowNegative && Number(value) > 0)
           return "Value must be a non-positive integer";
         if (Number(value) <= 0) return true;
         return "Value must be a non-positive integer";
       }
 
       case "PositiveFloat": {
-        if (!isFloat(value)) return "Value must be a float";
+        if (!isFloat(value) && !isInt)
```
YasielCabrera: `!isFloat(value) && !isInt` this condition is wrong:

Here is the decalation of those functions:

```
export const isInteger = (value: unknown): boolean =>
  Number.isInteger(Number(value));
export const isFloat = (value: unknown): boolean =>
  !Number.isInteger(Number(value));
```

So if we analyze them logically: let's say
isInteger = `I`
isFloat = `F` = `¬I`

So you condition is `¬F ∧ ¬I` => `¬(¬I) ∧ ¬I` => `I  ∧ ¬I`.

This is always going to be false because you're saying it can't be an integer and need to be integer at the same time.

Take in count that for float numbers, you only need to check if it is positive or negative, no if they're int or float, because if they're valid numbers they're going to be float as Integers are also Float

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/number-field/number-field-validations.ts
alejandrocabriales: not applicable, for that case its handle different use cases

packages/design-system/src/scalars/components/number-field/number-field-validations.ts
alejandrocabriales: you are right , done

YasielCabrera (approved):
lgtm

PR #689, opened 2024-12-12
feat(reactor-api): added operational processor

froid1911:


packages/reactor-api/src/processors/operational-processor.ts
```diff
@@ -0,0 +1,22 @@
+import { IAnalyticsStore } from "@powerhousedao/analytics-engine-core";
```
acaldas: Unused import

packages/reactor-api/src/processors/operational-processor.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Document, OperationScope } from "document-model/document";
+import { Processor } from "./processor";
+import { ProcessorOptions, ProcessorSetupArgs } from "src/types";
+import { Knex } from "knex";
```
acaldas: Use the Db type from "utils/get-db-client" instead. Maybe we should rename that file to `utils/db`

PR #690, opened 2024-12-12
fix(scalars): remove blink effect in SelectField when searchable=true & searchPosition="Input"

nesdc88:
## Ticket
https://trello.com/c/FpOHFVJ0/760-11-countrycodefield

## Description
- remove blink effect in SelectField when searchable=true & searchPosition="Input"


alejandrocabriales (approved):
lgtm

PR #691, opened 2024-12-12
fix(scalars): add the placeholders and show bigInt in the alert

alejandrocabriales:
## Ticket

## Description
- Deleting the values.  The placeholder should be set by default like the label is to avoid showing the field empty. **Current Output:** The field is empty without value or placeholder. 

- NumberField. numericType = NonNegativeInt. Inserting e.g - 76 and submit. **Expected Output:** The value should be submitted. **Current Output:** Nothing happens, the values is not submitted.

### DIFF
--- a/packages/design-system/src/scalars/components/number-field/number-field.tsx
+++ b/packages/design-system/src/scalars/components/number-field/number-field.tsx
@@ -47,7 +47,7 @@ export const NumberFieldRaw = forwardRef<HTMLInputElement, NumberFieldProps>(
       precision = 0,
       ...props
     },
-    ref
+    ref,
   ) => {
     const generatedId = useId();
     const id = propId ?? generatedId;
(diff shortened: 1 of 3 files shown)


nesdc88 (approved):
LGTM

packages/design-system/src/scalars/components/number-field/number-field.tsx
```diff
@@ -185,6 +185,22 @@ export const NumberFieldRaw = forwardRef<HTMLInputElement, NumberFieldProps>(
...
       onBlur?.(e);
     };
+
+    // Avoid to write letters directly in safari
+    const handleInput = (event: React.KeyboardEvent<HTMLInputElement>) => {
+      const char = event.key;
+      // Allow numbers, decimal point, hyphen (for negatives) and control keys
+      if (
+        !/^[0-9.-]$/.test(char) && // Numbers, decimal point and hyphen
+        !["Backspace", "Delete", "ArrowLeft", "ArrowRight", "Tab"].includes(
```
YasielCabrera: Why do we need to prevent those keys, they should be necessary for a good UX

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/number-field/number-field.tsx
alejandrocabriales: Its need for the  safari, but i will check only for the letter, done

YasielCabrera (approved):

PR #701, opened 2024-12-13
feat(scalars): allow the tokens and improve code

alejandrocabriales:
## Ticket

## Description
- Should set **allowedConcurrecies** as an Array of strings.
- Should set **allowedTokens** as an Array of strings.


packages/design-system/src/scalars/components/amount-field/amount-field.stories.tsx
```diff
@@ -176,11 +218,13 @@ export const HoverWithCurrency: Story = {
     label: "Enter Amount and Select Currency",
     name: "amount",
     allowedCurrencies: ["USD", "EUR"],
+    currencyPosition: "right",
     type: "AmountCurrency",
     value: {
       amount: 345,
       currency: "USD",
     },
+    step: 0,
```
YasielCabrera, line 227: Why are we setting step 0 in all the stories? We should allow the default value on it to show the step showing the component as it is. If needed we can create a variant "without steps"

packages/design-system/src/scalars/components/amount-field/use-amount-field.ts
```diff
@@ -1,40 +1,137 @@
...
+    //Create the event
+    const nativeEvent = new Event("change", {
+      bubbles: true,
+      cancelable: true,
+    });
+    Object.defineProperty(nativeEvent, "target", {
+      value: { value: newValue },
+      writable: false,
+    });
+    onChange?.(newValue as unknown as React.ChangeEvent<HTMLInputElement>);
```
YasielCabrera: that case doesn't make sense there as we're passing the value instead of the event

packages/design-system/src/scalars/components/amount-field/use-amount-field.ts
```diff
@@ -1,40 +1,137 @@
...
```
YasielCabrera: similar comment that above, we're sending the value instead the event

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/amount-field/amount-field.stories.tsx
alejandrocabriales: Thank you for pointing this out! You are correct that the default value for step is 1, as documented in the component's documentation.

However, the reason for explicitly defining the step =0,  here is to align with the designs provided in Figma. In those designs, only one variant with a specific step value is represented, so we're making this explicit to ensure consistency with the design specifications.

packages/design-system/src/scalars/components/amount-field/use-amount-field.ts
alejandrocabriales: You're right—thank you for catching that. Done

packages/design-system/src/scalars/components/amount-field/use-amount-field.ts
alejandrocabriales: You're right—thank you for catching that. Done

YasielCabrera (approved):
lgtm

PR #703, opened 2024-12-14
feat(scalars): cast field values if needed before calling onSubmit

YasielCabrera:
## Ticket

## Description
Cast fields before the `onSubmit` is called. To cast values a `data-cast` attribute should be added to the input where the name attribute is set,


nesdc88 (approved):
LGTM

PR #705, opened 2024-12-15
feat: subgraph generator

froid1911:


packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-subgraph/index.esm.t
```diff
@@ -0,0 +1,43 @@
+---
+to: "<%= rootDir %>/<%= h.changeCase.param(name) %>/index.ts"
+force: true
+---
+import { Subgraph } from "@powerhousedao/reactor-api";
+import { Knex } from "knex";
```
acaldas: Knex is only used as a type here right?
To avoid adding it as a dependency in the boilerplate we could export the type from "reactor-api".
We could even rename it to Db:
`export { Knex as Db } from "knex";`

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-subgraph/index.esm.t
```diff
@@ -0,0 +1,43 @@
...
+import { gql } from "graphql-tag";
```
acaldas, line 6: This will require graphql-tag and graphql as dependencies in the boilerplate. Maybe the typedefs could just be defined as a string here?
The subgraph manager would parse it into typedefs.

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-subgraph/index.esm.t
```diff
@@ -0,0 +1,43 @@
...
+    }
+  }
+
+  typeDefs = gql`
+    type Query {
+      example(id: ID!): String
+    }
+  `;
+
+  additionalContextFields = {
```
acaldas, line 26: Feature request:
Add support to adding context per request. The subgraph would have access to an incoming request and could set the context dynamically.

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-subgraph/index.esm.t
```diff
@@ -0,0 +1,43 @@
...
+    example: "test"
+  }
+
+  async onSetup() {
+    await this.createOperationalTables(this.operationalStore);
+  }
+
+  async createOperationalTables(db: Knex) {
```
acaldas: Here it would become db: Db

packages/codegen/src/codegen/index.ts
```diff
@@ -188,3 +189,8 @@ export async function generateProcessor(
     { skipFormat },
   );
 }
+
+export async function generateSubgraph(name: string, config: PowerhouseConfig) {
+  console.log(name, config);
```
acaldas: stray log

packages/reactor-api/src/processors/analytics-processor.ts
```diff
@@ -13,11 +15,9 @@ export abstract class AnalyticsProcessor<
...
     super(args, options);
-    this.analyticsStore = args.dataSources.analyticsStore;
-  }
-
-  onSetup(args: ProcessorSetupArgs) {
-    super.onSetup(args);
-    this.analyticsStore = args.dataSources.analyticsStore;
+    this.analyticsStore = new KnexAnalyticsStore({
+      executor: new KnexQueryExecutor(),
+      knex: args.operationalStore,
```
acaldas: Not sure what this does. Why is the operational store passed to the analytics store?

Is this something that needs to be done by each processor or can just be done once in the reactor-api?

packages/reactor-api/src/processors/manager.ts
```diff
@@ -1,16 +1,16 @@
...
+import { IProcessor, IProcessorManager, ProcessorSetupArgs } from "../types";
 
 export class ProcessorManager implements IProcessorManager {
   private reactor: IDocumentDriveServer;
   private processors: IProcessor[] = [];
 
   constructor(
     driveServer: IDocumentDriveServer,
-    private analyticsStore: IAnalyticsStore,
+    private operationalStore: Knex,
```
acaldas: Knex -> Db :)
The less direct references we have to Knex in the code the better.

packages/reactor-api/src/server.ts
```diff
@@ -1,18 +1,11 @@
...
 import express, { Express } from "express";
 import { Pool } from "pg";
-import { ReactorRouterManager } from "./router";
-import { getKnexClient } from "./utils/get-knex-client";
-import { ProcessorManager } from "./processor-manager";
+import { ProcessorManager } from "./processors";
+import { SubgraphManager } from "./subgraphs/manager";
 import { API } from "./types";
-import { initialize } from "./subgraphs/analytics";
+import { getKnexClient } from "./utils/get-knex-client";
```
acaldas: Here could be getDbClient, and only `utils/db` "knows" about knex. If in the future we decide to use something else we just need to change `utils/db`.

packages/reactor-api/src/subgraphs/base/index.ts
```diff
@@ -0,0 +1,31 @@
+import { ISubgraph } from "../types";
+import { DocumentNode } from "graphql";
+import { IDocumentDriveServer } from "document-drive";
+import { SubgraphArgs } from "../types";
+import { GraphQLResolverMap } from "@apollo/subgraph/dist/schema-helper";
```
acaldas, line 5: Let's reexport GraphQLResolverMap from "types.ts" to have a simpler import instead of that @apollo dist path.

packages/reactor-api/src/subgraphs/base/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import { gql } from "graphql-tag";
+import { Knex } from "knex";
+import { Context } from "../types";
+
+export class Subgraph implements ISubgraph {
+  name = "example";
+  resolvers: GraphQLResolverMap<Context> = {
+    Query: {
+      hello: () => "world",
```
acaldas: Let's instead do a name query where it returns the subgraph name.

packages/reactor-api/src/subgraphs/drive/index.ts
```diff
@@ -1,4 +1,392 @@
...
+  DocumentModelState,
+} from "document-model/document-model";
+import { gql } from "graphql-tag";
+import { Context } from "../types";
+import { Subgraph } from "../base";
 
-export { typeDefs, resolvers };
+export class DriveSubgraph extends Subgraph {
+  name = "drive/:driveId";
+  typeDefs = gql`
```
acaldas, line 28: Let's define the drive typedefs and resolvers in different files. It will also make it easier to unit test the resolvers without having to instantiate a DriveSubgraph instance.

acaldas (commented):
Nice!

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-subgraph/index.esm.t
froid1911: I would argue, that if you decide to develop a subgraph you might accept that you have a graphql dependency... gql is a nice helper and supported by vscode graphql extension with syntax highlighting. if its just a string the dev dont get syntax higlighting :(

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-subgraph/index.esm.t
froid1911: nice idea, but lets do this in the next pr

PR #713, opened 2024-12-16
fix(scalars): allow trailingZeros in the ui

alejandrocabriales:
## Description
- NumberField. Submitting the value and clearing up the field.  If the field is not required, non-error message should be displayed. An error message is shown. 

- NumberField. precision = 2, trailingZeros = false. Inserting a number with two zeros (e.g. 34.00)**Expected Output:** The zeros should not be preserved, 34 should be displayed. The full number and the zeros are displayed with trailing zeros in true or false. The same result is displayed.


YasielCabrera (approved):
lgtm

PR #714, opened 2024-12-16
feat(scalars): add url platform icons

YasielCabrera:
## Ticket

## Description
Allow to show a built-in icon or a custom react element as icon.
Add `showWarnings` to show/hide URL warnings


packages/design-system/src/scalars/components/url-field/url-favicon.tsx
```diff
@@ -0,0 +1,47 @@
+import { IconName, iconNames } from "@/powerhouse";
+import { Icon } from "@/powerhouse/components/icon";
+// import { type PlatformIcon } from "./url-field";
```
alejandrocabriales: Is this comment necessary?

packages/design-system/src/scalars/components/url-field/url-favicon.tsx
```diff
@@ -0,0 +1,47 @@
...
+        }
+      }
+
+      return icon;
+    } catch {
+      return defaultIcon;
+    }
+  }, [url, platformIcons]);
+
+  if (Object.keys(platformIcons ?? {}).length === 0) return null;
```
alejandrocabriales, line 37: I think if we move this declaration
if (Object.keys(platformIcons ?? {}).length === 0) return null;
 up of the useMemo, we can avoid some calculations in the case that the conditions match

alejandrocabriales (changes requested):

alejandrocabriales:
LGTM

alejandrocabriales:
## The last changes were approved, but the tag was not updated correctly to reflect this.

## The two suggestions made were:

- Deleting a comment in the code.
- An optimization suggestion.
- The comment in the code was removed, and the optimization suggestion was deemed not applicable.

The pull request was merged by mistake without the "Approved" label.

PR #715, opened 2024-12-17
fix(scalars): add news validations allowNegative and add Int and Float

alejandrocabriales:
##  Ticket

## Description
- Should allowNegative Prop with PositiveInt: When numericType is set to PositiveInt, the allowNegative prop is redundant and should be ignored or disabled to prevent any conflicts. 

- Introducing Int and Float Scalars: Allow Int (Positive/Negative) and Float (Positive/Negative)


nesdc88 (approved):
LGTM

PR #716, opened 2024-12-17
fix(scalars): fix issues in StringField & update snapshots

nesdc88:
## Ticket

## Description
- ...The field should remain with the same heigh
- ...The field should not allow the break line
- ...The field should show the placeholder if there is no text inserted
- fix gaps
- improve min & max length behavior


alejandrocabriales:
LGTM

PR #719, opened 2024-12-17
fix(scalars): add hooks for number and rename some functions

alejandrocabriales:
## Ticket

## Description
- Should isBigInt Prop with Float Types: The isBigInt prop should only apply when the numeric type is an integer type. It should be ignored when the numeric type is set to a float type to ensure data consistency.

- Refactoring the hook


YasielCabrera (approved):
lgtm

PR #721, opened 2024-12-18
fix(scalars): trigger validation on blur in the select fields

YasielCabrera:
## Ticket
https://trello.com/c/FpOHFVJ0/760-11-countrycodefield

## Description
Trigger the validation onBlur on the select field and the country code field when the `showErrorsOnBlur` is enabled


alejandrocabriales:
LGTM

PR #730, opened 2024-12-18
fix(scalars): clean up unused props, correct storybook examples, and …

alejandrocabriales:
## Ticket

## Description
-  NumberField. precision = 2. Inserting any number (23, 34.456666) Expected Output: The field should allow only 2 decimals places.  Current Output: The field allows any number to be inserted.

- Connect. Required not set. numericType="NegativeInt". Leave the field empty and click on the Add button. Expected Output: The value submitted should be "null" or a message indicating no value has been submitted. Current Output: An error message appears: "Value must be a negative integer."

- Connect.  isBigtInt prop. Inserting a big number. Expected Output: The value should be submitted. Current Output: The value is submitted, but an error appears when you check the entry in the State section. 

- Connect.  Float field. Inserting 0.001. Expected Output: The value should be allowed as a positive float. Current Output: The value is not allowed and is redounded to 0.

- Float field. Insert a number (5.5) and enter. Expected Output: The value should be submitted as a float 5.5. Current Output: The value is redounded to 6 as an integer.


YasielCabrera (approved):
lgtm

PR #731, opened 2024-12-18
fix(scalars): number add the validations for safe int and float

alejandrocabriales:
## Ticket

## Description
-Connect.  Float field. Insert a number (5.5) and enter. The value should be submitted as a float 5.5.  The value is redounded 
  to 6 as an integer.


YasielCabrera (approved):
lgtm

PR #732, opened 2024-12-19
Reactor Browser

gpuente:


acaldas (commented):
It's good to move all the related code here to see what functionality is needed.
The atlas app should be a lot simpler as it just shows a single document.
We should try to integrate into Atlas as soon as possible, and then work on this library more to integrate in Connect. Apart from the drive initialization, I think we just need useDocument, useDocumentDispatch and addDebouncedOperations?

 
I think the goal in the future will be to move as most of the code as possible to be run outside of React. Then we just provide reactive queries to get state from the reactor, `getDocument` and mutations to add operations.
I've seen a lot of talk around TanStack Query: https://tanstack.com/query/latest

Ideally, the reactor is running on a WebWorker, which can be shared by all open tabs.

PR #733, opened 2024-12-19
fix(scalars): fix edge case of trim transformation (on Enter key down)

nesdc88:
## Ticket

## Description
- The space at the beginning and at the end should be removed


alejandrocabriales (approved):
LGTM

PR #744, opened 2025-01-06
fix(scalars): validate a number when not valid symbol arrive

alejandrocabriales:
## Ticket

## Description
- Storybook, Connect. numericType = PositiveInt. Inserting 3478098098-89899+8098.  **Expected Output:** An error message should be shown. **Current Output:** The Storybook submits the value and the alert shows it as "null". Connect: Clicking twice, the error message is displayed: "Value must be a positive integer" but the number is an integer with the signs '-' and '+' in the middle.

- Connect. Optional field. ShowErrorOnBlur = true. numericType = NegativeFloat. 1. Inserting a positive number 2. Add a negative number and submit it.  **Expected Output:** No error message should be visible after submitting the value. **Current Output:** Submitting the correct number and clicking outside the field, the error message related to the first wrong number is still visible even if the field is empty.


YasielCabrera (approved):
lgtm

PR #778, opened 2025-01-08
fix: change renown endpoint to auth.renown.id

froid1911:


packages/document-drive/src/server/listener/transmitter/subscription.ts
acaldas, line 1: Mixed file!

acaldas (approved):

packages/document-drive/src/server/listener/transmitter/subscription.ts
froid1911: w000ps

PR #792, opened 2025-01-08
feat(scalars): add a few validations and send correct value in submit

alejandrocabriales:
## Description
-  Should set **tokenIcons** as an Object.
- Should ensure that the field has a default value if none is provided.
- Should the field be non-editable if **ReadOnly (Disable)** is true.
-  Should Value represent the current value of the field.


YasielCabrera (approved):
lgtm

PR #798, opened 2025-01-09
feat(scalars): add the cast in form and trailingZeros and viewPrecision

alejandrocabriales:
## Ticket

## Description
- Should set **viewPrecision** as a number.
- Should set **trailingZeros** as a Boolean.
- Should ensure the field is not empty if required is true.
- Should the field be non-editable if **ReadOnly (Disable)** is true.
- Should display the error message as the user types if showErrorOnChange is true


YasielCabrera (approved):
lgtm

PR #800, opened 2025-01-09
fix(scalars): improve naming & behavior of some props in SelectField

nesdc88:
## Ticket
https://trello.com/c/zZlPyvEY/754-final-design-4-enum

## Description
- improve naming & behavior of some props in SelectField
- now selectionIconPosition="right" can only be specified if selectionIcon prop is "checkmark"


YasielCabrera (approved):
lgtm

PR #802, opened 2025-01-10
feat(reactor-browser): removed peerDependencies

gpuente:
## Description:

- removed peerDependencies


packages/reactor-browser/package.json
```diff
@@ -9,45 +9,46 @@
...
-    "@powerhousedao/design-system": "workspace:*",
-    "@powerhousedao/scalars": "workspace:*",
-    "document-drive": "workspace:*",
-    "document-model": "workspace:*",
-    "document-model-libs": "workspace:*",
     "react": "^18.3.1",
     "react-dom": "^18.3.1"
   },
   "dependencies": {
+    "@powerhousedao/config": "^1.5.0",
```
ryanwolhuter: I think these should still use the workspace:* protocol. putting them here in dependencies should still work the same, but the versions will be kept in sync

packages/reactor-browser/tsconfig.json
```diff
@@ -1,19 +1,27 @@
...
     "jsx": "react-jsx",
-    "outDir": "./dist",
-    "types": ["vitest/globals"],
-    "resolveJsonModule": true,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "paths": {
+      "document-drive": ["../document-drive/src/index.ts"],
+    }
   },
```
ryanwolhuter, line 25: doing this will make it include node_modules etc. I usually find its easier to maintain explicit includes than implicit includes with excludes, since you can accidentally add stuff that you meant to exclude and forget to update the array.

packages/reactor-browser/tsup.config.ts
```diff
@@ -0,0 +1,17 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+  entry: ["src/**/*.ts"],
+  splitting: true,
+  sourcemap: false,
+  clean: true,
+  format: ["esm", "cjs"],
```
ryanwolhuter: I think this is a good example of where the cjs can just be removed forever — we are never going to ship browser code that is meant to be cjs

packages/reactor-browser/tsup.config.ts
```diff
@@ -0,0 +1,17 @@
...
+  treeshake: true,
+  bundle: true,
+  platform: "node",
+  target: "node20",
```
ryanwolhuter, line 12: I see the target here is different from the one in the tsconfig — that could lead to inconsistencies in the build

gpuente: added fix here: [eb85db7](https://github.com/powerhouse-inc/powerhouse/pull/802/commits/eb85db7fda5a7e7f15fe6946cc81a6742fecef41)

packages/reactor-browser/tsup.config.ts
gpuente: added fix here: [eb85db7](https://github.com/powerhouse-inc/powerhouse/pull/802/commits/eb85db7fda5a7e7f15fe6946cc81a6742fecef41)

packages/reactor-browser/tsconfig.json
gpuente: added fix here: [eb85db7](https://github.com/powerhouse-inc/powerhouse/pull/802/commits/eb85db7fda5a7e7f15fe6946cc81a6742fecef41)

packages/reactor-browser/package.json
gpuente: added fix here: [eb85db7](https://github.com/powerhouse-inc/powerhouse/pull/802/commits/eb85db7fda5a7e7f15fe6946cc81a6742fecef41)

PR #803, opened 2025-01-10
feat(scalars): sidebar pin items

YasielCabrera:
## Ticket
https://trello.com/c/UBs2FHZ0

## Description
- Implemented item pinning
- When an item is pinned the root of nodes changed to the pinned item


alejandrocabriales (approved):
LGTM

PR #805, opened 2025-01-13
feat(scalars): add the viewPrecision and trailingZeros

alejandrocabriales:
## Ticket

## Description
 - Should set **numberPrecision** as a number.
 - Should the field allow the decimal places to be set in precision config

### DIFF
--- a/packages/design-system/src/scalars/components/amount-field/use-amount-field.tsx
+++ b/packages/design-system/src/scalars/components/amount-field/use-amount-field.tsx
@@ -7,6 +7,7 @@ import {
 } from "./types";
 import { getCountryCurrencies, getTokens } from "./utils";
 import { isValidNumber } from "../number-field/number-field-validations";
+import { useState } from "react";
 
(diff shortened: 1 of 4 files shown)


packages/design-system/src/scalars/components/amount-field/use-amount-field.tsx
```diff
@@ -148,40 +170,78 @@ export const useAmountField = ({
   };
 
   const handleBlur = (e: React.FocusEvent<HTMLInputElement>) => {
-    let newValue = {};
-    // Handle the blur of the input if type is AmountCurrency or AmountToken should be a bigint
+    setIsFocus(false);
+    const inputValue = e.target.value;
+    console.log("viewPrecision", viewPrecision);
```
nesdc88: please, remove this console.log

alejandrocabriales: done

nesdc88 (approved):
LGTM

PR #811, opened 2025-01-13
fix(scalars): add allowNegative and add validations along with number…

alejandrocabriales:
## Ticket

##Description
- Should the field show the decimals places set in viewPrecision, when the control is not focused.
- Should the field allow the numbers of decimals to be set in number precision.


YasielCabrera (approved):
lgtm

PR #812, opened 2025-01-13
feat:(scalars): add custom validation props fix style border

alejandrocabriales:
## Ticket

## Description
- Should the field show the numbers of decimals allowed only when the user has the input control focused.
- Should allow values till the minimum value set in the minValue.
- Should allow values till the maximum value set in the maxValue.
- Should the field allow negative numbers if allowNegative is true.
- Should the field allow the list of accepted currency codes set in the allowedConcurrencies config.


packages/design-system/src/scalars/components/amount-field/amount-field.stories.tsx
```diff
@@ -320,6 +332,17 @@ export const WithError: Story = {
...
 export const WithMultipleErrors: Story = {
   args: {
+    customValidators: (value: unknown) => {
+      if (
+        typeof value === "object" &&
+        value !== null &&
+        "amount" in value &&
+        value.amount !== "0"
+      ) {
+        return "Tiene que ser igual a cero";
```
YasielCabrera: Please, translate this message to english

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/amount-field/amount-field.stories.tsx
alejandrocabriales: done

YasielCabrera (approved):
lgtm

PR #814, opened 2025-01-14
feat(scalars): match server errors with fields after submission

YasielCabrera:
## Ticket
https://trello.com/c/X30k6NmB/769-fix-existing-component-issues

## Description
- Reconciliation of server errors with their corresponding field or show errors as form-level errors
- Added form prop to add custom matcher function which allows the dev to handle the error reconciliation


alejandrocabriales (approved):
LGTM

PR #827, opened 2025-01-15
fix(scalars): add ref for the amount-field and fix styles modified th…

alejandrocabriales:
## Ticket

## Description
 - The variant displays the number field without the currency option. 
 
- Adding a placeholder. The field should show the placeholder when no values have been added. The placeholder is not visible in the field. Please, make sure the placeholder appears when no values have been inserted. 

- Disabled status. The label should have color: Gray/700 (light) and Gray/300 (dark).  Currently the label have Gray/400 (light and Dark).  Please, apply the proper color for the label in the disable status.


YasielCabrera (approved):
lgtm

PR #828, opened 2025-01-15
fix(scalars): remove unnecesary stories and correct border for amount…

alejandrocabriales:
## Ticket

## Description
-  Default variant. The variant should contain the currency option. **Current Output:** The variant displays the number field without the currency option. 
- Adding a placeholder. The field should show the placeholder when no values have been added. **Current Output:** The placeholder is not visible in the field. 
- Disabled status.  The label should have color: Gray/700 (light) and Gray/300 (dark). Currently the label have Gray/400 (light and Dark).  Please, apply the proper color for the label in the disable status.


nesdc88 (approved):
LGTM

PR #829, opened 2025-01-15
fix(scalars): remove the opacity from select and update props descrip…

alejandrocabriales:
## Ticket

## Description
-  Disabled status. * The field should have the same styles in both section. The currency section is displayed with an opacity therefore it looks different that the input. 

 - Type prop.  The type prop should allow to select the type in a list. It is not possible to see the field with the different types in one variant. Navigate for all variants it's needed. 
 
 - Should the trailing zeros be preserved when viewPrecision is set, if trailingZeros is true. ( for example: 25.00,7.50, etc.) or not ( for example: 25, 7.5).


nesdc88 (approved):
Check the linting error in amount-field-validations.ts

YasielCabrera (approved):
lgtm

PR #830, opened 2025-01-15
fix(scalars): highlight selected option by default in Single Select

nesdc88:
## Ticket
https://trello.com/c/pWR6lA5L/796-fix-default-hover-behavior-in-the-single-select

## Description
- Should the hover be visible in the selected element by default when the dropdown is opened. (Single Select only)


YasielCabrera (approved):
lgtm

PR #831, opened 2025-01-15
feat(ph-cli): added install command

gpuente:
## Description:

- added install command

### DIFF
--- a/packages/ph-cli/src/commands/install.ts
+++ b/packages/ph-cli/src/commands/install.ts
@@ -3,14 +3,9 @@ import fs from "node:fs";
 import { execSync } from "node:child_process";
 import { homedir } from "node:os";
 import { Command } from "commander";
+import { PowerhouseConfig } from "@powerhousedao/config/powerhouse";
 
 import { CommandActionType } from "../types.js";
-
-type PowerhouseConfig = {
-  documentModelsDir: string;
-  editorsDir: string;
-  include?: string[];
-};
 
 const PH_BIN = "ph";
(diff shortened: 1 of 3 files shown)


packages/ph-cli/src/commands/install.ts
```diff
@@ -0,0 +1,323 @@
+import path, { dirname } from "node:path";
+import fs from "node:fs";
+import { execSync } from "node:child_process";
+import { homedir } from "node:os";
+import { Command } from "commander";
+
+import { CommandActionType } from "../types.js";
+
+type PowerhouseConfig = {
```
acaldas: We should update the type on @powerhousedao/config and import it here

packages/ph-cli/src/commands/install.ts
```diff
@@ -0,0 +1,323 @@
...
+  documentModelsDir: string;
+  editorsDir: string;
+  include?: string[];
```
acaldas: I think we should call this field packages or projects, not sure what term has been decided.

packages/ph-cli/src/commands/install.ts
gpuente: added fix here: [e797588](https://github.com/powerhouse-inc/powerhouse/pull/831/commits/e7975889b22389348f6702b076e8ad3e04513463)

packages/ph-cli/src/commands/install.ts
gpuente: added fix here: [e797588](https://github.com/powerhouse-inc/powerhouse/pull/831/commits/e7975889b22389348f6702b076e8ad3e04513463)

PR #833, opened 2025-01-15
feat(ph-cli): service manager

froid1911:


packages/ph-cli/powerhouse-services.config.json
```diff
@@ -0,0 +1,12 @@
+{
+  "apps": [
+    {
+      "name": "reactor",
+      "script": "dist/cli.js reactor"
```
acaldas: This only works when running inside the ph-cli project no?
Should it be `ph reactor` instead?

PR #835, opened 2025-01-16
fix(scalars): add more especific type in typescript and delete unnces…

alejandrocabriales:
## Ticket

## Description
Type prop. The type prop should allow you to select the type in a list.  It is impossible to see the field with the different types in one variant. Navigate for all variants it's needed.


nesdc88 (approved):
LGTM

PR #838, opened 2025-01-16
fix(scalars): rename and add stories to the amount

alejandrocabriales:
## Ticket

## Description
- The default state should represent the field without value in both sections. In the currency section should be an option when there is no option selected.

- Remove the unnecessaries stories 
- Add disabled variant 
- Add With Value variant (currently is the variant represented as Default) 
- Storybook: Update the Variants name with the names specified in the user story. 

## Notes:
- This pr will be affect to the number field behavior too

### DIFF
--- a/packages/design-system/src/scalars/components/amount-field/amount-field.stories.tsx
+++ b/packages/design-system/src/scalars/components/amount-field/amount-field.stories.tsx
@@ -177,6 +177,22 @@ type Story = StoryObj<typeof meta>;
 
 export const Default: Story = {
   args: {
+    placeholder: "0",
+    label: "Enter Amount and Select Currency",
+    name: "amount",
+    step: 0,
+    type: "AmountCurrencyFiat",
+    allowedCurrencies: ["USD", "EUR"],
+    currencyPosition: "right",
+    value: {
+      amount: undefined,
+      currency: "USD",
+    },
+  },
+};
+
+export const WithValue: Story = {
+  args: {
     placeholder: "Enter Amount",
     label: "Enter Amount and Select Currency",
     name: "amount",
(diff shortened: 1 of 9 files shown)


packages/design-system/src/scalars/components/amount-field/amount-field.stories.tsx
```diff
@@ -176,6 +176,22 @@ export default meta;
...
+    name: "amount",
+    step: 0,
+    type: "AmountCurrencyFiat",
+    allowedCurrencies: ["USD", "EUR"],
+    currencyPosition: "right",
+    value: {
+      amount: undefined,
+      currency: "USD",
+    },
+  },
```
YasielCabrera: Why do we need `step` and `currencyPosition` here? This is the default variant so it should be minimal args to represent the default usage of the component

packages/design-system/src/scalars/components/amount-field/amount-field.stories.tsx
```diff
@@ -185,13 +201,23 @@ export const Default: Story = {
...
-export const TokenIcon: Story = {
-  name: "Token Icon",
+export const WithAmount: Story = {
+  args: {
+    placeholder: "Enter Amount",
+    label: "EnterAmout ",
+    name: "amount",
+    type: "Amount",
+    value: 345,
+    step: 0,
```
YasielCabrera: similar here, why do we need step here?

YasielCabrera (changes requested):

YasielCabrera (approved):
lgtm

packages/design-system/src/scalars/components/amount-field/amount-field.stories.tsx
alejandrocabriales: done

packages/design-system/src/scalars/components/amount-field/amount-field.stories.tsx
alejandrocabriales: done

PR #842, opened 2025-01-19
fix(scalars): fix avoid transformations invalid number 

alejandrocabriales:
## Ticket

## Description
- Steps to Reproduce Precision = 2. Type = AmountCurrencyFiat. Entering a number (78.897) and click submit. The value should be rounded automatically to 78.90. **Current Output:** The alert shows the 78.9 without zero and in the field the value is not rounded until another click is done in the field.


packages/design-system/src/scalars/components/amount-field/use-amount-field.tsx
```diff
@@ -38,25 +39,56 @@ export const useAmountField = ({
...
+  const [formattedAmountState, setFormattedAmountState] = useState(
+    baseValue?.toString() ?? "",
+  );
+  // Funciona para datos compuestos
+  useEffect(() => {
+    setRawAmountState(baseValue === undefined ? "" : baseValue.toString());
+    setFormattedAmountState(
+      baseValue === undefined ? "" : baseValue.toString(),
+    );
+  }, [baseValue, type]);
```
YasielCabrera: why do we need a state that is going to be updated in an effect later? is it not better just create the vars and assign the value?

```
const rawAmountState = baseValue?.toString() ?? "";
const formattedAmountState = baseValue?.toString() ?? "";
```

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/amount-field/use-amount-field.tsx
alejandrocabriales: you are right , done

YasielCabrera (approved):
lgtm

PR #875, opened 2025-01-21
feat(monorepo): remove problematic peer deps

ryanwolhuter:
Currently we have several usages of peer dependencies that are pointless at best and harmful at worst.

Let's differentiate between some different types of dependencies:

- `dependencies` - these are the dependencies that are required for the package to work. Their code must be bundled with the package, or the package will either not build or break at runtime.

- `devDependencies` - these are the dependencies that are only needed for the package to build. They are not required for the package to work at runtime. _These are not included in the bundle, so they do not contribute to the bundle size._

The vast majority of dependencies should fall into one of the categories above.

- `peerDependencies` - these are the dependencies that _are_ required for the package to work. However, _we assume that the consumer of the package has already installed them_. An example is how `design-system` has `react` and `react-dom` as peer dependencies. We can safely assume that the consumer would not be using our components in an app that does not have `react` and `react-dom` installed.

Peer dependencies, like dev dependencies, are not included in the bundle. This has lead to a lot of issues where people are trying to use peer dependencies to reduce bundle size. This is a mistake in most cases.

First example:

Putting dev dependencies in peer dependencies is a mistake. Here in `scalars`, we have `zod` and `graphql` as both dev and peer dependencies.

```json
 "devDependencies": {
    "glob": "^11.0.0",
    "graphql": "^16.9.0",
    "zod": "^3.23.8"
  },
  "peerDependencies": {
    "graphql": "^16.9.0",
    "zod": "^3.23.8"
  },
```

- this is pointless because since they are already dev dependencies, they are _already not included in the bundle_.

- this tells that package manager to complain if the consumer does not have these packages installed, _but they don't even need to have them because they are dev dependencies_.

Second example:

Putting our own packages in that are needed for our packages to work in peer dependencies is a mistake. Here in `reactor-browser`, we have several of our own packages as peer dependencies.

```json
"peerDependencies": {
    "@powerhousedao/config": "workspace:*",
    "@powerhousedao/design-system": "workspace:*",
    "@powerhousedao/scalars": "workspace:*",
    "document-drive": "workspace:*",
    "document-model": "workspace:*",
    "document-model-libs": "workspace:*",
  },
```

There are actually multiple issues here:

```json
    "@powerhousedao/config": "workspace:*",
    "@powerhousedao/design-system": "workspace:*",
    "@powerhousedao/scalars": "workspace:*",
    ```

These dependencies are in fact only used as dev dependencies in the `renown` package. This tells the user they need to install these packages, even though our package is not even going to use them.

```json
    "document-drive": "workspace:*",
    "document-model": "workspace:*",
    "document-model-libs": "workspace:*",
```

These packages have code that the package can't run without. They should be dependencies. When the package is bundled, the code from these packages _that is actually used_ will be included in the bundle, with the rest being ignored. By making these peer dependencies, we are making the `reactor-browser` a tiny bit smaller by not including this code. But then we are making the consumer install _the entire package_, even if we only actually need one line of code from it.

In my opinion we are optimizing our library code in the wrong way. We are packaging and bundling code that is intended to be consumed by an app, which will have its own bundling process. Our bundling / optimizing is maybe making our npm packages a bit smaller, but at the cost of making the code harder to tree shake and bundle efficiently for the consumer. This makes the _consumer's bundler heavier_, which is the opposite of what we are trying to achieve.

I believe that removing these problematic peer dependencies is a good first step, especially since it will fix a lot of the build issues that have caused problems when implementing `ph-cmd`.

### DIFF
--- a/packages/document-drive/package.json
+++ b/packages/document-drive/package.json
@@ -1,6 +1,6 @@
 {
   "name": "document-drive",
-  "version": "1.13.5",
+  "version": "1.13.2",
   "license": "AGPL-3.0-only",
   "type": "module",
   "module": "./src/index.ts",
@@ -35,18 +35,6 @@
     "test:watch": "vitest watch",
     "clean": "rimraf dist",
     "clean:node_modules": "rimraf node_modules"
-  },
-  "peerDependencies": {
-    "@powerhousedao/scalars": "workspace:*",
-    "change-case": "^5.4.4",
-    "document-model": "workspace:*",
-    "document-model-libs": "workspace:*",
-    "graphql": "^16.9.0",
-    "graphql-request": "^6.1.0",
-    "json-stringify-deterministic": "^1.0.12",
-    "nanoevents": "^9.0.0",
-    "prisma": "^5.18.0",
-    "sanitize-filename": "^1.6.3"
   },
   "optionalDependencies": {
     "@powerhousedao/scalars": "workspace:*",
--- a/packages/document-drive/package.json
+++ b/packages/document-drive/package.json
@@ -1,6 +1,6 @@
 {
   "name": "document-drive",
-  "version": "1.13.2",
+  "version": "1.13.5",
   "license": "AGPL-3.0-only",
   "type": "module",
   "module": "./src/index.ts",
--- a/packages/reactor-browser/package.json
+++ b/packages/reactor-browser/package.json
@@ -29,26 +29,20 @@
     "@powerhousedao/config": "workspace:*",
     "@powerhousedao/design-system": "workspace:*",
     "@powerhousedao/scalars": "workspace:*",
-    "document-drive": "workspace:*",
-    "document-model": "workspace:*",
-    "document-model-libs": "workspace:*",
     "@types/react": "^18.3.1",
     "@types/react-dom": "^18.3.1",
     "react": "^18.3.1",
     "react-dom": "^18.3.1"
   },
   "peerDependencies": {
-    "@powerhousedao/config": "workspace:*",
-    "@powerhousedao/design-system": "workspace:*",
-    "@powerhousedao/scalars": "workspace:*",
-    "document-drive": "workspace:*",
-    "document-model": "workspace:*",
-    "document-model-libs": "workspace:*",
     "react": "^18.3.1",
     "react-dom": "^18.3.1"
   },
   "dependencies": {
     "did-key-creator": "^1.2.0",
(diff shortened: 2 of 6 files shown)


froid1911 (approved):

packages/document-drive/package.json
```diff
@@ -36,18 +36,6 @@
     "clean": "rimraf dist",
     "clean:node_modules": "rimraf node_modules"
   },
-  "peerDependencies": {
```
acaldas, line 39: These are needed because the document-drive code will be bundled by the packages that import "document-drive". They need to be able to resolve imports from these dependencies. Not sure if there is a better way to do this.

packages/reactor-browser/package.json
```diff
@@ -29,26 +29,20 @@
...
     "react": "^18.3.1",
     "react-dom": "^18.3.1"
   },
   "dependencies": {
     "did-key-creator": "^1.2.0",
-    "jotai": "^2.10.3"
+    "jotai": "^2.10.3",
+    "document-model": "workspace:*",
+    "document-model-libs": "workspace:*",
+    "document-drive": "workspace:*"
```
acaldas, line 46: "document-drive" is a ts only package with no build step, I think that "reactor-browser" needs to bundle it.
Currently all dependencies are excluded from the bundle.

packages/renown/package.json
```diff
@@ -37,9 +37,6 @@
     "clean": "rimraf dist/",
     "clean:node_modules": "rimraf node_modules"
   },
-  "peerDependencies": {
-    "document-model": "workspace:*"
```
acaldas, line 41: Renown is also a ts-only package, it only imports the `User` type from "document-model". It probably needs to change the bundle to include the type definition.

packages/scalars/package.json
```diff
@@ -28,10 +28,6 @@
     "graphql": "^16.9.0",
     "zod": "^3.23.8"
   },
-  "peerDependencies": {
```
acaldas, line 31: Not sure why these are peerDependencies.
There is a nasty issue that has come up in the past due to having different versions of graphql being loaded: https://stackoverflow.com/questions/70918391/duplicate-graphql-modules-cannot-be-used-at-the-same-time
Just to keep in mind in case this error reappears

acaldas (approved):
I'm unsure what's the best way to deal with the ts only packages (renown and document-drive).
Where you able to test if `mips-portal-next` still works with these changes?

PR #890, opened 2025-01-21
fix(scalars): avoid to convert number and fix some validation

alejandrocabriales:
## Ticket

## Description
- Fix some wrong behavior relate to the change in the reset
- Avoid converting to scientific notation


YasielCabrera (approved):
lgtm

PR #891, opened 2025-01-21
fix(scalars): add placeholder for when options dont match with value

alejandrocabriales:
## Description
- Steps to ReproduceType = AmountCurrencyCrypto, allowedTokens = {ETH}.  **Expected Output:** The currency crypto should be shown properly in the dropdown.  **Current Output:** The currency displayed in the dropdown is cut off at the right side. It is barely visible.


YasielCabrera (approved):
lgtm

PR #895, opened 2025-01-22
feat(scalars): improve dropdown selected state in PHID field

nesdc88:
## Ticket

## Description
- Selected (3 enums variants) (in progress)
- Improve code estructure

### DIFF
--- a/packages/design-system/src/scalars/components/phid-field/use-phid-field.ts
+++ b/packages/design-system/src/scalars/components/phid-field/use-phid-field.ts
@@ -1,6 +1,6 @@
 import { useCallback, useState, useEffect, useRef } from "react";
 import type { PHIDProps, PHIDListItemProps } from "./types";
-import { mockFetchPHIDOptions } from "./utils";
+import { fetchPHIDOptions } from "./utils";
 
 interface UsePHIDFieldProps {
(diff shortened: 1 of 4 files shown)


packages/design-system/src/scalars/components/phid-field/use-phid-field.ts
```diff
@@ -17,6 +18,25 @@ export function usePHIDField({
...
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    const fetchOptions = async () => {
+      setIsLoading(true);
+      try {
+        const options = await mockFetchPHIDOptions();
+        setOptions(options);
+      } catch (error) {
+        console.error("Error fetching PHID options:", error);
```
YasielCabrera: we should handle the error state instead of printing out the error

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/phid-field/use-phid-field.ts
nesdc88: Done

YasielCabrera (approved):
lgtm

PR #900, opened 2025-01-22
Updates for analytics processors

thegoldenmule:
- Updated the template
- Updated the rwa-analytics processor
- Updated some setup


packages/ph-cli/src/commands/reactor.ts
```diff
@@ -22,7 +22,12 @@ export const DefaultReactorOptions = {
 
 async function startLocalReactor(reactorOptions: ReactorOptions) {
   const baseConfig = getConfig();
-  const options = { ...DefaultReactorOptions, ...reactorOptions };
+  console.log("baseConfig", baseConfig);
```
froid1911, line 25: lets remove this

packages/reactor-local/src/index.ts
```diff
@@ -2,6 +2,6 @@ import { startServer } from "./server";
 
 export * from "./server";
 
-startServer({ dev: true }).catch((error: unknown) => {
+startServer({ dev: true, dbPath: "postgresql://postgres:password@localhost:5555/analytics" }).catch((error: unknown) => {
```
froid1911, line 5: do we want, that the local reactor requires a local postgres installation by default? I think thats going to break a lot of things.

packages/reactor-local/processors/index.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * This is a scaffold file meant for customization.
+ * Delete the file and run the code generator again to have it reset
+ */
+
+export * as RwaAnalyticsProcessor from "./rwa-analytics";
```
froid1911, line 6: doe we want the rwa analytics processor as default package for all local reactor installations?

froid1911 (changes requested):

froid1911:
@thegoldenmule do we want to close this?

thegoldenmule:
Sorry @froid1911 just saw this. I will check it out tomorrow.

packages/reactor-local/src/index.ts
thegoldenmule: Yah, probably.

PR #906, opened 2025-01-22
fix(scalars): check that type match with value

alejandrocabriales:
## Ticket

## Description

- Step to reproduce Type = AmountCurrencyFiat.  **Expected Output:** The Generics and specifics props should be displayed.  **Current Output:** The CurrencySymbol prop is not visible.  Please, add the CurrencySymbol prop to the proper type.

- Step to reproduce Type = AmountCurrencyCrypto. Changing the type to Amount or Percentage. **Expected Output:** The field should represent the proper type. **Current Output:** The field is displayed with "[object Object]" inside.


YasielCabrera (approved):
lgtm

PR #909, opened 2025-01-22
feat(scalars): change the root where the search work when elements are pinned

YasielCabrera:
## Ticket

## Description
Search over the root elements and those elements now change when an element is pinned


nesdc88 (approved):
LGTM

PR #910, opened 2025-01-22
feat(scalars): improve autoComplete behavior on PHID field

nesdc88:
## Ticket

## Description
- Should the field be represented as an input when **autoComplete** is false
- Should the field allow to select an item from the list when **autoComplete** is true
- Should the **variant** prop not be visible when **autoComplete** is false
- Select the input text only if not have the focus
- Position the dropdown below the input


alejandrocabriales (approved):
LGTM

PR #930, opened 2025-01-27
fix(ph-cli): improved setup script

froid1911:


acaldas (approved):
Nice!

Should we move to node 22? It's the default version currently.

PR #932, opened 2025-01-27
feat(reactor-api): added https support

froid1911:


packages/reactor-api/src/server.ts
```diff
@@ -10,12 +10,18 @@ import {
   KnexAnalyticsStore,
   KnexQueryExecutor,
 } from "@powerhousedao/analytics-engine-knex";
+import https from "https";
+import fs from "fs";
```
acaldas: Let's use the "node:" prefix for the core modules, "node:https" and "node:fs"

packages/reactor-api/src/server.ts
```diff
@@ -10,12 +10,18 @@ import {
...
 
 type Options = {
   express?: Express;
   port?: number;
   dbPath: string | undefined;
   client?: PGlite | typeof Pool | undefined;
+  ssl?: {
```
acaldas: Should we call it `https` to keep it consistent with Connect and Vite dev server?

packages/ph-cli/src/commands/dev.ts
```diff
@@ -117,6 +129,8 @@ export function devCommand(program: Command) {
     .description("Starts dev environment")
     .option("--generate", "generate code when document model is updated")
     .option("--reactor-port <port>", "port to use for the reactor")
+    .option("--ssl-key-file <SSL_KEY_FILE>", "path to the ssl key file")
+    .option("--ssl-cert-file <SSL_CERT_FILE>", "path to the ssl cert file")
```
acaldas: `--https-key` and `--https-cert` here?

packages/ph-cli/src/commands/reactor.ts
```diff
@@ -29,7 +31,15 @@ async function startLocalReactor(reactorOptions: ReactorOptions) {
     ...reactorOptions,
   };
 
-  const reactor = await startServer(options);
+  const ssl =
+    reactorOptions.sslKeyFile && reactorOptions.sslCertFile
+      ? {
+          keyPath: reactorOptions.sslKeyFile,
```
acaldas: I'm not sure how this will handle relative imports.
Let's verify it after this is published.

acaldas (approved):

PR #933, opened 2025-01-27
fix(scalars): build on nextjs

YasielCabrera:
## Ticket

## Description
Add `use client` directive to import file so we make all the components in the library client side as all of them depend on client-side elements


nesdc88 (approved):
LGTM

PR #934, opened 2025-01-27
fix(ph-cmd): inject ph-cmd version in build time

gpuente:
## Description:

- Inject app version


acaldas (commented):
How does this work if the semantic versioning is only applied after the build?

gpuente:
> How does this work if the semantic versioning is only applied after the build?

hey @acaldas! this package has a `prepublish` build script, this one runs after the semantic versioning 👍🏻

acaldas (approved):

PR #953, opened 2025-01-28
feat(document-model-libs): remove gen from gitignore

ryanwolhuter:
We currently have the generated code for document models ignored by git. We have had feedback from people developing document models that this causes a lot of problems:

- it encourages developers not to look at the generated code, making the mechanics feel obscure
- it allows differing business logic code to exist on developers' machines without them knowing
- it forces us to perform a generate step every time we do a build
- it prevents developers from seeing the full implications of the changes to their models (which might break things unexpectedly)
- it allows developers to make unrecoverable mistakes

The current approach of ignoring this code is treating the generated code like distribution code, which is usually ignored by git, similar to a bundler's build output for example. However, I would argue that this is not the correct approach for our generated code. Our generators output code which is itself business logic, and will be included in the distribution code. Therefore, developers do need to have awareness of the generated code and approve any changes to it, since this can and will change business logic.

A good example of generated code that is mostly meant to be left as-is and not edited directly are the lockfiles our package managers create, e.g. package-lock.json. This code is auto-generated and never intended to be edited directly, and should in theory be a direct reflection of the package.json. However, we do still always commit this file to git, because if the effects and side effects of the packages results in differing output, especially when this is not expected, then we want to know about it and approve it.

By removing /gen from .gitignore, we get these key benefits:

- developers must look at and agree with the generated code for new or changed document models, making sure that they understand what their code does
- once code is generated, it will be the same on every developer's machine who clones the repo
- we only need to generate code ad-hoc when the actual document model is changed, meaning that we don't need to generate every time a build happens (especially significant in ci environments)
- when a developer changes the document model and runs a new generate, they will immediately see all of the implications of their change to the generated code. if their change has an unintended effect, they can always roll back and do the generate again
- if a developer makes a mistake that breaks their document model and generation no longer works, they can roll back both their change and also the generated code to a working state


liberuum:
Yes, I agree. I was wondering why we had that ignored. LGTM

acaldas (approved):
The added visibility is interesting, should make it easier for the dev to understand the impact of a change in the document model.

This PR doesn't change when the code generation runs right? So it might happen that the published code is different from the generated code in the repo.

Let's try this way and see how it works out.

We could make this an option when generating the project.

ryanwolhuter:
> The added visibility is interesting, should make it easier for the dev to understand the impact of a change in the document model.
> 
> This PR doesn't change when the code generation runs right? So it might happen that the published code is different from the generated code in the repo.
> 
> Let's try this way and see how it works out.
> 
> We could make this an option when generating the project.

sweet yeah, I do think this should be the default behavior though. I went and checked what graphql-codegen recommends, and they don't ignore generated files.

this PR does change when the code generation runs. Before it was running on build and when running storybook, which is now removed. I have run generate and committed that generated code. the only way for the output of generate to change now is for someone to alter one of the document models. 

If they do this and do not run generate, they will get an error. This means that you can't publish this code, because the build will fail.

The only way for published generated code to not match the generated code in the repo is for a developer to manually publish generated code which they can see in their git changes but have chosen not to commit. Since we are using a CI-based flow, this should not happen.

NicWickman:
LGTM. This also prevents any changes to codegen or a myriad of other places from retroactively breaking things which are (supposedly) final.

froid1911 (approved):
LFG!

PR #958, opened 2025-01-28
Migrate connect

gpuente:
## Description:

- migrate connect to monorepo


acaldas (commented):
These 1200 commits will be added to the changelog of all the packages right?
I think we should run a script to update the title of the commits to be `feat(connect)` and so on.
For all commits that don't respect semantic versioning, like `Merge pull request ...` we can add a `chore(connect): `

froid1911:
The question is here whether the entire history really brings a benefit or whether we simply should squash all the commits to "feat: added connect".  I doubt that someone will look for a specific commit in the past.

acaldas:
> The question is here whether the entire history really brings a benefit or whether we simply should squash all the commits to "feat: added connect". I doubt that someone will look for a specific commit in the past.

We have been trying to keep the git history when importing projects to the monorepo because it can be very helpful when bugs come up. Especially when checking regressions.
If chatgpt can come up with a script I think it would be worth it.

gpuente:
> These 1200 commits will be added to the changelog of all the packages right? I think we should run a script to update the title of the commits to be `feat(connect)` and so on. For all commits that don't respect semantic versioning, like `Merge pull request ...` we can add a `chore(connect): `

hey @acaldas 
The semantic versioning uses a combination of the files changed and the commit messages:

- commit messages without semantic structure: this changes are ignored, they don't change the version and changelog of the projects affected (similar to chore)
- commit messages without target (`feat: commit message`): Updates version and changelog only for projects affected by the commit

if we want to change the commit messages for connect, I'll need to run the script in connect first, and then do the migration again :/

acaldas:
> > These 1200 commits will be added to the changelog of all the packages right? I think we should run a script to update the title of the commits to be `feat(connect)` and so on. For all commits that don't respect semantic versioning, like `Merge pull request ...` we can add a `chore(connect): `
> 
> hey @acaldas The semantic versioning uses a combination of the files changed and the commit messages:
> 
> * commit messages without semantic structure: this changes are ignored, they don't change the version and changelog of the projects affected (similar to chore)
> * commit messages without target (`feat: commit message`): Updates version and changelog only for projects affected by the commit
> 
> if we want to change the commit messages for connect, I'll need to run the script in connect first, and then do the migration again :/

So these commits will only go to the Connect changelog?
If so that's great. Otherwise the script could run in this branch, before merging to main.

gpuente:
> So these commits will only go to the Connect changelog? If so that's great. Otherwise the script could run in this branch, before merging to main.

yes, I ran a test locally and only connect will be affected

PR #960, opened 2025-01-29
fix(scalars): add each icon and fix the test mocking the icon

alejandrocabriales:
## Ticket

## Description
- Should set the **Description** as a String.
-  Should set the **required** field as a Boolean (true/false).
- Should set the **errors** as an Array.
- Should set the **Value** as a String. True, false (or null if the field is optional)


YasielCabrera (approved):
lgtm

PR #961, opened 2025-01-29
feat(scalars): improve item hover and search

YasielCabrera:
## Ticket

## Description
Hover pin icon when unpinning and move to the next element when pressing enter. Highlight active search result

### DIFF
--- a/packages/design-system/src/scalars/components/sidebar/subcomponents/sidebar-item.tsx
+++ b/packages/design-system/src/scalars/components/sidebar/subcomponents/sidebar-item.tsx
@@ -56,6 +56,7 @@ export const Item: React.FC<ItemProps> = ({
             "group/sidebar-item dark:hover:bg-charcoal-900 relative flex cursor-pointer select-none items-center justify-between gap-2 rounded-md px-2 py-1.5 text-gray-700 hover:bg-gray-100 dark:text-gray-400",
             allowPinning && "hover:pr-6",
             isPinned && "pr-6",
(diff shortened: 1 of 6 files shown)


packages/design-system/src/scalars/components/sidebar/subcomponents/sidebar-item.tsx
```diff
@@ -56,6 +56,7 @@ export const Item: React.FC<ItemProps> = ({
             "group/sidebar-item dark:hover:bg-charcoal-900 relative flex cursor-pointer select-none items-center justify-between gap-2 rounded-md px-2 py-1.5 text-gray-700 hover:bg-gray-100 dark:text-gray-400",
             allowPinning && "hover:pr-6",
             isPinned && "pr-6",
+            isSearchActive && "bg-yellow-100 dark:bg-[#604B0033]",
```
alejandrocabriales, line 59: Can we move on to the color palette?

alejandrocabriales (changes requested):

packages/design-system/src/scalars/components/sidebar/subcomponents/sidebar-item.tsx
YasielCabrera: no, we can't. That's colors used in the design that are not part of the color palette

alejandrocabriales: perfect, then

alejandrocabriales (approved):
LGTM

PR #977, opened 2025-01-31
feat(scalars): add the calendary monthly for light mode

alejandrocabriales:
## Ticket

## Description
- Open Month

### DIFF
--- a/packages/design-system/src/scalars/components/date-picker-field/use-date-picker-field.tsx
+++ b/packages/design-system/src/scalars/components/date-picker-field/use-date-picker-field.tsx
@@ -1,0 +1,54 @@
+import React from "react";
+import { DateFieldValue } from "./types";
+import { format, isValid } from "date-fns";
+
+interface DatePickerFieldProps {
+  value?: DateFieldValue;
+  defaultValue?: DateFieldValue;
+}
+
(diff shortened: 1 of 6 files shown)


packages/design-system/src/scalars/components/date-picker-field/use-date-picker-field.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    if (typeof value === "string") {
+      return value;
+    }
+
+    if (typeof value === "object" && isValid(value)) {
+      return format(value, "dd/MM/yyyy");
+    }
+    return "";
+  };
```
YasielCabrera: you should arrow function or lambda function but no both styles or it could be confusing. I also recommend wrapping those functions in an `useCalback` as they're inside a hook or could be dependencies, this way we save some renders

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/date-picker-field/use-date-picker-field.tsx
alejandrocabriales: You are right done

YasielCabrera (approved):
lgtm

PR #978, opened 2025-01-31
feat(scalars): make the API call when the user stops typing & update the API call functions

nesdc88:
## Ticket

## Description
- Make the API call when the user stops typing
- Should update the fetchPHIDOptions function
- Should create the fetchSelectedOption function


YasielCabrera (approved):
lgtm

PR #999, opened 2025-02-05
 feat(scalars): add default value for date and time

alejandrocabriales:
## Ticket

## Description
- Should set the **Default** as a String. True, false (or null if the field is optional)


YasielCabrera (approved):
lgtm

PR #1000, opened 2025-02-05
Feat/hover pinned items

YasielCabrera:
## Ticket

## Description
- Hover items on pinned items
- take the whole node on the onNodeChange callback instead just the node Id


packages/design-system/src/scalars/components/sidebar/sidebar.stories.tsx
```diff
@@ -106,8 +106,8 @@ const meta: Meta<typeof Sidebar> = {
       </div>
     ),
     enableMacros: 4,
-    onActiveNodeChange: (nodeId) => {
-      console.log("onActiveNodeChange", nodeId);
+    onActiveNodeChange: (node) => {
+      console.log("onActiveNodeChange", node);
```
alejandrocabriales, line 110: Is this console.log necessary?

alejandrocabriales (changes requested):

packages/design-system/src/scalars/components/sidebar/sidebar.stories.tsx
YasielCabrera: Note that this is in a story, it is for testing on the storybook

alejandrocabriales: nice, thanks for the feedback

alejandrocabriales (approved):
LGTM

PR #1002, opened 2025-02-06
feat(scalars): adds test cases and behavior improvements to the PHID field

nesdc88:
## Ticket

## Description
- Add test cases
- Should autoselect only when the phid match exactly
- Render only an Input when autoComplete is false
- Improve copy/paste support
- Minor improvements


alejandrocabriales (approved):
LGTM

PR #1018, opened 2025-02-07
fix(scalars): should the Textarea field resize with its container

nesdc88:
## Ticket
https://trello.com/c/acdZMNJh/823-fix-auto-expand-text-field-not-resizing-with-its-container

## Description
- Should the auto-expand field resize with its container.


alejandrocabriales (approved):
LGTM

PR #1037, opened 2025-02-10
Feat/add ensure the field validation date

alejandrocabriales:
## Ticket

## Description
- Should ensure the field is not empty if required is true
- Should ensure the field can be set as true/false/not set if required is false. Only true/false if required is true.


YasielCabrera (approved):
lgtm

PR #1043, opened 2025-02-10
fix(scalars): should clear the selected option when clearing the form & minor improvements

nesdc88:
## Ticket

## Description
- Should clear the selected option when clearing the form

### DIFF
--- a/packages/design-system/src/scalars/components/phid-field/phid-field.test.tsx
+++ b/packages/design-system/src/scalars/components/phid-field/phid-field.test.tsx
@@ -37,8 +37,7 @@ describe("PHIDField Component", () => {
   const defaultGetOptions = vi.fn().mockResolvedValue(mockOptions);
   const defaultGetSelectedOption = vi
     .fn()
-    .mockImplementation(async (phid: string) => {
-      await Promise.resolve();
+    .mockImplementation((phid: string) => {
       return mockOptions.find((option) => option.phid === phid);
     });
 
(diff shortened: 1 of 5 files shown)


packages/design-system/src/scalars/components/phid-field/phid-field.test.tsx
```diff
@@ -34,19 +34,40 @@ describe("PHIDField Component", () => {
     },
   ];
 
+  const defaultGetOptions = vi.fn().mockResolvedValue(mockOptions);
+  const defaultGetSelectedOption = vi
+    .fn()
+    .mockImplementation(async (phid: string) => {
+      await Promise.resolve();
```
YasielCabrera: This line doesn't do anything here

packages/design-system/src/scalars/components/phid-field/phid-field.tsx
```diff
@@ -126,7 +124,7 @@ const PHIDFieldRaw = React.forwardRef<HTMLInputElement, PHIDFieldProps>(
             {label}
           </FormLabel>
         )}
-        {autoComplete ? (
+        {autoComplete && fetchOptionsCallback && fetchSelectedOptionCallback ? (
```
YasielCabrera: the `fetchSelectedOptionCallback` should not determine if the input can have autocomplete or not, it should be used to refresh the selected item data only. With `autoComplete && fetchOptionsCallback` condition should be enough here

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/phid-field/phid-field.test.tsx
nesdc88: Removed

packages/design-system/src/scalars/components/phid-field/phid-field.tsx
nesdc88: Done

YasielCabrera (approved):

PR #1045, opened 2025-02-10
document-drive incremental refactor

thegoldenmule:
Items:

- Pulled synchronization logic out of graphql resolvers and into standalone functions.
- Added a benchmark for importing Andromeda data into a document-drive (no GQL).
- Got `document-drive` workflow humming: README, docker-compose for testing, package.json scripts, etc.
- Starting to iron out OOP issues:
  - Using `interface` instead of `abstract` class where appropriate.
  - Pulling dependencies up the chain, like listeners, triggers, listener manager.
  - Pulled objects into constructor (so we can later inject).
  - Added some factories for better abstraction of concrete implementations.
  - Hit some dead-ends for now with circular dependencies.
- Various smaller refactors, like renaming `drive` -> `driveId` where appropriate.

Testing notes:

- All document-drive tests pass.
- `pnpm build:all` builds successfully in root monorepo directory.
- `pnpm test:all` fails in `main`, and fails with the same tests in this branch.


packages/reactor-api/test/benchmarks/sync.bench.ts
```diff
@@ -0,0 +1,151 @@
...
+        {
+          documentType: document.documentType,
+          id: documentId,
+          name: "BlocktowerAndromeda",
+        },
+        ["global"],
+      );
+      await serverA.addDriveAction(driveAId, addFileAction);
+
+      const result = await serverA.addOperations(
```
acaldas, line 131: The drives are not syncing here yet.

This bench seems to get stuck, or it just takes way too long.
It seems to be related to having the internal listener, since it is the only difference. It receives the addFile action from the drive but not the document specific operations.

thegoldenmule: Sorry, didn't include that in my notes. The `sync.bench.ts` is not yet operational.

acaldas (approved):

PR #1046, opened 2025-02-10
feat(scalars): add non-editable if readOnly and validate for time format

alejandrocabriales:
## Ticket

## Description
- Should the field be non-editable if ReadOnly (Disabled) is true.
- Should show the error message when the field loses focus if showErrorOnBlur is true


YasielCabrera (approved):
lgtm

PR #1054, opened 2025-02-11
fix(scalars): handles changes to the "multiple" prop avoiding calling onChange on mount on Select field

nesdc88:
## Ticket
https://trello.com/c/73OWlp1E/827-fix-issue-where-multiple-selected-items-persist-in-the-single-select-field

## Description
- Should the field show only the first element selected when change from multi select to single select.


YasielCabrera (approved):
lgtm

PR #1061, opened 2025-02-12
Server Dependency Refactor

thegoldenmule:
- Introduced a few new objects to pull functionality from `BaseDocumentDriveServer`.
  - `IEventEmitter` + `DefaultEventEmitter`
  - `ISynchronizatioinManager` and `SynchronizationManager`
  - `ITransmitterFactory` and `TransmitterFactory`
- This allows us to fix all cyclic dependencies, as concerns are separated and passed to appropriate pieces.
- Introduced a new builder pattern for dealing with all the different ways we need to create this drive server.
  - Works now, before we have DI, and will be compatible with it later.
  - Works great, probably long-term, for unit + integration tests.


packages/document-drive/src/server/listener/listener-manager.ts
```diff
@@ -375,23 +347,18 @@ export class ListenerManager implements IListenerManager {
     return false;
   }
 
-  getListenerSyncUnits(
-    driveId: string,
-    listenerId: string,
-    loadedDrive?: DocumentDriveDocument,
```
acaldas, line 492: The `loadedDrive` was a hacky performance optimisation to avoid refetching the drive.
The actual fix for this is to remove the sync units state from the Drive document and improve the storage layer. For the time being, is there a strong reason not to keep this parameter?

packages/document-drive/src/server/builder.ts
```diff
@@ -0,0 +1,141 @@
...
+  type IEventEmitter,
+  type IListenerManager,
+  type ISynchronizationManager,
+  type ITransmitterFactory,
+} from "./types";
+
+/**
+ * Builder class for constructing BaseDocumentDriveServer instances with proper configuration
+ */
+export class DocumentDriveServerBuilder {
```
acaldas: It would be good to already name this ReactorBuilder to starting migrating the naming from DocumentDriveServer

acaldas (commented):
In the CI job it seems the reactor initialization also needs to be updated in `reactor-browser`

packages/document-drive/src/server/listener/listener-manager.ts
thegoldenmule: Yah, the reason is because then everything that uses this interface would also require being able to get drives. This is one big reason there were cyclic dependencies. Instead, we should implement caching inside of the implementation.

packages/document-drive/src/server/builder.ts
thegoldenmule: done

nx.json
```diff
@@ -28,10 +28,10 @@
       "dependsOn": ["^build"]
```
acaldas: Let's keep these dependencies for now. After the changes to the bundles are merged they will no longer be needed.

PR #1063, opened 2025-02-13
feat(scalars): add the timeZone and add some test for timeIntervals

alejandrocabriales:
## Ticket

## Description
- Should set the **timezone** as a String. [IANA Time Zone Identifiers](https://www.iana.org/time-zones)
- Should the past dates be disabled if disablePastDates is true.


nesdc88 (approved):
LGTM

PR #1079, opened 2025-02-17
feat(design-system): added new home screen

froid1911:


packages/design-system/src/connect/components/home-screen/home-screen.tsx
```diff
@@ -0,0 +1,31 @@
+import { Icon } from "@/powerhouse";
+import { ComponentPropsWithRef, ForwardedRef, forwardRef } from "react";
+import { twMerge } from "tailwind-merge";
+import { HomeScreenItem } from "../home-screen-item";
+type InputProps = ComponentPropsWithRef<"input">;
+type HomeScreenProps = Omit<InputProps, "className"> & {
+  readonly children: React.ReactNode;
+  readonly containerClassName?: string;
+};
+export const HomeScreen = forwardRef(function HomeScreen(
```
acaldas: I don't think we need to use forwardRef on all components. Only when there is a need for it.

packages/design-system/src/connect/components/home-screen/home-screen.tsx
```diff
@@ -0,0 +1,31 @@
+import { Icon } from "@/powerhouse";
+import { ComponentPropsWithRef, ForwardedRef, forwardRef } from "react";
+import { twMerge } from "tailwind-merge";
+import { HomeScreenItem } from "../home-screen-item";
+type InputProps = ComponentPropsWithRef<"input">;
```
acaldas: Since this is not an input there is no need to add this type.

packages/design-system/src/connect/components/home-screen/home-screen.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <div>
+      <div
+        className={twMerge(
+          "container flex h-screen flex-wrap justify-center gap-4 bg-[url(/home-bg.png)] bg-cover bg-center p-8 text-gray-800 placeholder:text-gray-500",
+          containerClassName,
+        )}
+      >
+        {children}
+        <HomeScreenItem
```
acaldas: Since the caller of this component will have to pass HomeScreenItems for the existing drives anyway, I'd not add this one by default.
The HomeScreen component doesn't need to know anything about drives.

Let's create a HomeScreenCreateDriveItem instead for Connect to use instead.

packages/design-system/src/connect/components/home-screen-item/home-screen-item.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { ComponentPropsWithRef, forwardRef } from "react";
+import { twMerge } from "tailwind-merge";
+
+type InputProps = ComponentPropsWithRef<"input">;
+type HomeScreenItemProps = Omit<InputProps, "className"> & {
+  readonly icon?: React.JSX.Element;
+  readonly title: string;
+  readonly description?: string;
+  readonly containerClassName?: string;
+  readonly sharable?: boolean;
```
acaldas: "shareable" typo

packages/design-system/src/connect/components/home-screen-item/home-screen-item.tsx
```diff
@@ -0,0 +1,46 @@
...
+        )}
+      </div>
+      <div>
+        <h3>{title}</h3>
+        {description && <p className="text-gray-500">{description}</p>}
+      </div>
+      {sharable && (
+        <div className="mb-4 flex w-full justify-start">
+          <div className="absolute left-8 top-8 ">
+            <PeopleFill width={12} height={12} />
```
acaldas, line 42: Screenshot 2025-02-17 at 17 41 08
There seems to be an issue with the svg, it doesn't fill the 12 pixels.

packages/design-system/src/connect/components/home-screen-item/home-screen-item.tsx
acaldas: Shareable is the only prop directly related to a Drive.
I'd keep this one simple and create a separate HomeScreenDriveItem component.
This can receive the drive location (public, local, ...) and set the correct icon.

packages/design-system/src/connect/components/home-screen-item/home-screen-item.tsx
```diff
@@ -0,0 +1,46 @@
+import PeopleFill from "@/assets/icon-components/PeopleFill";
+import { ComponentPropsWithRef, forwardRef } from "react";
+import { twMerge } from "tailwind-merge";
+
+type InputProps = ComponentPropsWithRef<"input">;
+type HomeScreenItemProps = Omit<InputProps, "className"> & {
```
acaldas: This component needs an onClick prop, so we can select the drive on Connect

acaldas (changes requested):

packages/design-system/src/connect/components/home-screen/home-screen.tsx
ryanwolhuter: For both of these there is no need to use the component props of input. this makes things a lot more complicated by including more than 100 extra props. we can see from usage that these other props are not used or forwarded to other components, so they are redundant and give the impression that these props are doing something when they actually are not. 

In general its best to only define the props you actually intend to use. 

this is all you need here, for example

`
type Props = {
  children: React.ReactNode;
  containerClassName?: string;
}
`

packages/design-system/src/connect/components/home-screen/home-screen.tsx
ryanwolhuter: Generally you should only forwardRef when you see an actual error that results from not forwarding a ref.

packages/design-system/src/connect/components/home-screen/home-screen.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+  component: HomeScreen,
+} satisfies Meta<typeof HomeScreen>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+const Template: Story = {
+  args: {
+    children: <div className="mx-auto flex flex-row gap-4"></div>,
```
ryanwolhuter, line 20: using children as a props like this is an antipattern. This actually doesn't make sense, because this children prop is not being passed to your component in the render function (no {...args} on HomeScreen component). You are already passing children to the HomeScreen component (your HomeScreenItem components) so if you want to pass other children you should do it there.

However I would actually recommend against passing the items as props at all. By doing this, you are saying that HomeScreen is meant to work with _any_ random elements. But really you only want it to be used with HomeScreenItem components. Your styles are going to break if you don't pass the correct children.

It would be easier for you and more clear for consumers if you instead have an `items` prop which takes an array of objects which are the props for HomeScreenItem like so:

`type Items = {
title: string;
description: string;
icon: Element;
}
`

and then in HomeScreen you can do

`{items.map(item => <HomeScreenItem key={item.title} {...item} />`

this makes it easy for a consumer to pass a single object as props to HomeScreen since otherwise they would have to define multiple components like here.

PR #1115, opened 2025-02-19
feat(monorepo): use node next

ryanwolhuter:


apps/connect/src/app/document-drive.ts
```diff
@@ -64,7 +63,7 @@ export default (
             id: string,
             operation: Operation,
             forceSync?: boolean,
-        ) => documentDrive.addOperation(drive, id, operation),
+        ) => documentDrive.addOperation(drive, id, operation, { forceSync }),
```
ryanwolhuter, line 65: this was always being ignored

apps/connect/src/store/document-model.ts
```diff
@@ -1,37 +1,37 @@
...
+
+    for (const model of documentModels) {
+        uniqueModels.set(model.documentType, model);
     }
 
     return Array.from(uniqueModels.values());
 }
 
-function getDocumentModelsFromModules(modules: DocumentModelsModule[]) {
+function getDocumentModelsFromModules(modules: DocumentModelLib[]) {
```
ryanwolhuter, line 35: using the type goldenmule defined which is basically the same but has the manifest and generics

acaldas (approved):
Amazing work!

PR #1118, opened 2025-02-20
feat(scalars): add supportedNetworks config and validation

nesdc88:
## Ticket
https://trello.com/c/ET3Uf7pc/789-7-aid-ethereum-dids

## Description
- Should set the supportedNetworks as an Array of objects.
- Should the supported network inserted be set in the supportedNetworks prop.


alejandrocabriales (approved):
LGTM

PR #1119, opened 2025-02-20
feat(scalars): add validations date for different format and allow ti…

alejandrocabriales:
## Ticket

## Description
- Should the time picker allows to select the timezone.
- Should show TimeZone selector if showTimezoneSelect is true.


nesdc88 (approved):
LGTM

PR #1123, opened 2025-02-20
feat(scalars): add system default timezone and new behavior in the ti…

alejandrocabriales:
## Ticket

## Description
- Should the field select the system default timezone if the TimeZone is not specified.
- Should the time picker allow to select the time using the steps in minutes set in the timeIntervals.


YasielCabrera (approved):
lgtm

PR #1128, opened 2025-02-21
fix(scalars): improve generic validation & messages of min/max length

nesdc88:
- Fix min/max length config when its value is 0


YasielCabrera (approved):
lgtm

PR #1135, opened 2025-02-24
fix(scalars): add "path" to the "withValueAndTitle" variant

nesdc88:
## Ticket
https://trello.com/c/30tNL6Eo/848-use-the-path-for-the-withvalueandtitle-variant

## Description
- Should the path be also displayed for the withValueAndTitle variant.


YasielCabrera (approved):
lgtm

PR #1144, opened 2025-02-24
fix(scalars): add support for return direct values (non promises) in callback functions

nesdc88:
## Ticket
https://trello.com/c/8gCklvDv/845-support-both-promises-and-non-promise-values-in-the-callback-functions

## Description
- Should the callback functions support non-promise values


YasielCabrera (approved):
lgtm

PR #1145, opened 2025-02-24
feat(scalars): add componente for navigations

alejandrocabriales:
## Ticket

## Description
- Should the time picker allows to cancel the selection when the Cancel option is clicked.
- Should ensure that the time inserted is closer to the step set when the user types a time and click outside the field


YasielCabrera (approved):
lgtm

PR #1153, opened 2025-02-25
feat: added useSyncStatus

gpuente:
## Description:

- Added `useSyncStatus`


acaldas (commented):
Nice! When fixing the conflicts note that the hooks in `common` were moved to `reactor-browser`

acaldas (approved):

PR #1157, opened 2025-02-26
fix(scalars): restore default key behavior when dropdown is closed on PHID field

nesdc88:
## Ticket
https://trello.com/c/JNgiWUJ0/847-restore-default-key-behaviors

## Description
- Should enable the default key behaviors for the input field.


alejandrocabriales (approved):
LGTM

PR #1159, opened 2025-02-26
fix(connect): connect bundling fixes

acaldas:


ryanwolhuter (approved):

packages/builder-tools/connect-studio/server.ts
```diff
@@ -122,6 +132,11 @@ export async function startServer(options: StartServerOptions = {}) {
           "@powerhousedao",
           "scalars",
         ),
+        "document-model-libs": join(
```
ryanwolhuter: what's up with this? isnt document-model-libs dead?

PR #1180, opened 2025-02-27
fix(scalars): navigating/expanding/collapsing to new nodes while searching

YasielCabrera:
## Ticket
https://trello.com/c/krAwOShg/830-sidebar-high-priorities-tasks

## Description
- Refactor `SidebarProvider` moving its state to a reducer.
- Used references in the `SidebarProvider` to reduce re-renders and memorization/effects dependencies
- Fix the navigation issue in the sidebar when the search was active and an item was expanded/collapsed


nesdc88 (approved):
LGTM

PR #1184, opened 2025-02-27
fix(connect): update type of sync status

ryanwolhuter:


apps/connect/src/hooks/useDocumentDriveServer.ts
```diff
@@ -652,7 +652,7 @@ export function useDocumentDriveServer() {
     );
 
     const onSyncStatus = useCallback(
-        (cb?: (driveId: string, status: SyncStatus, error?: Error) => void) => {
+        (cb: (driveId: string, status: SyncStatus, error?: Error) => void) => {
```
ryanwolhuter, line 655: document drive expects the callback to be defined. if an empty one needs to be used, pass an empty function

apps/connect/src/hooks/useSyncStatus.ts
```diff
@@ -1,8 +1,12 @@
 import { SharingType } from '@powerhousedao/design-system';
+import { SyncStatus } from 'document-drive';
 import { useSyncExternalStore } from 'react';
 import { useDocumentDriveServer } from './useDocumentDriveServer';
 
-export function useSyncStatus(driveId: string, documentId?: string) {
+export function useSyncStatus(
```
ryanwolhuter, line 6: in cases where a function's inferred return type is a type imported from elsewhere, we need to import the type and use an explicit type annotation. this is because otherwise when the code is built, the actual type definition for the inferred return type will not be present, meaning that an external user of the code will not be able to have their return type inferred.

froid1911 (approved):

PR #1188, opened 2025-02-27
feat(scalars): remove the minutes in the display in the case are 00

alejandrocabriales:
## Ticket

## Description
- Short value of offset in the select view
- Should the chosen time be shown with the proper format in the field when the Save option is clicked.


nesdc88 (approved):
LGTM

PR #1219, opened 2025-03-04
fix(switchboard): use tsc instead up tsup and vite node

ryanwolhuter:
- remove usage of tsup and vite node
- add commands for building with tsc for whole workspace
- add dev command for running tsc in watch mode


froid1911:
this is already on main. closing

PR #1226, opened 2025-03-05
fix(scalars): should ensure that EnumField support the onBlur prop

nesdc88:
## Ticket
https://trello.com/c/ZwS2FgEF/851-fix-type-error-in-the-enumfield

## Description
- Should ensure that EnumField support the onBlur prop


alejandrocabriales (approved):
LGTM

PR #1231, opened 2025-03-06
fix(scalars): close popover when Enter is pressed and there is no matching option in IdAutocomplete based fields

nesdc88:
## Ticket

## Description
- Close popover when Enter is pressed and there is no matching option


alejandrocabriales (approved):
LGTM

PR #1244, opened 2025-03-07
fix(scalars): add new properties for amount

alejandrocabriales:
## Ticket

## Description
- Should set **Units** as an Array[].
- Should set **Symbol** as a Boolean. **Before it was currencySymbol.**


nesdc88 (approved):
LGTM

PR #1245, opened 2025-03-07
feat(scalars): add symbol position currency

alejandrocabriales:
## Ticket

## Description
- Should set **symbolPosition** as an Enum. Right by default.


nesdc88 (approved):
LGTM

PR #1251, opened 2025-03-07
fix(scalars): allow withFieldValidation HOC to pass ref & improve components types (first batch)

nesdc88:
## Ticket
https://trello.com/c/ZwS2FgEF/851-fix-type-error-in-the-enumfield-onblur

## Description
- Should ensure that all components support the onBlur prop (in progress)
- Should ensure that all components allow ref (in progress)

AIDField
PHIDField
IdAutocompleteField
CountryCodeField
EnumField
SelectField
RadioGroupField
StringField
TextField
TextareaField
withFieldValidation HOC

### DIFF
--- a/packages/design-system/src/scalars/components/aid-field/aid-field.tsx
+++ b/packages/design-system/src/scalars/components/aid-field/aid-field.tsx
@@ -156,7 +156,7 @@ const AIDFieldRaw = React.forwardRef<HTMLInputElement, AIDFieldProps>(
   },
 );
 
-const AIDFieldWithValidation = withFieldValidation<AIDFieldProps>(AIDFieldRaw, {
+export const AIDField = withFieldValidation<AIDFieldProps>(AIDFieldRaw, {
   validations: {
     _validAIDFormat:
       ({ supportedNetworks }) =>
@@ -212,11 +212,5 @@ const AIDFieldWithValidation = withFieldValidation<AIDFieldProps>(AIDFieldRaw, {
       },
   },
 });
-
-export const AIDField = React.forwardRef<HTMLInputElement, AIDFieldProps>(
-  (props, ref) => {
-    return <AIDFieldWithValidation {...props} innerRef={ref} />;
-  },
-);
 
 AIDField.displayName = "AIDField";
--- a/packages/design-system/src/scalars/components/aid-field/aid-field.tsx
+++ b/packages/design-system/src/scalars/components/aid-field/aid-field.tsx
@@ -156,7 +156,7 @@ const AIDFieldRaw = React.forwardRef<HTMLInputElement, AIDFieldProps>(
   },
 );
 
-export const AIDField = withFieldValidation<AIDFieldProps>(AIDFieldRaw, {
+const AIDFieldWithValidation = withFieldValidation<AIDFieldProps>(AIDFieldRaw, {
   validations: {
     _validAIDFormat:
       ({ supportedNetworks }) =>
@@ -212,5 +212,11 @@ export const AIDField = withFieldValidation<AIDFieldProps>(AIDFieldRaw, {
       },
   },
 });
+
+export const AIDField = React.forwardRef<HTMLInputElement, AIDFieldProps>(
+  (props, ref) => {
+    return <AIDFieldWithValidation {...props} innerRef={ref} />;
+  },
+);
 
 AIDField.displayName = "AIDField";
--- a/packages/design-system/src/scalars/components/fragments/text-field/text-field.tsx
+++ b/packages/design-system/src/scalars/components/fragments/text-field/text-field.tsx
@@ -3,7 +3,7 @@ import { Input } from "../input";
 import { FormLabel } from "../form-label";
 import { FormMessageList } from "../form-message";
 import { FormGroup } from "../form-group";
-import { ErrorHandling, FieldCommonProps, TextProps } from "../../types";
+import type { ErrorHandling, FieldCommonProps, TextProps } from "../../types";
 import { FormDescription } from "../form-description";
 import { CharacterCounter } from "../character-counter";
 import { withFieldValidation } from "../with-field-validation";
@@ -25,7 +25,7 @@ export interface TextFieldProps
   autoComplete?: boolean;
 }
 
-const TextFieldRaw = forwardRef<HTMLInputElement, TextFieldProps>(
+export const TextFieldRaw = forwardRef<HTMLInputElement, TextFieldProps>(
   (
     {
       label,
@@ -107,4 +107,13 @@ const TextFieldRaw = forwardRef<HTMLInputElement, TextFieldProps>(
   },
 );
 
-export const TextField = withFieldValidation<TextFieldProps>(TextFieldRaw);
+const TextFieldWithValidation =
+  withFieldValidation<TextFieldProps>(TextFieldRaw);
+
+export const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
+  (props, ref) => {
+    return <TextFieldWithValidation {...props} innerRef={ref} />;
+  },
+);
+
+TextField.displayName = "TextField";
--- a/packages/design-system/src/scalars/components/fragments/text-field/text-field.tsx
+++ b/packages/design-system/src/scalars/components/fragments/text-field/text-field.tsx
@@ -25,7 +25,7 @@ export interface TextFieldProps
   autoComplete?: boolean;
 }
 
-export const TextFieldRaw = forwardRef<HTMLInputElement, TextFieldProps>(
+const TextFieldRaw = forwardRef<HTMLInputElement, TextFieldProps>(
   (
     {
       label,
--- a/packages/design-system/src/scalars/components/fragments/text-field/text-field.tsx
+++ b/packages/design-system/src/scalars/components/fragments/text-field/text-field.tsx
@@ -107,13 +107,6 @@ export const TextFieldRaw = forwardRef<HTMLInputElement, TextFieldProps>(
   },
 );
 
-const TextFieldWithValidation =
-  withFieldValidation<TextFieldProps>(TextFieldRaw);
-
-export const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
-  (props, ref) => {
-    return <TextFieldWithValidation {...props} innerRef={ref} />;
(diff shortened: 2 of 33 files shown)


packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
```diff
@@ -24,18 +24,24 @@ export interface ValidationOptions<T> {
...
     value,
     name,
     showErrorOnBlur,
     showErrorOnChange,
     validators: customValidators,
+    innerRef,
     ...props
-  }: T) => {
+  }: T & { innerRef?: React.Ref<unknown> }) => {
     const { onChange: onChangeProp, onBlur: onBlurProp } =
```
YasielCabrera: Instead of passing the ref as a prop, the right way of doing this would be to allow to forward the ref in the component that return the HOC. Minimal example:

```tsx
return React.forwardRef(props, ref) => Component
```

This way you don'y need to to create the `WithFieldValidationComponentType` as this is just going to return a `ComponentType` (or an exotic component, not 100% sure).

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
```diff
@@ -182,6 +188,7 @@ export const withFieldValidation = <T extends PossibleProps>(
               onBlur={onBlurCallback}
               onChange={onChangeCallback}
               errors={errors}
+              ref={innerRef}
```
YasielCabrera: note here that the `rest` spread already has a reference from `react-hook-form` and by doing this you are overriding that ref. You should either merge those two refs or find another solution as this could break some of the library behavior and cause unexpected issues

packages/design-system/src/scalars/components/aid-field/aid-field.tsx
```diff
@@ -213,4 +213,10 @@ export const AIDField = withFieldValidation<AIDFieldProps>(AIDFieldRaw, {
   },
 });
 
+export const AIDField = React.forwardRef<HTMLInputElement, AIDFieldProps>(
+  (props, ref) => {
+    return <AIDFieldWithValidation {...props} innerRef={ref} />;
+  },
+);
```
YasielCabrera: It would be best if the HOC returned a component that already accepts references so we avoid doing this in each component which complicate even more the components and after we finish the separation of the components from the Form it is going to be even more unreadable/unmaintainable

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
nesdc88: code updated, please check again

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
nesdc88: code updated, please check again

packages/design-system/src/scalars/components/aid-field/aid-field.tsx
nesdc88: code updated, please check again

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
```diff
@@ -27,257 +27,302 @@ export interface ValidationOptions<T> {
...
-      formState: { errors: formErrors, defaultValues },
-      trigger,
-      setValue,
-      getValues,
-    } = useFormContext();
-    const { submitCount } = useFormState();
+): React.ForwardRefExoticComponent<
+  React.PropsWithoutRef<T> & React.RefAttributes<unknown>
+> => {
+  return React.forwardRef<unknown, T>(
```
YasielCabrera: Instead of having this as unknown, please add a generic type for this to have a strong typed refs. Eg:

```ts
export const withFieldValidation = <T extends PossibleProps, R = unknown>(
  Component: React.ComponentType<T>,
  options?: ValidationOptions<T>,
): React.ForwardRefExoticComponent<
  React.PropsWithoutRef<T> & React.RefAttributes<R>
> => {
  return React.forwardRef<R, T>(...)
```

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
```diff
@@ -27,257 +27,302 @@ export interface ValidationOptions<T> {
...
+        (event: React.ChangeEvent<HTMLInputElement>) => {
+          // update value state
+          if (onChangeProp) {
+            if (Object.hasOwn(event, "target")) {
+              // it is probably an actual event
+              setValue(name, event.target.value);
+            } else {
+              // it is a custom onChange and it pass the value directly
+              setValue(name, event);
+            }
```
YasielCabrera: Why are we moving out the `onChangeCallback` and `onBlurCallback` ?

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
```diff
@@ -27,257 +27,302 @@ export interface ValidationOptions<T> {
...
-      ...(formErrors[name]?.types
-        ? (Object.values(formErrors[name].types ?? []) as string[])
-        : []),
-    ];
-    if (errors.length === 0 && !!formErrors[name]) {
-      // the field is invalid but no error message was provided
-      errors.push("Invalid value");
-    }
+      const onBlurControllerRef = React.useRef<(() => void) | null>(null);
+      const internalValueRef = React.useRef<unknown>(null);
```
YasielCabrera: What is this internalValueRef, and why it is needed? The values is accessible from the form context

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
```diff
@@ -27,257 +27,302 @@ export interface ValidationOptions<T> {
...
+
+      const updateInternalValueRefAndSignal = useCallback(
+        (internalValue: unknown) => {
+          if (internalValueRef.current !== internalValue) {
+            internalValueRef.current = internalValue;
+            setInternalValueSignal((prev) => !prev);
+          }
+        },
+        [],
+      );
```
YasielCabrera: What is this and why it is needed now?

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
```diff
@@ -27,257 +27,302 @@ export interface ValidationOptions<T> {
...
-              if (onChangeProp) {
-                if (Object.hasOwn(event, "target")) {
-                  // it is probably an actual event
-                  setValue(name, event.target.value);
-                } else {
-                  // it is a custom onChange and it pass the value directly
-                  setValue(name, event);
-                }
+            // extract ref from rest
+            const { ref: fieldRef } = rest;
```
YasielCabrera: to make it more readable, rename it to `formFieldRef` as the "field reference" is actually `ref`

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
nesdc88: done

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
nesdc88: rollbacked

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
nesdc88: rollbacked

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
nesdc88: rollbacked

packages/design-system/src/scalars/components/fragments/with-field-validation/with-field-validation.tsx
nesdc88: done

packages/design-system/src/scalars/components/fragments/textarea-field/textarea-field.tsx
```diff
@@ -49,7 +49,10 @@ const textareaBaseStyles = cn(
   "disabled:dark:border-charcoal-800 disabled:dark:bg-charcoal-900 disabled:dark:text-gray-300",
 );
 
-const TextareaFieldRaw = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
+export const TextareaFieldRaw = React.forwardRef<
```
YasielCabrera: Please, do not export "Raw" components yet as they're no ready to be used by developers yet

packages/design-system/src/scalars/components/fragments/text-field/text-field.tsx
```diff
@@ -25,7 +25,7 @@ export interface TextFieldProps
   autoComplete?: boolean;
 }
 
-const TextFieldRaw = forwardRef<HTMLInputElement, TextFieldProps>(
+export const TextFieldRaw = forwardRef<HTMLInputElement, TextFieldProps>(
```
YasielCabrera: Please, do not export "Raw" components yet as they're no ready to be used by developers yet

YasielCabrera (changes requested):

packages/design-system/src/scalars/components/fragments/textarea-field/textarea-field.tsx
nesdc88: Done

packages/design-system/src/scalars/components/fragments/text-field/text-field.tsx
nesdc88: Done

YasielCabrera (approved):
lgtm

PR #1253, opened 2025-03-07
fix(scalars): disable the dates from minData and maxDate

alejandrocabriales:
## Ticket

## Description
- Should change the minDate and maxDate to be disable instead of show message or error


YasielCabrera (approved):
lgtm

PR #1256, opened 2025-03-07
fix(scalars): keep the validations for minDate and maxDate for user i…

alejandrocabriales:
## Ticket

## Description
 Should change the minDate and maxDate to be disable only of show message or error when user write


YasielCabrera (approved):
lgtm

PR #1263, opened 2025-03-11
fix(scalars): change the name props, and fix the cast to send string …

alejandrocabriales:
## Ticket 

## Description
-  DateTime Picker. showTimezoneSelect prop. **Expected Output:** The timezone field should be disabled when the showTimezoneSelect prop = false. **Current Output:** The  field allows select the timezones even when the prop is false. 

- **Expected Output:** The prop should be timeIntervals **Current Output:** The props is named dateIntervals. *


YasielCabrera (approved):
lgtm

PR #1265, opened 2025-03-11
feat(scalars): validate the currency in the array of units

alejandrocabriales:
## Ticket
 
 ## Description
- Code with the symbol next to its status
- Should the unit in the value, be one of the values from units


YasielCabrera (approved):
lgtm

PR #1267, opened 2025-03-12
fix(scalars): allow withFieldValidation HOC to pass ref & improve components types (second batch)

nesdc88:
## Ticket
https://trello.com/c/LMOL17mF/870-fix-type-error-and-ref

## Description
- Should ensure that all components support the onBlur prop
- Should ensure that all components allow ref

AmountField
BooleanField
CheckboxField
ToggleField
CurrencyCodeField
DateTimeField
DatePickerField
TimePickerField
NumberField
UrlField

### DIFF
--- a/packages/design-system/src/scalars/components/fragments/checkbox-field/checkbox-field.tsx
+++ b/packages/design-system/src/scalars/components/fragments/checkbox-field/checkbox-field.tsx
@@ -1,74 +1,86 @@
-import { useId } from "react";
+import React, { useId } from "react";
 import { FormLabel } from "../form-label";
 import { Checkbox, CheckboxValue } from "./checkbox";
 import { cn } from "@/scalars/lib/utils";
 import { FormMessageList } from "../form-message";
-import { FieldCommonProps } from "../../types";
+import type { FieldCommonProps } from "../../types";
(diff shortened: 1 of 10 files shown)


alejandrocabriales (approved):
LGTM

packages/design-system/src/scalars/components/fragments/checkbox-field/checkbox-field.tsx
```diff
@@ -1,74 +1,86 @@
...
-  errors,
-  warnings,
-  onChange,
-  className,
-  ...props
-}) => {
-  const generatedId = useId();
-  const id = idProp ?? generatedId;
-  const hasError = !!errors?.length;
+const CheckboxRaw = React.forwardRef<HTMLButtonElement, CheckboxFieldProps>(
```
YasielCabrera, line 20: Why are we using `HTMLButtonElement`, is not the checkbox actually a checkbox element?

nesdc88: No, CheckboxPrimitive.Root, which is what receives the reference renders a button (also a hidden native checkbox is rendered).

YasielCabrera (approved):

PR #1272, opened 2025-03-12
Merge dspot branch

YasielCabrera:


packages/design-system/src/scalars/components/sidebar/sidebar.stories.tsx
```diff
@@ -1,9 +1,9 @@
 import { Icon } from "#powerhouse";
 import type { Meta, StoryObj } from "@storybook/react";
-import { useState } from "react";
-import mockedTree from "./mocked_tree.json";
+import { useCallback, useState } from "react";
+import mockedTree from "./mocked_tree.json" with { type: "json" };
```
acaldas: `with { type: "json" };` is not supported on stories.
This should be just `import mockedTree from "./mocked_tree.json";`

YasielCabrera: Fixed

acaldas (approved):

PR #1273, opened 2025-03-13
fix(scalars): avoid transformation of valid value

alejandrocabriales:
## Ticket

## Description
- Avoid to transform value when already had correct value


nesdc88 (approved):
LGTM

PR #1277, opened 2025-03-13
fix(scalars): add hiddend div for styles in connect

alejandrocabriales:
## Ticket

## Description
TimePicker **Expected Output:** The component should be shown with the proper styles. **Current Output:** The component looks broken.


YasielCabrera (changes requested):
that workaround should be done in connect instead of the library or it will not work

alejandrocabriales:
you are right, I will close this

PR #1278, opened 2025-03-13
fix(scalars): remove some props from stories and show symbol props as…

alejandrocabriales:
## Ticket
https://trello.com/c/RQlyOdrZ/778-12-currencycodefield-currencycodefiat-currencycodecrypto

## Description
- SymbolPosition prop should be visible only if includeCurrencySymbols is true. 
- Remove contentAlign prop 
- Remove contentClassName prop


YasielCabrera (approved):
lgtm

PR #1284, opened 2025-03-14
feat: added new amount scalars

gpuente:
## Description:

- added new amount scalars

### DIFF
--- a/packages/scalars/src/scalars/Amount.ts
+++ b/packages/scalars/src/scalars/Amount.ts
@@ -1,0 +1,77 @@
+import {
+  GraphQLError,
+  GraphQLScalarType,
+  type GraphQLScalarTypeConfig,
+  Kind,
+} from "graphql";
+import { z } from "zod";
+
+export type Amount = {
+  unit: string;
+  value: number;
+};
+
+export type ScalarType = {
+  input: Amount;
+  output: Amount;
+};
+
+export const type = "{ unit: string, value: number }";
+
+export const typedef = "scalar Amount";
+
+export const schema = z.object({
+  unit: z.string(),
+  value: z.number().finite(),
+});
+
+export const stringSchema =
+  "z.object({ unit: z.string(), value: z.number().finite() })";
+
+const amountValidation = (value: unknown): Amount => {
+  if (typeof value !== "object" || !value) {
+    throw new GraphQLError(`Invalid Amount value: ${JSON.stringify(value)}`);
+  }
+
+  const result = schema.safeParse(value);
+
+  if (result.success) return result.data;
+  throw new GraphQLError(result.error.message);
+};
+
+export const config: GraphQLScalarTypeConfig<Amount, Amount> = {
+  name: "Amount",
+  description:
+    "A custom scalar that represents a currency amount with its currency type",
+  serialize: amountValidation,
+  parseValue: amountValidation,
+  parseLiteral: (ast) => {
+    if (ast.kind !== Kind.OBJECT) {
+      throw new GraphQLError("Value must be an object", { nodes: ast });
+    }
+
+    const unitField = ast.fields.find((f) => f.name.value === "unit");
+    const valueField = ast.fields.find((f) => f.name.value === "value");
+
+    if (!unitField || unitField.value.kind !== Kind.STRING) {
+      throw new GraphQLError("unit must be a valid string value", {
+        nodes: ast,
+      });
+    }
+
+    if (!valueField || valueField.value.kind !== Kind.FLOAT) {
+      throw new GraphQLError("value must be a valid float value", {
+        nodes: ast,
+      });
+    }
+
+    const value = {
+      unit: unitField.value.value,
+      value: parseFloat(valueField.value.value),
+    };
+
+    return amountValidation(value);
+  },
+};
+
+export const scalar = new GraphQLScalarType(config);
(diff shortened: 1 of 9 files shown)


packages/scalars/src/scalars/AmountCrypto.ts
```diff
@@ -0,0 +1,89 @@
+import {
+  GraphQLError,
+  GraphQLScalarType,
+  type GraphQLScalarTypeConfig,
+  Kind,
+} from "graphql";
+import { z } from "zod";
+
+export type SupportedCrypto = "DAI" | "ETH" | "MKR" | "SKY" | "USDC" | "USDS";
```
YasielCabrera: This list is the default supported crypto, but it should support "any" crypto, so probably the units here should accept any string and the validation if it is supported or not should be delegated to the developer

packages/scalars/src/scalars/AmountCurrency.ts
```diff
@@ -0,0 +1,94 @@
...
+  value: number;
+};
+
+export type ScalarType = {
+  input: AmountCurrency;
+  output: AmountCurrency;
+};
+
+export const type =
+  "{ unit: 'DAI' | 'ETH' | 'MKR' | 'SKY' | 'USDC' | 'USDS' | 'USD' | 'EUR' | 'GBP', value: number }";
```
YasielCabrera: Similar to my previous comment, this is just the default Crypto/Fiat list, but it should support any

packages/scalars/src/scalars/AmountFiat.ts
```diff
@@ -0,0 +1,85 @@
+import {
+  GraphQLError,
+  GraphQLScalarType,
+  type GraphQLScalarTypeConfig,
+  Kind,
+} from "graphql";
+import { z } from "zod";
+
+export type SupportedFiat = "USD" | "EUR" | "GBP";
```
YasielCabrera: Similar to my previous comment, this is just the default Fiat list, but it should support any

packages/scalars/src/scalars/index.ts
```diff
@@ -27,6 +35,10 @@ export type { ScalarType as PHIDScalarType } from "./PHID.js";
 export type { ScalarType as URLScalarType } from "./URL.js";
 
 export {
+  Amount,
+  AmountCrypto,
+  AmountCurrency,
+  AmountFiat,
   AmountMoney,
```
YasielCabrera, line 42: is this `AmountMoney` needed now that we have the other amount types?

YasielCabrera (changes requested):

packages/scalars/src/scalars/Amount.ts
```diff
@@ -0,0 +1,77 @@
+import {
+  GraphQLError,
+  GraphQLScalarType,
+  type GraphQLScalarTypeConfig,
+  Kind,
+} from "graphql";
+import { z } from "zod";
+
+export type Amount = {
+  unit: string;
```
alejandrocabriales: In the case of amount, unit its optional

packages/scalars/src/scalars/index.ts
```diff
@@ -1,4 +1,8 @@
 // namespace imports -- DO NOT REMOVE OR EDIT THIS COMMENT
+import * as Amount from "./Amount.js";
```
alejandrocabriales, line 2: Also I think that we need AmountPercentage Scalar

alejandrocabriales (changes requested):

packages/scalars/src/scalars/AmountCrypto.ts
gpuente: added fix here: [9bf6769](https://github.com/powerhouse-inc/powerhouse/pull/1284/commits/9bf6769a1ba6225080fbfdcf37e7496847ed16fa)

packages/scalars/src/scalars/AmountCurrency.ts
gpuente: added fix here: [9bf6769](https://github.com/powerhouse-inc/powerhouse/pull/1284/commits/9bf6769a1ba6225080fbfdcf37e7496847ed16fa)

packages/scalars/src/scalars/Amount.ts
gpuente: added fix here: [9bf6769](https://github.com/powerhouse-inc/powerhouse/pull/1284/commits/9bf6769a1ba6225080fbfdcf37e7496847ed16fa)

packages/scalars/src/scalars/index.ts
gpuente: it's already there: https://github.com/powerhouse-inc/powerhouse/blob/dspot-scalars/packages/scalars/src/scalars/AmountPercentage.ts

packages/scalars/src/scalars/AmountFiat.ts
gpuente: added fix here: [9bf6769](https://github.com/powerhouse-inc/powerhouse/pull/1284/commits/9bf6769a1ba6225080fbfdcf37e7496847ed16fa)

packages/scalars/src/scalars/index.ts
gpuente: that's a good question... I believe `AmountMoney` and `AmountCurrency` are interchangeable, but `AmountCurrency` also allows you to specify the `unit`, whereas `AmountMoney only represents a single float value. The difference is more about semantic usage.

YasielCabrera (approved):

alejandrocabriales (approved):
LGTM

PR #1288, opened 2025-03-17
fix(scalars): update content items font weight to normal in Select field

nesdc88:
## Ticket
https://trello.com/c/EPExsKwE/837-change-the-weight-font-in-the-dropdown-items

## Description
- Should change the current font weight to 400


alejandrocabriales:
LGTM

PR #1291, opened 2025-03-17
fix(scalars): update items font weight in Select based fields

nesdc88:
## Ticket
https://trello.com/c/EPExsKwE/837-change-the-weight-font-in-the-dropdown-items

## Description
- Should change the current font weight to 400


YasielCabrera (approved):
lgtm

PR #1309, opened 2025-03-18
feat(scalars): add OID field base, stories and tests

nesdc88:
## Ticket

## Description
- Add Default, Empty, Filled and Open stories
- Should set "autoComplete" config as boolean (true by default)
- Should create the new OIDField wrapper component
- Should ensure the field is not empty if "required" is true
- Should "value" config reflect the current value of the field (the OID entered by the user), or null if not set
- Should "defaultValue" config reflect the default OID value if one is provided, or null if not set
- Should validate the OID format following the pattern 8-4-4-4-12 hexadecimal digits (uuid v4)
- Should implement the callback functions for OID
- Should show the error/warning messages according to the validations

### DIFF
--- a/packages/design-system/src/scalars/components/oid-field/index.ts
+++ b/packages/design-system/src/scalars/components/oid-field/index.ts
@@ -1,0 +1,1 @@
+export * from "./oid-field.js";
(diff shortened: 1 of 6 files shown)


alejandrocabriales (approved):
LGTM

PR #1315, opened 2025-03-18
fix(scalars): fix the docs in date-time component

alejandrocabriales:
## Description
- DateTime Picker. showDateSelect = true, showTimeSelect = false. The time should not be displayed.The components is showing all the time information. The controls are not working for these two props. 
- DateTime Picker. Readme file, timeZone prop. **Expected Output:** A description should shown. E.g. description provided in the user story. **Current Output:** The prop doesn't have any description.


YasielCabrera (approved):
lgtm

PR #1317, opened 2025-03-18
Merge dspot into main

YasielCabrera:
## Ticket
https://trello.com/c/n5FUv8a8


packages/design-system/src/scalars/lib/index.ts
```diff
@@ -1,4 +1,5 @@
-export * from "./currency-list.js";
+/* eslint-disable prettier/prettier */
+export * from "./decorators.js";
```
acaldas: This barrel file should only export files meant to be used on web apps.
The decorators file has storybook code so it should not be part of the final bundle.

YasielCabrera: Removed from the exports!

acaldas (approved):

PR #1319, opened 2025-03-18
fix(scalars): fix the scroll in main browsers

alejandrocabriales:
## Ticket
https://trello.com/c/t0aw2hKJ/903-update-styles-on-datetimefield-component-after-migration-to-tailwind-v4

## Description
- Should apply the styles defined on the design for the date picker.
- Should apply the styles defined on the design for the time picker.


YasielCabrera (approved):
lgtm

PR #1326, opened 2025-03-19
fix(scalars): add validations for input pass and future days

alejandrocabriales:
## Description
-DateTime Field, Date Picker. disablePastDates = true.  The past dates should be disabled in the date picker and the input should not allow to insert past dates.  The input allows to insert and submit past dates. **with disableFutureDates shows the same issue**


nesdc88 (approved):
LGTM

PR #1331, opened 2025-03-20
fix(scalars): add validations in the currencie field and expose same …

alejandrocabriales:
## Ticket

## Description
- Should implement a function to get the default list of currencies.


nesdc88 (approved):
LGTM

PR #1332, opened 2025-03-20
feat(scalars): add new feats to IdAutocomplete based fields and improve default placeholder text

nesdc88:
## Ticket

## Description
- Should the uppercase and lowercase letters be accepted in the uuid (OID,PHID)
- Should the leading and trailing white spaces be trimmed (OID, AID, PHID)
- Should change "path" to "Type", "uuid" to "oid" & "phd" to "phid" in the placeholder (OID, PHID)


packages/design-system/src/scalars/components/oid-field/oid-field.tsx
```diff
@@ -53,6 +53,8 @@ const OIDFieldRaw = React.forwardRef<HTMLInputElement, OIDFieldProps>(
     const id = idProp ?? `${prefix}-oid`;
     const autoComplete = autoCompleteProp ?? true;
 
+    const contextValue = useMemo(() => ({}), []);
```
alejandrocabriales: Does useMemo(() => ({}), []) provide any real optimization here, considering that the object {} is always the same? Wouldn't it be simpler to pass it directly in the Provider

nesdc88: Actually, in this case it is not necessary to pass anything as context, I have removed the related code

alejandrocabriales:
LGTM

PR #1338, opened 2025-03-21
fix(scalars): fix the message and fixed minDate and maxDate with disa…

alejandrocabriales:
## Ticket

## Description
- DateTime Field. maxDate = 03/21/2025. Selecting 03/21/2025 thought the date picker. **Expected Output:** The date should be allowed. **Current Output:** An error message is displayed referring to the date should be on or before of the maxDate set (03/21/2025) but the validation is not allowing to submit it.  

- DateTime Field. disableFutureDates = true, minDate = 03/06/2025.  The date before 03/06/2025 should be disabled, the dates after the current day should be disabled.  **Current Output:** Combining these two props, the date before the minDate value are enabled. **Using maxDate and disablePastDtaes happens the same issue

### DIFF
--- a/packages/design-system/src/scalars/components/time-field/utils.ts
+++ b/packages/design-system/src/scalars/components/time-field/utils.ts
@@ -351,27 +351,27 @@ export const convert12hTo24h = (input: string) => {
 };
 
 /**
- * Formatea una cadena de entrada de tiempo en un formato de visualización estandarizado
- * @param input - La cadena de tiempo de entrada. Puede estar en varios formatos:
- *   - Formato 24 horas (ej., "14:30", "08:00")
- *   - Formato 12 horas (ej., "2:30 PM", "8:00 AM")
- *   - Formato corto (ej., "1430", "0800")
- * @param is12HourFormat - Si es true, devuelve en formato 12 horas con AM/PM
- *                        Si es false, devuelve en formato 24 horas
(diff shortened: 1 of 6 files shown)


packages/design-system/src/scalars/components/time-field/utils.ts
```diff
@@ -347,23 +351,33 @@ export const convert12hTo24h = (input: string) => {
...
- *   - 12-hour format (e.g., "2:30 PM", "8:00 AM")
- *   - Short format (e.g., "1430", "0800")
- * @param is12HourFormat - If true, outputs in 12-hour format with AM/PM
- *                        If false, outputs in 24-hour format
- * @param timeIntervals - Optional interval in minutes to round the time to
- *                       (e.g., 15 would round to nearest quarter hour)
- * @returns Formatted time string
- *   - 12-hour format: "hh:mm AM/PM" (e.g., "02:30 PM")
- *   - 24-hour format: "HH:mm" (e.g., "14:30")
+ * Formatea una cadena de entrada de tiempo en un formato de visualización estandarizado
```
nesdc88: These comments must be in English

alejandrocabriales: you are right, done

nesdc88 (approved):
LGTM

PR #1341, opened 2025-03-21
feat(scalars): move time-component-to-new-struct

alejandrocabriales:
##Ticket
https://trello.com/c/9yIru7Mw/911-adapt-datetime-component-to-be-used-outside-of-the-form

## Description
- Should separate the types in two categories to be used outside and inside of the form. (TimePickerField)


YasielCabrera (approved):
lgtm

PR #1342, opened 2025-03-21
fix(scalars): fix-with-no-select-value-amount-fix-values-stories

alejandrocabriales:
## Ticket

## Description
- The only unit for this variant should be %. A list of units is added to the controls.
- The placeholder should be represented by default.  The entire field is represented with value.
 - The props should be displayed according of the variant/type. There are props with none effect over the field because doesn't have relation with the percent.
- The dropdown should have a proper width. The items are displayed with a big blank space at the right side. 

ScreenShots
image


nesdc88 (approved):
LGTM

PR #1344, opened 2025-03-21
refactor(scalars): add OIDInput & refactor IdAutocomplete base component

nesdc88:
## Ticket

## Description
- Should separate the types in two categories to be used outside and inside of the form
- Should create the component variant to be used outside of the form


alejandrocabriales (approved):
LGTM

PR #1353, opened 2025-03-24
fix(scalars): add correct value for default behavior

alejandrocabriales:
## Ticket 
https://trello.com/c/RQlyOdrZ/778-12-currencycodefield-currencycodefiat-currencycodecrypto

## Description
- Currencies control. It should allow to insert a list of currencies (units) .Clicking on the field (prop) an error appears.


YasielCabrera (approved):
lgtm

PR #1363, opened 2025-03-25
feat(scalars): prepare amount component to be use outside of form

alejandrocabriales:
## Ticket
https://trello.com/c/kWWRiefU/912-adapt-amount-component-to-be-used-outside-of-the-form

## Description
- Should separate the types in two categories to be used outside and inside of the form.
- Should create the component variant to be used outside of the form.
- Should apply the the new file structure in code (design system)


nesdc88 (approved):
LGTM

PR #1373, opened 2025-03-27
Merge/dspot into main

alejandrocabriales:
## Ticket
 https://trello.com/c/n5FUv8a8


acaldas (changes requested):
The monorepo is using pnpm, is there a reason for using npm?

Please remove the package-lock.json files.

acaldas (commented):
I updated the branch but the pnpm-lock needs to be updated before merging

PR #1391, opened 2025-04-01
fix(ui): show agent type at the right taking into account the field width

nesdc88:
## Ticket
https://trello.com/c/zIPQ2aH6/910-adapt-aid-component-to-be-used-outside-of-the-form

## Description
- The agent type should be displayed at the right taking into account the field width


alejandrocabriales (approved):
LGTM

PR #1439, opened 2025-04-14
fix(ui): default path & default button cursor pointer styles

nesdc88:
## Ticket
https://trello.com/c/HAFtVnyw/947-split-view-edit-mode-for-foundational-and-grounding-documents

## Description
- PHID: when the type is not specified, the placeholder message displayed is "Path not available" instead of "Type not available".
- The cursor pointer should be a hand when is hovering over the copy and reload icons.


alejandrocabriales (approved):
LGTM

YasielCabrera (approved):
lgtm

PR #1462, opened 2025-04-29
feat: add diff support to PHIDInput UI component

nesdc88:
## Ticket
https://trello.com/c/uOoYPkcV/987-create-a-read-only-diff-status-for-the-phid-component

## Description
- Should establish common elements to be used as base for all the components.
- Should add to the existing component the new status Read Only Diff.

### DIFF
--- a/packages/design-system/src/ui/components/data-entry/input/text-diff.tsx
+++ b/packages/design-system/src/ui/components/data-entry/input/text-diff.tsx
@@ -1,0 +1,59 @@
+import { cn } from "@powerhousedao/design-system/scalars";
+import { diffSentences, diffWords } from "diff";
+import { useMemo } from "react";
+import type {
+  DiffMode,
+  ViewMode,
+} from "../../../../scalars/components/types.js";
+
+interface TextDiffProps {
+  baseValue: string;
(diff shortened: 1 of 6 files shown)


packages/design-system/src/ui/components/data-entry/input/text-diff.tsx
```diff
@@ -0,0 +1,59 @@
+import { cn } from "@powerhousedao/design-system/scalars";
+import { diffSentences, diffWords } from "diff";
+import { useMemo } from "react";
+import type {
+  DiffMode,
+  ViewMode,
+} from "../../../../scalars/components/types.js";
+
+interface TextDiffProps {
```
YasielCabrera: you should extend from WithDifferences to reuse the diff props

YasielCabrera (changes requested):

packages/design-system/src/ui/components/data-entry/input/text-diff.tsx
nesdc88: Done

YasielCabrera (approved):
lgtm

PR #1463, opened 2025-04-30
feat(diff): add diff input single line string

alejandrocabriales:
## Ticket

## Description
 Should add to the existing component the new status Read Only Diff.


nesdc88 (approved):
LGTM

PR #1469, opened 2025-05-05
feat(atlas): add splitted text area for diff area

alejandrocabriales:
## Ticket

## Description
- Should a new prop be visible when the disabled prop is true.
- Should the new variant allow to see the diff given the additions/removals and changes.


YasielCabrera (approved):
lgtm

PR #1471, opened 2025-05-06
fix(scalars): add control for string field

alejandrocabriales:
## Ticket

## Description
- Should the new variant allow to see the diff given the additions/removals and changes.


nesdc88 (approved):
LGTM

PR #1479, opened 2025-05-07
feat(scalars): Add diff props to string field component

alejandrocabriales:
## Ticket
https://trello.com/c/X10a3hCg/989-add-the-read-only-diff-status-for-the-strings-to-the-document-types
## Description
- Fix add props in scalars and pass diffProps to String Field


nesdc88 (approved):
LGTM

PR #1527, opened 2025-05-14
feat(scalars & ui): add read only differences to the Enum component

nesdc88:
## Ticket
https://trello.com/c/yMkx5aiU/1001-create-a-read-only-diff-status-for-the-enum-component

## Description
- Should add the new props to show the new status
- Should the new status allow to see the diff given the additions/removals and changes
- Should create the read only diff status for the Select & EnumField (storybook variants)


alejandrocabriales (approved):
LGTM

PR #1529, opened 2025-05-14
feat(switchboard): added authenticated user and role check to graphql…

froid1911:
… context


packages/reactor-api/src/server.ts
```diff
@@ -246,13 +261,13 @@ export async function startAPI(
     app.use(async (req, res, next): Promise<void> => {
       const token = req.headers.authorization?.split(" ")[1];
       if (!token) {
-        res.status(401).json({ error: "Unauthorized" });
+        res.status(400).json({ error: "Missing authorization token" });
         return;
       }
 
       const verified = await verifyAuthBearerToken(token);
```
acaldas, line 272: This needs to be wrapped into a try/catch so we can throw a 400 error if an exception occurs in the token verification (and log the error for ourselves).

PR #1538, opened 2025-05-16
Decouple Sync Units from document drive state

acaldas:


packages/document-drive/src/server/base-server.ts
```diff
@@ -795,9 +713,54 @@ export class BaseDocumentDriveServer
     return this.documentStorage.getChildren(driveId);
   }
 
+  protected async addChild(
+    parentId: string,
+    documentId: string,
+  ): Promise<void> {
+    // TODO: check if document exists? Should that be a concern here?
```
thegoldenmule, line 860: This should be handled by storage, as it needs to test for a variety of other issues (like circular refs).

packages/document-drive/src/queue/base.ts
```diff
@@ -53,38 +59,29 @@ export class MemoryQueue<T> implements IQueue<T> {
     return this.id;
   }
 
-  async setBlocked(blocked: boolean) {
-    this.blocked = blocked;
-  }
-
   async isBlocked() {
-    return this.blocked;
+    return this.running || this.deleted || this.dependencies.length > 0;
```
thegoldenmule, line 47: The switch to `running` instead of a `blocked` seems less fault tolerant (eg - any code path that doesn't set it back to false will block indefinitely), and particularly because anything can toggle it externally. Why does that need to be public?

packages/document-drive/src/storage/types.ts
```diff
@@ -63,6 +74,33 @@ export interface IDocumentStorage {
...
+      /** The ids of the parent documents. If '*' then select all. */
+      parentId?: string[];
+      /** The ids of the documents. If '*' then select all. */
+      documentId?: string[];
+      /** The types of the document models. If '*' then select all. */
+      documentModelType?: string[];
+      /** The scopes of the documents. If '*' then select all. */
+      scope?: string[];
+      /** The branches of the documents. If '*' then select all. */
+      branch?: string[];
```
thegoldenmule: We'll need to line these parameters up with indexes.

packages/document-drive/src/queue/base.ts
thegoldenmule: Maybe an enum would be clearer? `Idle`, `Running`, `Deleted`, etc?

PR #1545, opened 2025-05-18
merge: dspot-scalars into main 18-05-2025

nesdc88:
## Ticket
- Merge dspot-scalars into main


gpuente:
LGTM!
Just a quick question — will the scalar components be removed from the design system in a follow-up PR?

gpuente (approved):

nesdc88:
I think they should be removed but not immediately (until the remaining repos that may be using components from the scalars and ui folders from design-system are migrated to use those from document-engineering repo)

YasielCabrera:
> I think they should be removed but not immediately (until the remaining repos that may be using components from the scalars and ui folders from design-system are migrated to use those from document-engineering repo)

Actually, I think we can remove them, they are already in the older version of the `design-system` so it doesn't make sense to keep them here

PR #1573, opened 2025-05-29
wip: Reactor 2: react harder

acaldas:


packages/reactor/docs/planning/Operations/index.md
```diff
@@ -0,0 +1,223 @@
...
+   * Base64 encoded signature of the payload.
+   */
+  signature: string;
+};
+
+type ActionContext = {
+  /**
+   * The signature of the action.
+   */
+  signature: Signature;
```
acaldas: Here it should be an array of signatures for when an operation is shuffled. The original signature is kept, and a new one is added by the reactor that did the shuffle.

packages/reactor/docs/planning/Operations/index.md
```diff
@@ -0,0 +1,223 @@
...
+};
+```
+
+### Operations
+
+An `Operation` represents the result of an `Action` being executed. It is a "fact", not an "intent".
+
+Unlike a typical ES approach, where Actions are generally discarded or stored in an outbox pattern, each `Action` is stored with the `Operation` that is produced from it. This allows every client to re-execute each `Action` and verify the resulting `Operation`.
+
+The lack of `previousState` is simply an optimization. As the previous state is always known, it is not necessary to store it (but perhaps could be necessary for debugging).
```
acaldas: I'd also remove `resultingState` from the `Operation` type as it can be very expensive.

packages/reactor/docs/planning/Operations/index.md
```diff
@@ -0,0 +1,223 @@
...
+
+The `hash` field stores the expected document state after applying the operation; reducers compute the state hash during execution and compare it to this value to detect divergence. Along with the unique `(documentId, scope, branch, index)` constraint in the storage schema, these identifiers ensure that replaying operations cannot introduce inconsistent state.
+
+### Attachments
+
+See the [Attachments doc](../Attachments/index.md) for more information on attachments. These are not stored with `Operation`s directly, but with the `Action` that consumes them.
+
+### Action Error Handling
+
+Operations cannot be in a failure state, as they represent facts about the document state. Instead, the `Action` that produced the `Operation` may fail.
```
acaldas, line 390: Currently, we do have a `error` field on operations. It might happen when shuffling an operation that becomes invalid when applied to a new state.
I think the error gets bubbled up to the sync protocol as mentioned, which interrupts syncing between two Reactors.
Maybe we should instead add the shuffled operations with the error state, and the other Reactor should then deal with that error. (IE, notify the user that is action failed)

packages/reactor/docs/planning/Operations/index.md
thegoldenmule: This signature scheme is intended to be composable. A list of signatures does not guarantee order (I could add or remove signatures with no proof that the other signatures are aware), but signing other signatures does.

```ts
type SignedPayloadParameters = {
  /**
   * Additional signatures to append to the payload.
   */
  signatures: Signature[];
};
```

packages/reactor/docs/planning/Operations/index.md
thegoldenmule: I think we should consider only if/when we see performance issues. At least for system development, it's going to be a nightmare to debug without it.

packages/reactor/docs/planning/Operations/index.md
thegoldenmule: Yah, I saw that, but that's because our current Operations are doing double duty as a command/event hybrid. They are not immutable and hold tons of pre-application, in-progress, and post-application state.

The reason I put this here is to draw a hard line about what our event log actually defines. I get that in a typical ES architecture we'd throw all our errors on the event log too, but that's because we wouldn't be syncing it.

With this new proposal, we don't need to create error Operations because we can bubble up errors from Action application (i.e. from the `IJobExecutor`).

So if we hit an error because of shuffling (or any other reason), we bubble that up to the sync manager through errors just like we do now, but we do not persist the errors in the operation store. The sync manager is free to do whatever it wants with that error, including sending it to another reactor to try to resolve-- but we keep it out of other core systems that do not need to be concerned with it.

packages/reactor/docs/planning/Operations/index.md
thegoldenmule: I added some information to address your comments:

```
- In general, when submitting a new action, the signature payload will include:
  - The action itself
  - The application information
  - The hash of the latest operation for the `(documentId, scope, branch)`
- In the case or reshuffling, the `ISigner` will wrap the signature of an action in a new signature. This is because the previously signed state hash may not be valid so a new one must be generated.
```

packages/reactor/docs/planning/Operations/index.md
acaldas: I was thinking from the perspective of having this as a type used by external devs.
For our internal use we could have a separate type, that adds the `resultingState` to the `Operation`. This could already be a parsed JS object instead of a string.
Just a nitpick.

packages/reactor/docs/planning/Queue/index.md
```diff
@@ -0,0 +1,29 @@
...
+
+The `IQueue` provides a simple API to queue new write jobs. Internally, it creates separate queues keyed by `(documentId, scope, branch)` to ensure proper ordering of jobs within each document context. Jobs are processed in FIFO order within each queue to maintain consistency. When jobs are enqueued, the queue emits 'jobAvailable' events to the event bus to notify job executors.
+
+When there is heavy contention (say many jobs enqueued from many clients on the same `(documentId, scope, branch)` combination), reshuffle logic is applied by the job executor to proactively prevent as many retries and requeues as possible.
+
+### Signing
+
+All jobs submitted to the queue should already have valid signatures on the `Action` objects. The `IQueue` itself does not verify signatures on queued jobs, nor does it do any signing.
+
+Signature validation is left to the `IJobExecutor`, as it may be considered a high-CPU operation.
```
acaldas: I wonder if the signature should just be validated at the API layer. If you call `addActions` with invalid signatures it fails right away.
Only verified actions (User "x" did this action) get added to the queue.`IJobExecutor` would instead enforce Authorization (User "x" is authorized to do this action).

packages/reactor/docs/planning/Jobs/index.md
```diff
@@ -0,0 +1,116 @@
...
+    new Date(b.timestamp || "").getTime()
+  )
+  .map((op, i) => ({
+    ...op,
+    index: startIndex.index + i,
+    skip: i === 0 ? startIndex.skip : 0,
+  }));
+```
+
+### Signature Verification
```
acaldas: Following on https://github.com/powerhouse-inc/powerhouse/pull/1573/files#r2126807255

This could be `Authorization Verification` instead

packages/reactor/docs/planning/Jobs/index.md
```diff
@@ -0,0 +1,116 @@
...
+
+In all cases, the `IJobExecutor` will verify that the `Action` signatures are valid.
+
+If the `Action` signature is not valid, the `IJobExecutor` will emit a `jobFailed` event and the job will not be retried.
+
+### Operation Verification 
+
+Optionally, a `Job` may contain information about expected `Operation` data. In this case, the executor will verify that resulting `Operation` hash matches the expected `Operation` hash.
```
acaldas: Should this be a concern of the `IJobExecutor`?
The flow could be:
- `ISyncManager` receives operations from another Reactor
- Adds them to the queue
- Waits for job result with the applied operations
- Checks for inconsistent hashes, operation errors, etc
- Proceeds with sync protocol flow

packages/reactor/docs/planning/Operations/index.md
thegoldenmule: That's fair. I'm wondering if we should always store the resultingState but only populate it in certain circumstances.

packages/reactor/docs/planning/Jobs/index.md
```diff
@@ -0,0 +1,116 @@
...
+- [Interface](interface.md)
+- [Usage](usage.md)
+
+### Notes
+
+- The job executor listens for 'jobAvailable' events from the event bus instead of polling the queue.
+- When a 'jobAvailable' event is received, the executor checks if it has capacity to process more jobs.
+- If capacity allows, the executor pulls a job from the queue and executes it.
+- When a job is fully executed, it tries to pull another job from the queue.
+- Jobs should be executed in the correct order per document/scope/branch to maintain consistency.
```
acaldas: I think we need to introduce cross-document/scope/branch dependencies.
For example, we want the first operation to be run for a document to be the `ACTION_CREATE`, so operations on other scopes are blocked by it. The same for version upgrades.

Will linked `authorization` scopes also create dependencies?

packages/reactor/docs/planning/Operations/upgrades.md
```diff
@@ -0,0 +1,76 @@
...
+```tsx
+
+const jobStatus = await reactor.create(doc);
+
+```
+
+3. Internally, the `IReactor` will:
+
+- Create an `Action` of type `ACTION_CREATE`.
+- Create an `Action` of type `ACTION_UPGRADE`.
```
acaldas: If we want to follow the imperative naming we have been using for operations, this could be `CREATE_DOCUMENT` and `UPGRADE_DOCUMENT` instead.

packages/reactor/docs/planning/Operations/upgrades.md
```diff
@@ -0,0 +1,76 @@
...
+- Submit both actions in a single `Job` to the `IQueue`.
+
+4. The `IJobExecutor` will pull the `Job` from the `IQueue` and execute it.
+
+5. The `Job` will be executed by the `IReactor`, which will:
+
+- Create an `Operation` for the `ACTION_CREATE`. This will result in the creation of a new document with a filled out `header` scope but an empty, `{}`, default scope (`global`).
```
acaldas: Should it be the `global` scope or a new `document` scope?
This `document` scope would also have the `SET_DOCUMENT_NAME` and `SET_META` operations.
For the upgrades we could either support a single upgrade operation at the document-level or one per scope.

packages/reactor/docs/planning/Queue/index.md
thegoldenmule: That's an interesting idea. Like, in `IReactor::mutate()`?

The reason I was thinking here is to make sure that load is in the right place, as `IJobExecutor` will be the thing we want to horizontally scale / make location-transparent (i.e. run it in a separate process, or a separate box altogether).

The other reason would be to only do it in one place, as we will have the API to add actions to the queue locally, but we'll also have sync adding them. The current thinking is that sync manager will need to add these directly to the queue rather than through the reactor API. This is because the job executor is what validates that the output operation hashes meet the requirements of the remote operation before adding them to its own operations store. I'd rather not clutter the reactor API by adding that stuff there.

packages/reactor/docs/planning/Jobs/index.md
thegoldenmule: The reason I had it here is because the job executor is what adds the resulting operations to the store, and of course we want that verification to happen before we do that.

packages/reactor/docs/planning/Operations/upgrades.md
```diff
@@ -0,0 +1,76 @@
...
+
+- Create an `Operation` of type `OPERATION_UPGRADE`. This will result in a document with a state object of `{ title: "My Todo List" }`, in the default scope (`global`).
+
+### Document Model Upgrade Flow
+
+For documents that already exist, we provide a specific upgrade flow. This requires:
+
+- Named `Action`s for each version upgrade.
```
acaldas, line 53: We could keep the same action name, but include the version in the action input. The reducer would throw an exception if the version is not correct.

packages/reactor/docs/planning/Operations/upgrades.md
```diff
@@ -0,0 +1,76 @@
...
+  dispatch?: SignalDispatch,
+  options?: ReducerOptions,
+) => TDocument;
+```
+
+These accept a `VersionedDocument` instead of a `PHDocument`, which contain side-by-side versions of the same document.
+
+### Document Model Package
+
+Each document model NPM package contains all the code necessary to move from version to version. This means that it must include multiple versions of the same model.
```
acaldas, line 75: I'm a bit reticent about this approach, we could rely on the package manager instead to keep multiple versions of a document model.
We'll need to go through implementing an upgrade to see how it plays out.

packages/reactor/docs/planning/Jobs/index.md
thegoldenmule: That's a good call-out. I don't have a plan there other than "the queue worries about that".

packages/reactor/docs/planning/Operations/upgrades.md
thegoldenmule: I am cool with that, it's definitely a better name. I was using `global` throughout to reduce all the refactor work we'd need to do in connect and reducers, but hey I like this better.

packages/reactor/docs/planning/Operations/upgrades.md
thegoldenmule: I was thinking we need separate actions so we would have separate reducers for each version upgrade.

packages/reactor/docs/planning/Operations/upgrades.md
thegoldenmule: I do not disagree. This is what I was trying to bring up with Prometheus but he seemed focused on this approach. I do admit, it is less complex for consumers if it's just one npm package. However, you lose any third party consensus on, say, reducers that haven't changed. I mean, NPM could essentially "soft-"vouch for you that a past reducer hadn't changed because it's versioned with a date.

packages/reactor/docs/planning/Queue/index.md
acaldas: I agree that the two code paths of calling `addActions` and going through the Sync protocol are problematic.

Maybe it would be worth it to have the Sync Protocol be a layer on top of the Reactor/Executor, and use the `addActions` methods instead of adding them directly.

Note that this use-case of adding actions with an index, that can cause a shuffle, might also be useful outside of the Sync Protocol.
For instance, when doing optimistic updates in the UI or on a script, where a reducer is run locally and the operations get batched before sending them to the actual Reactor to be stored.

Expanding on one use case:
When a document is being edited in Connect, we want to prevent operations coming from switchboard to cause a rerender and lose the user's ongoing edit.
To deal with this, we have to create a copy of the document while the user is editing and only sync when he's done. 
His edits can still be run with a local reducer, so he gets immediate feedback and the document state is updated, but then there has to be a merge with the operations that came from the sync protocol.
I guess the `branching` feature will come into play here but it's not very clear to me how.

packages/reactor/docs/planning/Jobs/index.md
acaldas: I think we'll need to check with Prometheus how to look at mismatching hashes.
From past discussions, I think the idea is to make that a concern of the sync/listeners.
A Reactor/Job executor just receives operations and stores them.

packages/reactor/docs/planning/Operations/upgrades.md
acaldas: I think it's safer to have a breaking change instead of reusing the `global` scope. Safer as in the errors should pop up sooner ahah

thegoldenmule: Actually, we already have `header`. I think `SET_DOCUMENT_NAME` and `SET_META` have to affect the `header`. This scope is what is auto-populated and would be the "low-bandwidth" scope that we, for instance, populate a folder structure with.

packages/reactor/docs/planning/Operations/upgrades.md
acaldas: Eventually we'll have our own package registry which could add those assurances.
Potentially, with hashes of the code inside the reducer signed by the author.

packages/reactor/docs/planning/Jobs/index.md
thegoldenmule: The sync system can still deal with mismatching hashes, it's just that it does it by receiving a failed job, not by checking the hash itself.

IMO having the job executor confirm the output of the job executed keeps everything super simple.

packages/reactor/docs/planning/Operations/upgrades.md
thegoldenmule: When the time comes, btw, we should check out https://verdaccio.org/

packages/reactor/docs/planning/Operations/upgrades.md
acaldas: Ah I didn't get to that part yet.
So `header` would also be a scope and be changed with operations, but instead of being inside the `state` object, it is at the top level of the `Document` object?

I like this approach, might be a bit harder to understand for end users but it also means that it will be out of their way if they just care about the state.

Yesterday I started to add the `document` scope and adding it to `state` field was not looking great.

packages/reactor/docs/planning/Queue/index.md
thegoldenmule: In that use case we might consider simply suspending job processing. It's going to be simpler to pause the queue than to reshuffle constantly. Then we just have a single reshuffle at the end.

packages/reactor/docs/planning/Operations/upgrades.md
thegoldenmule: I was actually thinking it was still on the state object, but we could hide that from a user with the `PHDocument` implementation -- doesn't really matter. But yes, the main idea is that it is a scope that is changed with operations.

packages/reactor/docs/planning/Queue/index.md
acaldas: Yeah, operation history pollution can become an issue.
We could create a separate branch for the user edits, then try to do a `rebase` before merging the branch operations. For many cases, a `rebase` might be cleaner to do instead of a `shuffle`. Also seems easier to show any conflicts to the user for him to fix before merging the branches.

packages/reactor/docs/planning/Jobs/index.md
acaldas: But if the hashes don't match, does the job executor still store the operations and then emits a `jobFailed` event, or aborts the job and forgets the operations?
I think that's the main question here.

packages/reactor/docs/planning/Synchronization/index.md
```diff
@@ -0,0 +1,383 @@
...
+// when something is added to our inbox...
+channel.inbox.added(handle => {
+    // create a job from the handle
+    const job = createJob(handle);
+
+    // queue the job for execution
+    await queue.enqueue(job);
+
+    // watch for local job completion
+    const result = await jobs.watch(job.id);
```
acaldas, line 361: I think the Channel and JobHandle abstractions make a lot of sense.
Looks easy to implement new communication protocols while keeping it separate from the sync logic.

To go back to the idea of not having a special entry point for the Sync Protocol, these 3 lines look like they could be replaced with a `reactor.addActions` call from the legacy Reactor implementation. It also creates a job and waits for it to resolve.

The graphql resolver for the PushUpdates mutation is just a wrapper of the `queueOperations` method:
https://github.com/powerhouse-inc/powerhouse/blob/0fe0d0a73c8a5cc8a00e80079bf3f52697c6a7d7/packages/reactor-api/src/graphql/drive/index.ts#L392-L425
https://github.com/powerhouse-inc/powerhouse/blob/0fe0d0a73c8a5cc8a00e80079bf3f52697c6a7d7/packages/reactor-api/src/sync/utils.ts#L20-L33

packages/reactor/docs/planning/Jobs/index.md
thegoldenmule: No, only successful operations are stored. The error is propagated on the JobResult for "other things" to handle. This should be the exhaustive list of job errors that can happen:

```
export const JobErrorCodes = {
  SIGNATURE_MISMATCH: 90001,
  HASH_MISMATCH: 90002,
  LIBRARY_ERROR: 90003,
  GRACEFUL_ABORT: 90010,
} as const;
```

packages/reactor/docs/planning/Synchronization/index.md
thegoldenmule: The difference here is that the Reactor API I'm proposing does not expose an API that takes Operations -- only one that takes Actions. It also only exposes an scheduling API -- no wait mechanism. This not only keeps the reactor API much smaller / simpler, it should also prevent abuse, and reduce the number of possible edge cases.

The `createJob` that is made up in this function, supposes that it creates a `Job` with the resulting hash on it so that the Job executor can verify the operation results.

The intention is to create a good balance between separation of concerns (reactor input should never be operations, fully async API with job queueing) and consistency (job system is the only thing that pushes to operation store, and it can keep this push inside an atomic transaction).

packages/reactor/docs/planning/Jobs/index.md
thegoldenmule: Check out the latest: https://github.com/powerhouse-inc/powerhouse/blob/feat/reactor/packages/reactor/docs/planning/Synchronization/index.md#error-handling

I've started to document every potential error flow for sync.

packages/reactor/docs/planning/Jobs/index.md
thegoldenmule: I took a pass at the authorization scope here:

https://github.com/powerhouse-inc/powerhouse/blob/feat/reactor/packages/reactor/docs/planning/Auth/index.md#scope-access

packages/reactor/docs/planning/Operations/index.md
acaldas: I'd do it at the IDocumentView just to make the OperationStore write lighter, as it happens inside a Write Lock.
Another advantage is that we can tune the query methods of the IDocumentView to allow selecting, or not, the `resultingState`.

thegoldenmule: If we put the state only in the document view, then because it's only a projection of the operations store, it does not have the state information unless it replays the operations to get it. It would be tempting just to pass the state through the in-memory event bus to the view, but this doesn't really help out as there are always cases in which we replay operations (say the tab is closed or switchboard goes down or there is a network issue, etc). This means we'd be double applying operations: once in the job executor and once in the document view. Then we'd need to add the same error handling around the document view as the job executor.

We keep saying "Write Lock" but in reality this will simply use optimistic locking, so it should not be an issue.

acaldas: On the last call we talked about each operation resulting on an event (ie. `OPERATION_ADDED`) with the `resultingState` and `previousState`, which would be consumed through the event bus. The dev would also have the option to dispatch other events from the reducer, like `ADD_CHILD_DOCUMENT`.

The events sound like the right place to store the state: "The action was processed and this was the resulting state".

**As I was trying to think of a way to keep the `resultingState` out of the Operations, I realised that it does seem to be the simplest solution. We also need to store the user dispatched events though.**

Any other solution, like an Outbox for the worker, would make the atomic operation more complex and error prone. We can think about it later if it turns out it's needed.

As a side note, from the cpu profiles I did in the past, stringifying and hashing was the big bottleneck, especially as state grew. The actual reducer code is a lot more performant since it's usually simple pure code.
Replaying operations without verifying the hash does make me worried about bugs in the reducers creating conflicts between the OperationsStore and the projections.

acaldas: I've brought up relying on **optimistic locking** some times before, but Prometheus was against it and preferred relying on locks. We should bring it up today in the call.

packages/reactor/docs/planning/Jobs/index.md
acaldas: Ok, so the sync protocol will be responsible for keeping the failed job and dealing with it.
If I understand correctly, the Reactor trying to push the mismatching Operation, will keep a separate branch just for the Reactor syncing, where the user's operations are rebased.
However a mismatching hash will probably only be fixed with an hotfix. So a migration and reconciliation will need to happen.

The sync protocol should be somewhat resilient to hash mismatches, where the operations still get synced even if reactors get different results. They could be marked as `inconsistent` so that the user is aware. We could make this behaviour be configurable, by setting the `strictness` level of the sync.

The use cases we'll be doing in the near future will be the usual Client-Server sync, like in Arbgrants.
Thinking about that issue with the user in Chile, Connect just stopped syncing with Swichboard which made the app useless. If the operations were still synced, he would still be able to use the app, except he'd see different dates than what is shown in Switchboard. This seems a better UX.

packages/reactor/docs/planning/Synchronization/index.md
```diff
@@ -0,0 +1,535 @@
...
+```
+
+The `ChannelError` object has a `source` property that indicates the source of the error (could be inbox, outbox, or channel).
+
+The `error` property can be one of two types:
+
+- **JobError** - these are propagated job execution errors, documented in the [Jobs interface](../Jobs/interface.md).
+- **InternalChannelError** - these are errors from the channel itself, not the job.
+
+### Job Errors
```
acaldas, line 447: Another possible error is `MISSING_OPERATIONS`. The `mailbox` might guarantee that it doesn't happen, but I think we should still handle it to be safe. The handling is simple as it just needs to add the missing operations to the outbox.


Another thing that I've seen happening, that we should consider an error and deal with, is clients triggering big shuffles. Like pushing conflicting operations with index `0`, which shuffles the full operations history, that then needs to be synced with other clients, possibly generating a shuffle loop.

This is likely coming from a bug in the current implementation, but it can happen with normal usage if a user is offline for some time.
The fix would be for the pushing reactor to pull the current operations and rebase his operations instead.
The shuffle limit could be configurable.
Some ideas for the name: `EXCESSIVE_SHUFFLE`, `SHUFFLE_LIMIT_EXCEEDED`, `OUTDATED_PUSH`, `REBASE_REQUIRED`.

packages/reactor/docs/planning/Synchronization/index.md
acaldas: But the resulting hash is not enough, the executor also needs to accept actions with indexes and signatures to be able to shuffle and verify the signatures. Right?
That is what I mean with `addOperations`.

packages/reactor/docs/planning/Synchronization/index.md
thegoldenmule: These are great, yah. For large shuffles we could just auto-branch. The user would then be responsible for iitiating rebase flow.

packages/reactor/docs/planning/Operations/index.md
thegoldenmule: 1. Optimistic locking - I'll add a section that details the alternatives and then we can bring it up with Prometheus.

2. Dispatched events - let me add a section to the spec on that where we can discuss in more detail.

3. CPU bottleneck - good to know. As the job executor is built to be the thing that scales out horizontally, putting the load there is ideal. Having good benchmarks we can compare commit-to-commit will be essential.

thegoldenmule: Links:

- For locking considerations, see [IOperationsStore](https://github.com/powerhouse-inc/powerhouse/pull/1573/commits/55a93d12b09605327b67a5636c5d7df1d8f67f9f).
- For dispatched actions, see the [System Stream](https://github.com/powerhouse-inc/powerhouse/pull/1573/files#diff-92412a25905a1627010bc145b8235dece1b9482e8dcf5ad3a5554baa78cd5c80R223) and [Action Dispatches](https://github.com/powerhouse-inc/powerhouse/pull/1573/files#diff-92412a25905a1627010bc145b8235dece1b9482e8dcf5ad3a5554baa78cd5c80R244) sections of the Operations spec.

packages/reactor/docs/planning/Synchronization/index.md
thegoldenmule: All actions submitted to a reactor already have signatures, but you are right that there is no index on them!

I think we should address this by putting the previous index on the actions submitted, rather than the index of the operation generated. Then action submission always looks the same.

packages/reactor/docs/planning/Synchronization/index.md
thegoldenmule: I added these.

packages/reactor/docs/planning/Jobs/index.md
thegoldenmule: That's a good idea, I'll make a note of it.

packages/reactor/docs/planning/Synchronization/index.md
thegoldenmule: To clarify: I've added `prevOpIndex` and `prevOpHash` to `Action`. My goal here is to keep `Action` and `Operation` distinct: keeping `Action` completely immutable before the resulting `Operation`.

I've also added a `nonce` to the signature payload. `prevOpIndex` ensures ordering, `prevOpHash` ensures state, and a `nonce` is both good security hygiene (an attacker could craft an `Action`, the signing of which reveals information about the private key) and can be used in the future against replay attacks in the case of no-ops (or any action that does not result in a state change but may have other side effects).

PR #1660, opened 2025-07-01
feat(connect): enable external processors

gpuente:
## Description:

- enable external processors


gpuente:
Hey @acaldas, there were no details in the [ticket](https://github.com/powerhouse-inc/powerhouse/issues/1636) related to this PR.  
Could you let me know if there's anything missing?

- Should we add some kind of config in the processors to let developers decide whether a processor should be plugged into Connect? So they can export multiple processors but only plug in the ones they need?
- I'm not sure about operational processors;  are they going to run the same way? Same DB? Should we use a different provider? Should we distinguish between operational and analytics processors?

apps/connect/src/context/reactor-analytics.tsx
```diff
@@ -110,6 +126,41 @@ export function DriveAnalyticsProcessor() {
...
+            !manager ||
+            hasRegistered.current ||
+            externalProcessors.length === 0
+        ) {
+            return;
+        }
+
+        hasRegistered.current = true;
+
+        console.log('>>> registering external processors', externalProcessors);
```
acaldas, line 147: Avoid `console.log`, use `logger.verbose` or `logger.debug` instead.

apps/connect/src/context/reactor-analytics.tsx
```diff
@@ -110,6 +126,41 @@ export function DriveAnalyticsProcessor() {
...
+        let index = 0;
+        for (const processor of externalProcessors) {
+            registerExternalProcessors(
+                manager,
+                store.data,
+                `external-processor-${index}`,
```
acaldas, line 153: Let's add the package name to make it easier to identify the processors.

acaldas (changes requested):
> Hey @acaldas, there were no details in the [ticket](https://github.com/powerhouse-inc/powerhouse/issues/1636) related to this PR. Could you let me know if there's anything missing?
> 
> * Should we add some kind of config in the processors to let developers decide whether a processor should be plugged into Connect? So they can export multiple processors but only plug in the ones they need?

Let's load all by default for now. We just need to ensure that trying to import a processor that uses `node` modules does not break the app.

> * I'm not sure about operational processors;  are they going to run the same way? Same DB? Should we use a different provider? Should we distinguish between operational and analytics processors?

The host app doesn't know what a processor will do. It should provide both a AnalyticsStore and a OperationalStore. On Connect, both of these will be on the same PGLite instance. This instance is created in the `createPgLiteFactoryWorker` method on Connect. This will need to be extracted out of the reactor-analytics context so it can be reused by both.

I'll create a separate ticket to create a Operational DB in Connect.

apps/connect/src/context/reactor-analytics.tsx
```diff
@@ -8,6 +8,10 @@ import {
 import { childLogger } from 'document-drive';
 import type { ProcessorManager } from 'document-drive/processors/processor-manager';
 import { useEffect, useRef, type PropsWithChildren } from 'react';
+import {
```
acaldas, line 11: Let's rename this context to `processor-manager`.
It's responsible for instantiating the pglite db and then the AnalyticsStore and the OperationalStore. I don't think it's worth it to create 3 separate contexts for now.

For this ticket, OperationStore is not needed yet.

PR #1668, opened 2025-07-07
Operational Query React integration

acaldas:


packages/reactor-browser/src/pglite/hooks/usePGlite.ts
```diff
@@ -0,0 +1,60 @@
...
+  error: Error | null;
+}
+
+const defaultPGliteState: PGliteState = {
+  db: null,
+  isLoading: true,
+  error: null,
+};
+
+export const usePGliteDB = () => {
```
acaldas, line 18: After the rest of the feature is implemented, I'd like to try and replace this with jotai, following the same pattern of the reactor initialization:
https://github.com/powerhouse-inc/powerhouse/blob/c652ceb821606db65b134fe738743ac0215d78c2/packages/common/state/atoms.ts#L20-L65

Not a blocker for this demo, we can move forward with this approach for now.

packages/reactor-browser/src/pglite/hooks/usePGlite.ts
```diff
@@ -0,0 +1,60 @@
+import { type PGliteWithLive } from "@electric-sql/pglite/live";
+import { useEffect, useState } from "react";
+
+export const PGLITE_UPDATE_EVENT = "pglite-update";
```
acaldas: Let's add a `ph:` prefix to the event to avoid conflicts.

packages/reactor-browser/src/pglite/hooks/usePGlite.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export const usePGlite = () => {
+  const pglite = usePGliteDB();
+  const setPGlite = useSetPGliteDB();
+
+  return [pglite, setPGlite];
+};
+
+declare global {
```
acaldas: Declare this in `packages/reactor-browser/src/global/types.ts` instead, inside `PowerhouseGlobal`

packages/reactor-browser/src/pglite/README.md
```diff
@@ -0,0 +1,764 @@
...
+const useTypedQuery = createTypedQuery<MyDatabase>();
+```
+
+### 3. Use It in Your Component
+
+```typescript
+// Static query (no parameters)
+export function useUserList() {
+  const result = useTypedQuery(db => {
+    return db.selectFrom('users').selectAll().compile();
```
acaldas: Since we have first-class support for `kysely`, let's also support return the query with calling `compile()`.
The example can just be a one liner:
```ts
export function useUserList() {
  return useTypedQuery(db => db.selectFrom('users').selectAll());
}
```

packages/reactor-browser/src/pglite/hooks/useOperationalDB.ts
```diff
@@ -0,0 +1,37 @@
+import { type PGlite } from "@electric-sql/pglite";
```
acaldas: Let's have a separate folder for  the `Operational` module. The `pglite `module would be for internal use, external devs import from `reactor-browser/operational` and `reactor-browser/analytics`.

packages/reactor-browser/src/pglite/hooks/useOperationalDB.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Kysely } from "kysely";
+import { PGliteDialect } from "kysely-pglite-dialect";
+import { useMemo } from "react";
+import { usePGliteDB } from "./usePGlite.js";
+
+export interface OperationalDB<Schema> {
```
acaldas: `OperationStore` instead of `OperationalDB` to keep consistency with `AnalyticsStore`.

packages/reactor-browser/src/pglite/hooks/useOperationalDB.ts
```diff
@@ -0,0 +1,37 @@
...
+  db: Kysely<Schema> | null;
+  isLoading: boolean;
+  error: Error | null;
+}
+
+export const useOperationalDB = <Schema>() => {
+  const pglite = usePGliteDB();
+
+  const operationalDB = useMemo<OperationalDB<Schema>>(() => {
```
acaldas: When we implement pglite with `jotai`, this will just be an atom derived from the PGlite atom. Avoiding all the react rendering pitfalls.

packages/reactor-browser/src/pglite/hooks/usePGlite.ts
gpuente: fixed

packages/reactor-browser/src/pglite/hooks/usePGlite.ts
gpuente: fixed

packages/reactor-browser/src/pglite/hooks/useOperationalDB.ts
gpuente: fixed

packages/reactor-browser/src/pglite/hooks/useOperationalDB.ts
gpuente: fixed

packages/reactor-browser/src/pglite/README.md
gpuente: I did it this way because we can support both `Kysely` and raw `SQL` approaches:

```tsx
// Kysely users
export function useUserList() {
  return useTypedQuery(db => db.selectFrom('users').selectAll().compile());
}

// Raw SQL users (can ignore Kysely completely)
export function useUserList() {
  return useTypedQuery((_, params) => ({
    sql: `SELECT * FROM users WHERE active = ${params.active}`,
    parameters: { active: true }
  }));
}
```

This way users not familiar with `Kysely` can write their queries in raw `SQL`, while Kysely users get the familiar `.compile()` pattern. Both return the same { sql, parameters } format, so the implementation stays simple and predictable.

let me know what do you think

packages/design-system/src/connect/components/toast/toast.tsx
```diff
@@ -85,6 +85,8 @@ const CloseButton: ToastContainerProps["closeButton"] = ({ closeToast }) => (
   </button>
 );
 
-export const ToastContainer: React.FC<ToastContainerProps> = (props) => (
-  <RToastContainer closeButton={CloseButton} {...props} />
-);
+export const ToastContainer: React.FC<ToastContainerProps> = (props) => {
+  console.log("ToastContainer");
```
acaldas: Stray console.log

packages/reactor-browser/src/operational/hooks/useLiveQuery.ts
```diff
@@ -0,0 +1,55 @@
...
+import { useEffect, useState } from "react";
+import { usePGliteDB } from "../../pglite/hooks/usePGlite.js";
+import { useOperationalStore } from "./useOperationalStore.js";
+
+export type QueryCallbackReturnType = {
+  sql: string;
+  parameters?: readonly unknown[];
+};
+
+export function useLiveQuery<Schema, T = unknown, TParams = undefined>(
```
acaldas: Let's name it `useOperationalQuery` to make it clear.

packages/reactor-browser/src/operational/hooks/useLiveQuery.ts
```diff
@@ -0,0 +1,55 @@
...
+    db: Kysely<Schema>,
+    parameters?: TParams,
+  ) => QueryCallbackReturnType,
+  parameters?: TParams,
+) {
+  const [result, setResult] = useState<LiveQueryResults<T> | null>(null);
+  const [queryLoading, setQueryLoading] = useState(true);
+
+  const pglite = usePGliteDB();
+  const operationalDB = useOperationalStore<Schema>();
```
acaldas: Can we make it so that this hook only calls `useOperationalStore`, without needing to call the pgliteDB too?

packages/design-system/src/connect/components/toast/toast.tsx
gpuente: fixed

packages/reactor-browser/src/operational/hooks/useLiveQuery.ts
gpuente: fixed

packages/reactor-browser/src/operational/hooks/useLiveQuery.ts
gpuente: fixed

PR #1680, opened 2025-07-09
fix(reactor-api): update document exopect revision

gpuente:


gpuente:
hey @acaldas,
I changed the `responseForDocument` here https://github.com/powerhouse-inc/powerhouse/commit/ab00dc3c1b660de509f376306cf8ec3edcf5ebc8
because it was returning a revision that didn't match the `IDocument` interface:

```graphql
interface IDocument {
      id: String!
      name: String!
      documentType: String!
      revision: Int!
      created: DateTime!
      lastModified: DateTime!
      operations(first: Int, skip: Int): [Operation!]!
      stateJSON: JSONObject
    }
```
https://github.com/powerhouse-inc/powerhouse/blob/596aedbd58f2096e1ad9e65cb88d0d430aea562b/packages/reactor-api/src/utils/create-schema.ts#L164-L173

now I'm updating the test in this PR, but I wanted to confirm with you before continue...

PR #1703, opened 2025-07-21
feat(connect,reactor-browser,common,state):  use new state hooks

ryanwolhuter:


apps/connect/src/components/router.tsx
```diff
@@ -34,13 +31,6 @@ function createRoutes() {
     const routes: RouteObject[] = [
         {
             index: true,
-            element: (
```
ryanwolhuter, line 37: we can still render the same content for the home page as before, but by treating this route as the main route, we are able to maintain the context for the atoms.

packages/state/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "document-drive": "workspace:*",
+    "document-model": "workspace:*",
+    "jotai": "^2.12.5",
+    "jotai-devtools": "^0.12.0",
+    "jotai-effect": "^2.0.4",
+    "jotai-optics": "^0.4.0",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1"
+  },
+  "peerDependencies": {
```
acaldas, line 34: Should these be `peerDependencies`? Particularly the `jotai` ones.
Wouldn't it be simpler to make them normal `dependencies`?
For example, in this PR `Connect` adds `@powerhousedao/state` as a dependency but doesn't add `jotai-optics`.

packages/state/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/powerhouse-inc/powerhouse"
+  },
+  "scripts": {
+    "build:tsc": "tsc --build",
+    "lint": "eslint"
+  },
+  "exports": {
+    ".": "./dist/index.js"
```
acaldas: Let's allow separate exports with `"./*": "./dist/*.js"`.
Also let's add `"sideEffects": false` so bundlers tree shake imports from `index.js`.
I'm not sure if it's a webpack only thing, but it's worth is as a good practice.

packages/common/editors/generic-drive-explorer/generic-drive-explorer.stories.tsx
```diff
@@ -1,14 +1,17 @@
...
+  Editor as EditorStoryComponent<DocumentDriveDocument>,
+);
 
-const meta: Meta<typeof Editor> = {
+const meta = {
   ..._meta,
   title: "Generic Drive Explorer",
-};
-export const Empty: StoryObj<typeof Editor> = CreateDocumentStory;
+} as Meta<typeof Editor>;
```
acaldas, line 14: Not sure why this was needed but storybook types have been a pain.
Does `satisfies Meta<typeof Editor>` work here?

acaldas (commented):
Nice to see things centralizing in the state instead of being on multiple React rendering methods.

As a nice-to-have, it would be good to already create a test setup on the `state` package. For example, with a simple test that checks if  `useDocument` returns the correct document and is triggered when the document updates.

`reactor-browser` has some tests using `renderHook` from `vitest-browser-react`

acaldas (approved):

PR #1754, opened 2025-08-13
feat(state,connect,reactor-browser): eliminate jotai

ryanwolhuter:
We have established a pattern which causes many problems and offers no benefits.

1. load reactor
2. Copy reactor into window object
3. load reactor data
4. copy reactor data into atom in connect
5. copy data from atom into window object
6. copy data from window object into atom in project package

this whole process is necessary if we want to use atoms and keep their data in sync across connect and external packages, because _jotai cannot share a store across apps built in separate environments_.

We can bypass all of this by accepting that since we are going to copy data into the window object anyway (for syncing across connect and packages), we might as well just use the window object data as the source of truth and do away with jotai entirely.

we can instead do this:

1. load reactor
2. load reactor data
3. copy reactor and its data into window
4. use this reactor instance and data everywhere

this allows us to use a unified data strategy for everything:

- reactor
- renown
- user
- crypto
- drives
- documents
- nodes
- document models
- editors
- vetra packages
- app config

with this in place, we can finally do away with the cause of so many problems with connect, the `useDocumentDriveServer` hook.

this hook provides many pieces of data that drive explorers and editors need to work. Being a react hook, it can only be used inside of react and the data it provides cannot be used outside of the react tree in which it is invoked. This has meant we must pass this data as props and/or context instead of providing it via hooks.

However, the only reason this hook is a hook is because it uses other hooks. And when we look at the other hooks it uses, those hooks are only hooks because they use data from window or env that is wrapped in jotai.

_if we don't need jotai, we don't need any of this_.

By eliminating jotai and using the new reactor implementation which no longer separates queueAction and queueDriveAction, we can ship all of these functions as regular code _that does not need to live inside a hook_.

This has many fantastic benefits, but the most important one is this:

We can now ship dispatch functions which wrap the crypto code etc in normal functions, which can be encapsulated inside hooks in reactor-browser.

Users can now do this:

```ts
const driveId = 'my-drive-id'
const documentId = 'my-document-id'
const [drive, driveDispatch] = useDriveById(myDriveId)
const [selectedDrive, selectedDriveDispatch] = useSelectedDrive()
const [document, documentDispatch] = useDocumentById(myDocumentId)
const [selectedDocument, selectedDocumentDispatch] = useSelectedDocument()
... etc
```


apps/connect/package.json
```diff
@@ -120,7 +117,7 @@
         "vite": "^6.3.3",
         "vite-envs": "^4.6.0",
         "vite-plugin-html": "^3.2.2",
-        "vite-plugin-node-polyfills": "^0.23.0",
+        "vite-plugin-node-polyfills": "^0.24.0",
```
ryanwolhuter, line 97: had to update these because it all of a sudden got typed as `any` at the import

apps/connect/src/components/analytics.tsx
```diff
@@ -1,6 +1,6 @@
+import connectConfig from '#connect-config';
 import { useEffect } from 'react';
 import { useAcceptedCookies } from '../hooks/useAcceptedCookies';
-import { useConnectConfig } from '../hooks/useConnectConfig';
```
ryanwolhuter, line 3: I've removed this hook since it was useless. all it did was `import connectConfig from '#connect-config';` and then return exactly that.

apps/connect/src/components/app-loader.tsx
```diff
@@ -1,7 +1,4 @@
 import { useLoadInitialData, useRenown } from '#hooks';
-import { Provider } from 'jotai';
```
ryanwolhuter, line 2: we no longer need a jotai provider or its dev tools

apps/connect/src/components/app-loader.tsx
```diff
@@ -19,23 +16,19 @@ const ModalManager = lazy(() =>
 function Load() {
     useLoadInitialData();
     useLoadData();
-    useRenown();
```
ryanwolhuter, line 22: we need clarity on how renown is loaded. currently there is no logic in the reactor initiation that properly waits for it to be initialized before proceeding.

we can wrap window.renown in a `useSyncExternalStore` as I have done elsewhere, but I think it would be better for the reactor to just be able to add renown once it loads and handle these things separately.

apps/connect/src/components/app-loader.tsx
```diff
@@ -19,23 +16,19 @@ const ModalManager = lazy(() =>
...
     return null;
 }
 
 export const AppLoader = (
     <StrictMode>
-        <Provider>
```
ryanwolhuter, line 28: jotai provider removed

apps/connect/src/components/document-editor-container.tsx
```diff
@@ -23,15 +23,11 @@ export function DocumentEditorContainer() {
...
+    const parentFolder = useParentFolder(selectedDocument?.header.id);
     const documentType = selectedDocument?.header.documentType;
     const isRemoteDrive = useDriveIsRemote(unwrappedSelectedDrive?.header.id);
     const remoteUrl = useDriveRemoteUrl(unwrappedSelectedDrive?.header.id);
-    const getDocumentModelModule = useGetDocumentModelModule();
-    const documentModelModule = documentType
-        ? getDocumentModelModule(documentType)
-        : undefined;
-    const setSelectedNode = useSetSelectedNode();
+    const documentModelModule = useDocumentModelModuleById(documentType);
```
ryanwolhuter, line 29: this function should really be called `useDocumentModelByNameOrDocumentTypeOrIdOrSpecIdOrAnythingElseReallyIDontKnow()` but I have left it as byId for now

apps/connect/src/components/document-editor-container.tsx
```diff
@@ -73,14 +65,14 @@ export function DocumentEditorContainer() {
                     },
                     onContinue(closeModal) {
                         closeModal();
-                        return exportFile(document, getDocumentModelModule);
```
ryanwolhuter, line 76: this exportFile function shouldnt even need the whole document model module, much less a function which returns one. for now i have at least made it so that it doesnt have this redundant copying and complexity

apps/connect/src/components/drive-editor-container.tsx
```diff
@@ -95,25 +86,25 @@ export function DriveEditorContainer() {
...
+    const driveEditor = useDriveEditorModuleById(
         selectedDrive?.header.meta?.preferredEditor,
     );
+    const defaultDriveEditor = useDefaultDriveEditorModule();
 
     const DriveEditorComponent =
-        driveEditor?.Component ?? GenericDriveExplorer.Component;
+        driveEditor?.Component ??
+        defaultDriveEditor?.Component ??
+        GenericDriveExplorer.Component;
```
ryanwolhuter, line 46: we are now loading the generic drive explorer code twice, once in the default drive editor and then again in this fallback to GenericDriveExplorer. I would remove the fallback, but I am not confident that the system is stable enough not to break in some cases if i do.

apps/connect/src/components/modal/modals/AddDriveModal.tsx
```diff
@@ -16,7 +17,7 @@ type Props = {
 export function AddDriveModal(props: Props) {
     const { open, onAddLocalDrive, onAddRemoteDrive, onClose } = props;
     const user = useUser();
-    const apps = useApps();
```
ryanwolhuter, line 19: what we have been calling apps are just a special case of editor. they use the exact same code, but have a known documentType of `powerhouse/document-drive`. 

There is no reason for us to have a whole set of special code for handling this case, and calling these drive editors "apps" in our source code is needlessly confusing.

apps/connect/src/components/modal/modals/settings/default-editor.tsx
```diff
@@ -15,8 +14,9 @@ const documentModelEditorOptions = [
 ] as const;
 
 export const DefaultEditor: React.FC = () => {
-    const [documentModelEditor, setDocumentModelEditor] =
-        useDefaultDocumentModelEditor();
+    const [documentModelEditor, setDocumentModelEditor] = useState<
```
ryanwolhuter, line 17: i have done this as a quick fix to silence the error, but this code should just be deleted.

we have not used the v1 document model editor in months, and as far as i am aware we are literally not even loading it in the app so selecting it here just breaks

apps/connect/src/components/modal/modals/settings/package-manager.tsx
```diff
@@ -119,10 +125,23 @@ export const PackageManager: React.FC = () => {
 
     return (
         <BasePackageManager
-            mutable={isMutable}
+            mutable={true}
```
ryanwolhuter, line 128: we need to clarify what is up with this `mutable` property. all it does is check if hmr is available.

apps/connect/src/globals.ts
```diff
@@ -1,15 +1,26 @@
...
 import type { DocumentEditorDebugTools } from './utils/document-editor-debug-tools.js';
-export {};
 
 declare global {
     interface Window {
         electronAPI?: ElectronAPI;
         connectCrypto?: IConnectCrypto;
         renown?: IRenown;
         documentEditorDebugTools?: DocumentEditorDebugTools;
+        reactor?: IDocumentDriveServer;
```
ryanwolhuter: we are storing these references in the window. there are others, but those types are kept in the state package to avoid polluting this interface.

apps/connect/src/hooks/useConnectCrypto.ts
```diff
@@ -59,11 +58,8 @@ export function useConnectCrypto(): IConnectCrypto {
     );
 }
 
-const didAtom = atom<DID | undefined>(undefined);
-didAtom.debugLabel = 'didAtomInConnect';
-
 export function useConnectDid(): DID | undefined {
```
ryanwolhuter: this works as is without an atom. it is a bit wasteful since it is done on every use of the hook, but really it is just a string so it makes no real difference in performance. this can be updated to use my `useSyncExternalStore` solution if needed

apps/connect/e2e-tests/app.spec.ts
```diff
@@ -1,46 +0,0 @@
-import { expect, type Page, test } from '@playwright/test';
```
ryanwolhuter, line 1: @gpuente these arent the e2e tests that you added recently right? I removed them because they were referencing electron stuff that has been broken for months. git blame says these files haven't changed in more than a year

PR #1776, opened 2025-08-22
fix(monorepo): linting and type checking

ryanwolhuter:
# Linting

This PR says there are many changes, but almost all of them are just automatically applied lint fixes. We need to bite the bullet and accept that this must be done. If we don't, our build tools will never give consistent results.

I have reduced the number of linting rules significantly by changing our config to only use recommended rules. Much of the other rules are just stylistic and our team is quite easy going with regards to those things, so there is no need for constant warnings in the console about them.

# Circular references

The rest of the changes are to deal with circular references. This is one of the biggest causes of our inconsistent builds and constant errors.

Up until now we have relied on build tools to handle the imports/exports of our very complex codebase, much of which contains naming clashes and circular re-exports. These tools are not sophisticated enough to handle this, and they all have their own ways of dealing with ambiguity. This means that as long as we need multiple build tools, we will either have to simplify and handle this or else accept that we will never be consistent.

# Package name imports

The vast majority of these circular reference issues can be resolved simply by using javascript modules' default behavior when an "exports" field is present in package.json — this lets you do import { thing } from "@my-org/my-package" both in other places in the monorepo, and crucially, inside the actual package itself. At first this might seem like it would also lead to circular references, but in fact it actually fixes them. By always importing everything with the package name style, javascript always has a single source for any given piece of code which it can use unambiguously.

# Naming clashes

As soon as I started fixing these circular import/exports, our tools were able to show that there are many, many naming clashes in our codebase. The only reason these errors did not show up before is that our build tools were just taking whichever one of the ambiguous names its algorithm picked, and there was no reason to assume that two tools will make the same choice.

We can't keep using the same names for things, for the sake of our own sanity but also because our build tools can't keep up with it. 

# Node code exports

We have been relying on our tools to figure out which runtime they are in and to choose which code to run accordingly. This can work and be convenient in a less complex and interdependent system, but in our case it is just never going to be able to handle all cases. This is especially true for node and browser code, where it can easily happen that we are using a node process to run browser code, or we want to import some code from a library with node code in it.

Having the runtime sort it out for us might seem convenient at first, but really it just pushes the hassle down the road a little bit. We end up having to write more and more code to handle the edge cases where this solution breaks, and then we end up with even more complexity than we would have otherwise.

In many cases we have code that has the same name and two different implementations for node and browser. This can be useful if you have covered every case, but for us, there are many places where there is only a node or only a browser implementation, and no way of seeing this by looking at the code.

I have chosen here the simplest way of handling this — providing another export in the package.json exports called "/node" and exporting all node-only code from there. We can do the same for browser, but really the vast majority of our code is either isomorphic or browser-based, so to me making node the special case is easier.

# tsc --build

All of our typescript packages had a "build" script intended to compile typescript code, but it would call this "tsc --build". By passing the --build command, we tell typescript to build all of the code for the entire monorepo, not just the code in a given package. This meant that:

1. we were rebuilding the whole thing 20 times when we only needed to do it once
2. we were rebuilding in 20 parallel processes which were all writing to the same files and causing weird behavior

In general what you want to do is always going to be to run "tsc --build" because you want to know if your code works _and_ if other code that uses your code also works. I have provided a "tsc" script in each package which calls "tsc" without --build for what should be the rare special case where you need to build just an individual package's typescript code.


.github/workflows/check-commit.yml
```diff
@@ -8,88 +8,68 @@ on:
...
-          fi
-
-          curl -H "Content-Type: application/json" -X POST "$DISCORD_WEBHOOK" -d '{
-            "embeds": [{
-              "title": "⚠️ Job Failed",
-              "description": "**Author:** '"${MENTION_STRING}"'\n**Commit:** [`${{ github.event.head_commit.message }}`](${{ github.server_url }}/${{ github.repository }}/commit/${{ github.sha }}) \n**Failed targets:** [`'"$FAILED_TARGETS"'`](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})",
-              "color": 16711680
-            }]
-          }'
+      - name: Checkout
```
ryanwolhuter, line 25: I have changed this to use the new scripts in the package.json root. it now does all of the checks including linting and type checking.

I have split these up into separate commands so that they can:

- be cached with each tool's caching strategy
- be resumed from each step if they fail
- show us which parts of the process are slow

.gitignore
```diff
@@ -1,13 +1,17 @@
-node_modules/
+node_modules
+dist
+docs
+.out
+test/**/temp
+.eslintcache
```
ryanwolhuter, line 5: we have a lot of different gitignores all over the place and most of them are redundant. putting things like node_modules and dist in here at the top applies to the whole monorepo.

apps/academy/babel.config.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
```
ryanwolhuter: are unifying the eslint configs and applying auto fixes, there are many changes like this, where things like single quotes are swapped for double etc. we need to do this because otherwise we cannot just run eslint at the top level, since there will be inconsistent results.

apps/connect/.prettierrc
ryanwolhuter, line 1: similarly here, we had a different prettier config for connect. this made autofix always break.

apps/connect/package.copy.json
```diff
@@ -0,0 +1,117 @@
+{
```
ryanwolhuter, line 1: a major source of build problems was the various places where we access a package's package.json file from within the package itself. if you just import the file, typescript will complain because the file is not included in tsconfig.

however, we can't just include a package.json file in the config like that, because when javascript is resolving the node modules in a given package, it walks down the directory tree looking for package.json files. If we have another package.json in the dist folder, then it will treat that one as the new package root. this breaks imports and exports etc.

I have added here a basic solution — just make a copy and add a "postinstall" hook which updates it when things change. this isn't the most robust but at least it stops things from breaking.

apps/connect/package.copy.json
```diff
@@ -0,0 +1,117 @@
...
+        "dist",
+        "nginx.conf",
+        "nginx.sh",
+        "scripts",
+        "assets"
+    ],
+    "exports": {
+        ".": "./src/index.ts",
+          "./config": "./src/connect.config.ts",
+        "./assets/*.mp4": "./assets/*.mp4",
```
ryanwolhuter: doing this lets us have a nice way of importing non-js stuff. I was tempted to do "./assets/*" but this actually creates a lot of problems, since it will then try to do this in some places with javascript code. I think that being explicit here is ultimately a good thing anyway.

apps/connect/package.json
```diff
@@ -91,8 +102,8 @@
         "uint8arrays": "^5.1.0",
         "uuid": "^9.0.1",
         "viem": "^2.8.13",
-        "vite": "^6.3.3",
-        "vite-envs": "^4.6.0",
+        "vite": "^7.1.3",
```
ryanwolhuter: I upgraded to the latest version of vite to see if that helped with some issues but it didn't change anything.

apps/connect/src/components/app.tsx
```diff
@@ -1,26 +1,29 @@
...
+import ProcessorManagerProvider from "../context/processor-manager.js";
+import Analytics from "./analytics.js";
+import { Router } from "./router.js";
 
-if (import.meta.env.MODE === 'development') {
-    window.documentEditorDebugTools = new DocumentEditorDebugTools();
-} else {
-    serviceWorkerManager.registerServiceWorker(false);
-}
+// if (import.meta.env.MODE === "development") {
```
ryanwolhuter: this was giving a "cannot access register service worker before initialization" error, which is the same error we get elsewhere when vite gets confused by circular import references.

apps/connect/src/components/demo/atlas-import.tsx
```diff
@@ -1,180 +1,178 @@
-import AtlasIcon from '#assets/icons/Atlas-Logomark.svg?react';
-import RefreshIcon from '#assets/icons/refresh.svg?react';
-import { Button } from '@powerhousedao/design-system';
-import { addRemoteDrive, useReactor } from '@powerhousedao/reactor-browser';
-import { gql, request } from 'graphql-request';
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { useLocation, useNavigate, useParams } from 'react-router-dom';
-import { toast } from '../../services/toast';
+import AtlasIcon from "@powerhousedao/connect/assets/icons/Atlas-Logomark.svg?react";
```
ryanwolhuter, line 1: Here is an example of how the assets path can be used.

apps/connect/src/components/document-editor-container.tsx
```diff
@@ -1,113 +1,113 @@
...
+  useDriveRemoteUrl,
+  useParentFolder,
+  useSelectedDocument,
+  useSelectedDrive,
+} from "@powerhousedao/reactor-browser";
+import type { PHDocument } from "document-model";
+import { useMemo } from "react";
+import { useTranslation } from "react-i18next";
+import { useModal } from "../components/modal/index.js";
+import { validateDocument } from "../utils/validate-document.js";
```
ryanwolhuter: there are many places like here where these imports should be changed to use the package name import style instead of these relative imports, both for convenience and to prevent further circular import problems.

apps/connect/src/components/modal/modals/settings/about.tsx
```diff
@@ -1,16 +1,16 @@
-import connectConfig from '#connect-config';
-import { About as BaseAbout } from '@powerhousedao/design-system';
-import packageJson from '../../../../../package.json' with { type: 'json' };
+import { connectConfig } from "@powerhousedao/connect/config";
+import { About as BaseAbout } from "@powerhousedao/design-system";
+import packageJson from "../../../../../package.copy.json" with { type: "json" };
```
ryanwolhuter, line 3: we should probably add a dedicated exports root for this

eslint.config.js
```diff
@@ -3,150 +3,279 @@ import { default as eslint } from "@eslint/js";
 import eslintPluginPrettierRecommended from "eslint-plugin-prettier/recommended";
```
ryanwolhuter, line 3: I have made this a source of truth for all of the linting and type checking disables that are happening throughout the code. this makes it so that we can actually see what is happening and stops us from hiding bugs from our tools.

apps/connect/vite.config.ts
```diff
@@ -61,6 +61,8 @@ export default defineConfig(({ mode }) => {
     nodePolyfills({
       include: ["events"],
       globals: {
+        Buffer: false,
+        global: false,
```
acaldas: Disabling `global` polyfill removes the import error on the `RunAsap` module in `DocumentDrive` which checks if some `node` only globals are available.

apps/connect/vite.config.ts
```diff
@@ -115,9 +117,6 @@ export default defineConfig(({ mode }) => {
     build: {
       minify: false,
       sourcemap: true,
-      rollupOptions: {
-        external: ["vite-plugin-node-polyfills"],
```
acaldas: Unsure if this is still needed, Connect builds without it.

packages/document-drive/src/utils/logger.ts
```diff
@@ -43,8 +43,10 @@ export class ConsoleLogger implements ILogger {
   get #levelValue(): number {
     if (this.#levelString === "env") {
       const envLevel =
-        typeof process !== "undefined" && "env" in process
-          ? process.env.LOG_LEVEL
+        typeof globalThis !== "undefined" &&
```
acaldas, line 46: There was also an import error here due to vite trying to polyfill `process`.
Checking for `globalThis.process` makes it work, I suspect that it just doesn't trigger the polyfill.

When we revisit Connect bundling we should try to remove the automatic node polyfills and apply it just to specific dependencies like Knex. This way it will be easier to detect node coding that is being added incorrectly to the bundle.

acaldas:
Had some issues building Connect. Added some comments to the changes I made: https://github.com/powerhouse-inc/powerhouse/pull/1776/commits/f38238b46d9458a0ff33a6629f4e3cbb0ba665cd

package.json
```diff
@@ -3,21 +3,19 @@
...
   "scripts": {
-    "clean": "pnpm nx run-many --target=clean --all",
-    "clean:node_modules": "pnpm nx run-many --target=clean:node_modules --all && rimraf node_modules",
     "prepare": "husky",
-    "build:tsc": "tsc --build",
-    "lint:all": "pnpm nx run-many --target=lint:nx --all",
-    "affected:check": "pnpm build:tsc && pnpm nx affected --target=build,build-storybook,lint:nx,test",
-    "affected:check:ci": "pnpm nx affected --target=build,build-storybook,lint:nx && pnpm nx affected --target=test --exclude=document-drive",
-    "build": "pnpm nx run-many --target=build --all",
-    "build:all": "pnpm nx run-many --target=build --all",
```
acaldas, line 14: I was unsure what commands to run when I checked out the branch.
Would be good to still have a simple `build` npm script for each package and then a `build` script at the root `package.json` to build everything.

Each package would take care of it's own building steps, ie: `"build: `pnpm run build:css && pnpm run tsc && pnpm run build:vite"`.

I know we were relying on configuring the dependencies of tasks on `nx` before, but I see that `pnpm -r` already takes into account interdependencies among the packages by default: https://pnpm.io/cli/recursive#--no-sort

package.json
```diff
@@ -3,21 +3,19 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "clean": "pnpm nx run-many --target=clean --all",
```
acaldas, line 6: Can we still keep this command? with `pnpm -r` instead.
It's helpful for when people did something wrong, I can see it being useful already for people do the transition when this branch gets merged.

apps/connect/vite.config.ts
ryanwolhuter: I think the problem was that dev doesn't work without it but build does work, and then it causes a problem with build when it is there. but if both dev and build work then we are all good and should remove it

packages/document-drive/src/utils/logger.ts
ryanwolhuter: cool yeah I agree. we need to have tight control of this

package.json
ryanwolhuter: I have the `simulate-check-all-ci` command here which does everything, its just a matter of naming.

The reasons we don't want to have everything in one script for the github action:

1. can't use each tool's caching strategy
2. can't resume from where it breaks
3. can't easily see how long each step takes

so I have the script here for simulating the process on your local when you need to, but ultimately there are several steps that the github action would take.

imo it would be even more confusing to have a script called "build" which in fact does so much more, it would check types, lint, transpile, copy paste various things, run tailwind, run vite etc. We could rename `simulate-check-all-ci` to be just `build` but I don't think that would make things any easier for us.

package.json
ryanwolhuter: I have replaced this with the "simulate-fresh-git-clone" script which does the same but better and more complete. This command here only removes some of the artifacts and actually misses quite a lot of other ones like cache files and out dirs etc. My script uses every gitignore in the repo to do this task, making sure it removes everythihng. We could rename that script to be just "clean" and it would have the same result as here.

Also we don't want this to only run inside what nx considers to be packages and projects, we want it to run for the whole monorepo regardless of nx or pnpm or anything else.

package.json
ryanwolhuter: really what we want is to have only one or two github actions instead of a whole bunch of unused ones. then we could have just the one that is "build and check" or something, which anyone can look at to see the exact steps that it does and in what order.

apps/connect/vite.config.ts
acaldas: Ah, I was only trying the build.

package.json
acaldas: I was thinking about local dev only.
Currently, when I checkout a new branch I do `pnpm i` and `pnpm build:all`.
What would be the commands now?
I would keep the naming of the scripts geared for local use by devs, and add new ones for CI if needed.

package.json
acaldas: It looks great, I would just keep the name as `clean` as I've said above.

package.json
ryanwolhuter: as it is now you only need to run `pnpm i`

typescript and eslint will already be working in your editor without you doing anything.

then there are a few random places that run tailwind and some others that run vite, and the vite ones depend on the tailwind ones so tailwind needs to be run first

and then there a bunch of random "build" scripts that really just copy paste files and other strange things.

and then there are a bunch of storybook builds which we have historically kept separate

so a minimal "build" script in the monorepo package.json would do this:

`pnpm tsc && pnpm build:css && pnpm build:vite`

but then this is not doing linting or building storybook or running tests.

what I am getting at here is that as much as it seems convenient to have one "build" command, there are so many things going on here that this command will end up doing a bunch of stuff you just don't want every time you run it. 

Up until now we have just been rolling this all up into that `build:all` command and it has caused all of these problems that we have been seeing make the builds fail. My fear is that if we go back down the road of convenience, we will very quickly end up right back here in build failure hell

apps/connect/src/context/processor-manager.tsx
```diff
@@ -69,39 +55,37 @@ async function registerExternalProcessors(
...
-  const wrappedFactory = (driveHeader: PHDocumentHeader) => {
-    return processorFactory({ analyticsStore })(driveHeader.id);
-  };
-
-  return await manager.registerFactory(
-    "@powerhousedao/diff-analyzer",
-    wrappedFactory,
-  );
-}
+// async function registerDiffAnalyzer(
```
ryanwolhuter, line 64: In one of our discussions we decided to comment this out because it has a weird error, says cannot find "./utils/misc" in "document-drive" even though that import is removed and not used anywhere in the monorepo

apps/connect/src/context/read-mode.tsx
```diff
@@ -1,242 +1,250 @@
...
-    return {
-      ...(ReadModeInstance as Omit<IReadModeContext, "readDrives">),
-      readDrives,
-    };
-  }, [readDrives]);
-
-  return <ReadModeContext.Provider {...props} value={context} />;
-};
-export const useReadModeContext = () => useContext(ReadModeContext);
+// import { useReactor, useUserPermissions } from "@powerhousedao/reactor-browser";
```
ryanwolhuter, line 1: I commented this out because it doesn't seem to be currently in use and it has _a lot_ of problems. needs major work to fix if we intend to use it, probably easier and better to just rewrite

apps/connect/vite.config.ts
```diff
@@ -99,24 +66,17 @@ export default defineConfig(({ mode }) => {
...
-      babel: {
-        parserOpts: {
-          plugins: ["decorators"],
-        },
-      },
-    }),
-    svgr(),
     createHtmlPlugin({
-      minify: true,
+      minify: false,
```
ryanwolhuter, line 80: I can change these back to true, but I honestly believe that leaving it unminified so we can see what's actually ending up in our bundle will have a much bigger and better impact than anything the minification can do

packages/builder-tools/document-model-editor/components/module-form.tsx
```diff
@@ -1,10 +1,10 @@
-import type { Module } from "document-model";
+import type { ModuleSpecification } from "document-model";
```
ryanwolhuter, line 1: I have renamed the Module type to ModuleSpecification because:

Module is already a type in javascript so this causes editor auto imports to break
This type is not actually for a module but rather for the specification of a module

packages/builder-tools/document-model-editor/components/operation-error-form.tsx
```diff
@@ -1,11 +1,14 @@
...
+  OperationErrorSpecification,
+  ModuleSpecification,
+} from "document-model";
 import { useCallback, useRef } from "react";
 import { TextField } from "./text-field.js";
 
 type Props = {
-  operation: Module["operations"][number];
-  error?: DocumentModelOperationError;
+  operation: ModuleSpecification["operations"][number];
```
ryanwolhuter, line 10: Similarly, this type actually is for the specification of the error and not the error itself

packages/codegen/package.json
```diff
@@ -5,33 +5,22 @@
...
-    "clean": "rimraf dist/",
-    "copy-templates": "copyfiles src/codegen/.hygen/templates/**/*.esm.t dist/",
-    "copy-hygen-pkg": "copyfiles src/codegen/.hygen/package.json dist/",
-    "prebuild": "npm run clean",
-    "build": "tsc --build",
-    "postbuild": "npm run copy-templates && npm run copy-hygen-pkg",
-    "prepublishOnly": "npm run build",
+    "tsc": "tsc",
+    "lint": "eslint",
+    "copy-hygen": "cp -r src/codegen/.hygen dist/src/codegen/",
```
ryanwolhuter, line 19: now that I have changed the index files in the hygen folder to be js instead of ts,  this is all we need to do for moving across the template code

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-document-model-module/object.esm.t
ryanwolhuter, line 1: I've deleted the various "object" stuff which was for the old broken class-based code

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-document-model/index.esm.t
```diff
@@ -4,10 +4,8 @@ force: true
 ---
 export * from './actions.js';
 export * from './document-model.js';
-export * from './object.js';
 export * from './types.js';
 export * as actions from './creators.js';
-export type { <%= h.changeCase.pascal(documentType) %>PHState } from './ph-factories.js';
```
ryanwolhuter, line 10: we should keep our types to the types files that we have set aside for this purpose. this avoids confusion and lets us do export type * instead of export * which prevents outputting empty files in the built code

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-document-model/types.esm.t
```diff
@@ -17,5 +16,16 @@ export type * from './schema/types.js';
 <% if(!hasLocalSchema) { -%>
 <%= 'type ' + h.changeCase.pascal(documentType) %>LocalState = Record<PropertyKey, never>;
 <%} -%>
```
ryanwolhuter, line 18: we can define and export the various types here in the types file, which is what users would usually expect

packages/design-system/src/ui/components/button/button.tsx
```diff
@@ -1,8 +1,8 @@
+import { cn } from "../../../utils/cn.js";
```
ryanwolhuter, line 1: there is something wrong with this component but I can't figure out what it is. `import { cn } from "@powerhousedao/design-system";` works everywhere else, but in this component doing that causes tests to fail with "cn is not a function".

.github/workflows/check-commit.yml
```diff
@@ -8,88 +8,68 @@ on:
   workflow_dispatch:
   push:
     branches: ["main"]
-    paths-ignore:
-      - "packages/document-drive/**"
-      - "apps/academy/**"
   pull_request:
```
ryanwolhuter, line 11: I'll need someone who is fimiliar with this stuff to help me figure out how to add it to my new workflow here

PR #1833, opened 2025-09-11
feat(vetra): add basic support to drop documents on vetra drive

acaldas:
This PR adds basic support to dropping documents on the vetra drive.
It doesn't reject unsupported document types.

https://github.com/user-attachments/assets/6e3e66c2-3e1f-4b3a-8626-658229715935


gpuente (approved):
LGTM

PR #1879, opened 2025-09-25
Document editor update

acaldas:
Re-added generic and `dispatch` method to EditorProps.
Improved boilerplate.

https://github.com/user-attachments/assets/24819492-9eb0-4295-b596-d0cb84a2f6db


apps/connect/src/components/drive-editor-container.tsx
```diff
@@ -40,10 +42,9 @@ export function DriveEditorContainer() {
   );
   const defaultDriveEditor = useDefaultDriveEditorModule();
 
-  const DriveEditorComponent =
-    driveEditor?.Component ??
+  const DriveEditorComponent = (driveEditor?.Component ??
     defaultDriveEditor?.Component ??
-    GenericDriveExplorer.Component;
+    GenericDriveExplorer.Component) as FC<DriveEditorProps>;
```
acaldas, line 47: This type cast is needed because DriveEditorProps is different from EditorProps.
This is because we pass in the Connect methods to show the Document create model, etc.
Once that is changed to use the window then we'll no longer need to have different props for drive editors.

apps/connect/src/components/editors.tsx
```diff
@@ -129,6 +126,10 @@ export const DocumentEditor: React.FC<Props> = (props) => {
     return <EditorLoader message="Loading editor" />;
   }
 
+  if (isLoadingDocument) {
```
acaldas, line 130: This makes it so that the Editor component is only called when the document is loaded.

Since we get the DocumentModule from the document documentType then we know it's a supported document. We should also validate the document when these methods are available: `documentModel.utils.validateDocument(document)` so the editor can assume it's receiving a valid document.

apps/connect/src/components/editors.tsx
```diff
@@ -256,19 +257,7 @@ export const DocumentEditor: React.FC<Props> = (props) => {
                   ),
                 }}
                 document={document}
-                documentNodeName={documentName ?? ""}
```
acaldas: Changing the component props type from `any` to `EditorProps` made it visible that none of these props are being used.

apps/connect/src/store/editor.ts
```diff
@@ -12,7 +14,7 @@ export async function loadGenericDriveExplorerEditorModule(): Promise<VetraEdito
     id: DEFAULT_DRIVE_EDITOR_ID,
     name,
     documentTypes,
-    Component,
+    Component: Component as FC<EditorProps>,
```
acaldas, line 17: Same as above, `VetraEditorModule` doesn't expect `DriveEditorProps`, which we should remove.

packages/builder-tools/connect-studio/server.ts
```diff
@@ -33,6 +33,13 @@ logger.warn = (msg, options) => {
   if (msg.includes("The above dynamic import cannot be analyzed by Vite.")) {
     return;
   }
+  if (
+    msg.includes(
```
acaldas, line 37: Hide the annoying warnings. Let's try to fix this with new Connect build.

packages/builder-tools/document-model-editor/editor.tsx
```diff
@@ -33,13 +34,9 @@ import {
...
 } from "./utils/helpers.js";
-import { pascalCase } from "change-case";
-import { useDocumentById } from "@powerhousedao/reactor-browser";
 
-export function DocumentModelEditor(props: EditorProps) {
-  const { document: initialDocument } = props;
-  const [unsafeDocument, dispatch] = useDocumentById(initialDocument.header.id);
-  const document = unsafeDocument as DocumentModelDocument;
+export function DocumentModelEditor(props: EditorProps<DocumentModelDocument>) {
+  const { document, dispatch } = props;
```
acaldas: Use the provided document and dispatch method instead of `useDocumentById`.
Connect must ensure that they are correct.

packages/builder-tools/editor-utils/storybook.tsx
```diff
@@ -39,38 +39,37 @@ export type EditorStoryArgs = Partial<{
   };
 }>;
 
-export type EditorStoryProps = EditorProps & EditorStoryArgs;
```
acaldas: This storybook is not working. But it already wasn't from my tests.

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-editor/editor.esm.t
```diff
@@ -3,24 +3,77 @@ to: "<%= rootDir %>/<%= h.changeCase.param(name) %>/editor.tsx"
...
 export default function Editor(props: IProps) {
-    return (
+  const { document, dispatch } = props;
+
+  function handleSetName(values: { documentName: string }) {
+    dispatch(baseActions.setName(values.documentName));
+  }
+
+  return (
+    <div className="ph-default-styles text-center">
```
acaldas: We shouldn't include the default stylesheet on every external package, as they will affect each other if the same class name is used (`html-defaults-container`).
We can add the styles under `.ph-default-styles` in the design-system css file.

packages/codegen/src/codegen/index.ts
```diff
@@ -92,25 +92,6 @@ async function getDocumentTypesMap(
       });
   }
 
-  // add documents from powerhouse package if installed
-  try {
-    /* eslint-disable */
-    // @ts-ignore-error TS2307 this import is expected to fail if document-model-libs is not available
-    const documentModels = await import("document-model-libs/document-models");
```
acaldas, line 99: We could do something similar to look for the installed external packages.
If `@powerhousedao/vetra` is added in `powerhouse.config.json` then it would be possible to create an editor for one of it's document models.

packages/document-model/src/document/types.ts
```diff
@@ -358,13 +358,11 @@ export type EditorContext = {
 
 export type ActionErrorCallback = (error: unknown) => void;
 
-export type EditorDispatch = (
-  action: Action,
-  onErrorCallback?: ActionErrorCallback,
```
acaldas: It would be nice to get the error information in case the action input is invalid or there is an error processing it. Not sure what's the best way to do it in React.

packages/document-model/src/document/types.ts
```diff
@@ -437,7 +435,7 @@ export type Manifest = {
 export type DocumentModelLib<TState extends PHBaseState = PHBaseState> = {
   manifest: Manifest;
   documentModels: DocumentModelModule<TState>[];
-  editors: EditorModule[];
+  editors: EditorModule<PHDocument<TState>>[];
```
acaldas: I'd like to make this accept `TState` as a generic to make it consistent with `DocumentModelModule`.
Didn't do that because currently there is no easy way to import the State type of a document model like we do in the editor boilerplate.

packages/reactor-browser/src/types/drive-editor.ts
```diff
@@ -71,8 +71,7 @@ export type DriveEditorConfig = {
   };
 };
 
-export type DriveEditorProps = {
-  document: PHDocument;
+export type DriveEditorProps = EditorProps<DocumentDriveDocument> & {
```
acaldas: We need to kill this

packages/vetra/html-defaults.css
```diff
@@ -1,4 +1,4 @@
-.html-defaults-container h1 {
```
acaldas, line 1: This was making it so an external package couldn't change these default styles.
We should remove it if it's not being used in the Vetra editors.

gpuente (approved):
LGTM 🚀

apps/connect/src/components/drive-editor-container.tsx
```diff
@@ -65,6 +66,7 @@ export function DriveEditorContainer() {
           showDeleteNodeModal,
         }}
         document={selectedDrive}
+        dispatch={dispatch}
```
ryanwolhuter: this might break things, the dispatch function here is tied to the selected drive

ryanwolhuter: the editor can just call the same hook to get the dispatch function. just worried about stale closures

packages/builder-tools/document-model-editor/editor.tsx
ryanwolhuter: doesn't this defeat the purpose of having the hook in the first place?

packages/builder-tools/editor-utils/storybook.tsx
ryanwolhuter: the thing with all of these generics in the storybook stories is they cause all kinds of problems and don't really offer much benefit. We don't really need much type inference in the stories and they actually behave quite differently from the actual app code anyway. if we end up doing a bunch of casts to make these types work, we are actually just asking for things to break at runtime

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-editor/editor.esm.t
ryanwolhuter: i'm a little hesitant about using these components from document engineering in the boilerplate code because things in there are not as tightly controlled since that package isn't in the monorepo. could we not rather just use regular html elements here? these are just examples to get started after all

packages/codegen/src/codegen/__tests__/generate-editor.expected.ts
```diff
@@ -0,0 +1,137 @@
...
+        id: 'test-document-model-editor',
+        disableExternalControls: true,
+        documentToolbarEnabled: true,
+        showSwitchboardLink: true,
+    },
+};
+
+export default module;`;
+
+const EXPECTED_EDITOR_METHOD = `export default function Editor(props: IProps) {
```
ryanwolhuter, line 18: I think we should start moving away from these default exports. what is the benefit of doing this instead of a named export? mixing named and default exports makes the bundlers and build tools more likely to get confused, and if you want to do an alias then you have to do a weird export { default as Something }

packages/reactor-browser/src/types/drive-editor.ts
ryanwolhuter: agreed

PR #1884, opened 2025-09-30
refactor: implement window-based drive context

acaldas:
- Added drive context event handling with dispatch and subscription methods.
- Refactored DriveEditorContainer to synchronize drive context with the window object.
- Removed DriveContextProvider from several components, simplifying context management.
- Updated related components and tests to reflect the new context handling approach.

### DIFF
--- a/apps/connect/src/components/drive-editor-container.tsx
+++ b/apps/connect/src/components/drive-editor-container.tsx
@@ -2,6 +2,7 @@ import { useNodeActions, useShowDeleteNodeModal } from "#hooks";
 import { GenericDriveExplorer } from "@powerhousedao/common";
 import type { DriveEditorProps } from "@powerhousedao/reactor-browser";
 import {
+  dispatchSetDriveContextEvent,
   useDefaultDriveEditorModule,
   useDriveEditorModuleById,
   useSelectedDocument,
@@ -9,7 +10,7 @@ import {
 } from "@powerhousedao/reactor-browser";
 import type { DocumentModelModule } from "document-model";
 import type { FC } from "react";
-import { useCallback } from "react";
+import { useCallback, useEffect } from "react";
 import type { FallbackProps } from "react-error-boundary";
 import { ErrorBoundary } from "react-error-boundary";
 import { DocumentEditorContainer } from "./document-editor-container.js";
@@ -39,6 +40,16 @@ export function DriveEditorContainer() {
     [showModal],
   );
   const showDeleteNodeModal = useShowDeleteNodeModal();
+
+  // Synchronize drive context with Window object
+  useEffect(() => {
+    const context = {
+      ...nodeActions,
+      showCreateDocumentModal,
+      showDeleteNodeModal,
+    };
+    dispatchSetDriveContextEvent(context);
+  }, [nodeActions, showCreateDocumentModal, showDeleteNodeModal]);
 
   const driveEditor = useDriveEditorModuleById(
     selectedDrive?.header.meta?.preferredEditor,
--- a/packages/common/editors/generic-drive-explorer/editor.tsx
+++ b/packages/common/editors/generic-drive-explorer/editor.tsx
@@ -5,7 +5,6 @@ import {
 } from "@powerhousedao/design-system";
 import type { DriveEditorProps } from "@powerhousedao/reactor-browser";
 import {
-  DriveContextProvider,
   getSyncStatusSync,
   makeFolderNodeFromDrive,
   setSelectedNode,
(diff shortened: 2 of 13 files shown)


packages/common/editors/generic-drive-explorer/editor.tsx
```diff
@@ -29,7 +28,7 @@ import { SearchBar } from "./components/search-bar.js";
 export type GenericDriveExplorerEditorProps = DriveEditorProps &
   React.HTMLProps<HTMLDivElement>;
 
-export function BaseEditor(props: GenericDriveExplorerEditorProps) {
+export default function Editor(props: GenericDriveExplorerEditorProps) {
```
acaldas, line 31: Let's change this to a named export: `export function Editor`

apps/connect/src/components/drive-editor-container.tsx
```diff
@@ -40,6 +41,17 @@ export function DriveEditorContainer() {
   );
   const showDeleteNodeModal = useShowDeleteNodeModal();
 
+  // TODO: Replace this with a different approach
```
acaldas: @ryanwolhuter Is there a way to do this internally in the hooks?
`useDriveContext` could listen for the `ph:selectedDriveIdUpdated` event.

I think we would need to migrate the modals to also use the window object, so this can be done later.

packages/reactor-browser/src/hooks/drive-context.tsx
```diff
@@ -24,12 +41,10 @@ export const DriveContextProvider: React.FC<
...
-      "useDriveContext must be used within a DriveContextProvider",
-    );
-  }
-  return context;
+export function useDriveContext(): IDriveContext {
+  const driveContext = useSyncExternalStore(
+    subscribeToDriveContext,
+    () => window.phDriveContext,
+  );
+  return driveContext ?? DEFAULT_DRIVE_CONTEXT;
```
acaldas, line 63: I'd just throw an error if `driveContext` is undefined. The default context will just hide potential errors in the setup.

apps/connect/src/components/drive-editor-container.tsx
ryanwolhuter: There is no reason to do this. addDriveContextEventHandler has already been included in the addPhEventHandlers function, which is called when initializing the app. this means that the value is initialized along with all the others.

really this approach is wrong because it is trying to keep using a big context object with a ton of stuff in it, when really this should just be broken down into some smaller parts.

for example, there is no reason that the "nodeActions" and various modal things should be kept together like this. with this setup as it is now, anything involving the modals will affect the rest of the node actions etc. for no reason.

packages/reactor-browser/src/hooks/drive-context.tsx
```diff
@@ -10,9 +12,24 @@ export const DriveContextProvider: React.FC<
   <DriveContext.Provider value={value}>{children}</DriveContext.Provider>
 );
 
+/**
+ * Default drive context value.
+ * Used when window.phDriveContext is not yet initialized.
+ */
+const DEFAULT_DRIVE_CONTEXT: IDriveContext = {
```
ryanwolhuter, line 33: these "node actions" as they are defined all just do various modal things anyway. this is redundant and should be simplified to just a modal context hook. for the modals, they really only need to have at most a name of a particular modal and the id of an item. the rest can be derived from the other hooks without needing to be passed as props

packages/vetra/editors/vetra-drive-app/DriveExplorer.tsx
```diff
@@ -45,7 +45,7 @@ export const DriveExplorer: React.FC<DriveExplorerProps> = ({
   onAddSubgraph,
   onAddProcessor,
   onAddCodegenProcessor,
-  context,
+  context = {},
```
ryanwolhuter, line 48: there is no reason to still have a context prop if the value can instead be derived from a hook

apps/connect/src/components/drive-editor-container.tsx
ryanwolhuter: @acaldas @gpuente everything in the `useNodeActions` hook is derived from calling other hooks that are available outside of connect anyway. I added this hook as an intermediary step, but it is redundant. there is no reason to put what comes from this hook into another hook or into the window.

ryanwolhuter: we could move the `useNodeActions` hook to reactor-browser if need be but really its all logic that shouldn't even need to be in a hook, it doesn't do anything that needs the react runtime.

with that eliminated, this context object has nothing but `showCreateDocumentModal` and `showDeleteNodeModal`, which could easily be refactored into a modals value in the window.

PR #1894, opened 2025-10-01
feat: update file extension to .phd for document export

gpuente:


packages/reactor-browser/src/actions/document.ts
```diff
@@ -149,14 +149,7 @@ export async function exportFile(document: PHDocument, suggestedName?: string) {
...
   }
-  const documentModelModules = reactor.getDocumentModelModules();
-  const documentModelModule = documentModelModules.find(
-    (module) => module.documentModel.id === document.header.documentType,
-  );
-  const extension = documentModelModule?.documentModel.extension;
-  const name = `${suggestedName || document.header.name || "Untitled"}.${
-    extension ? `${extension}.` : ""
-  }zip`;
+  const name = `${suggestedName || document.header.name || "Untitled"}.phd`;
```
acaldas: Was this a requirement? Are we no longer making use of the extension field in the document model specification?

gpuente: We never actually used the document extension. If you defined something like `.phdm`, what you ended up with was `my-file.phdm.zip`.  
I think I can keep the same behavior and just replace `.zip` with `.phd`.  
I'll push a new commit with this

PR #1896, opened 2025-10-02
Update Vetra according to the new Connect build 

acaldas:
Updated the Vetra package with the new vite config, might have missed some dependencies.

Connect was breaking on startup due to `CreateDocumentModal` calling `useSelectedDrive` which was throwing an error due to there not being a selected drive.
I changed the ModalsContainer to render the modals lazily on demand but the actual issue was `useSelectedDrive` being used in unsafe places.

Added the `useSelectedDriveSafe` hook which returns undefined instead of returning an error.
Not sure what naming to use.


packages/vetra/package.json
```diff
@@ -55,41 +55,43 @@
     "test": "vitest run",
     "test:watch": "vitest",
     "generate": "ph-cli generate",
-    "connect": "ph-cli connect",
+    "connect": "vite",
```
gpuente, line 58: ph-cli connect does not longer work?

gpuente (approved):

packages/vetra/package.json
acaldas: Yes, ph-cli is broken in this PR as we need to change `ph connect` to simply run `vite dev/build` on the project. I think most of the code with Vite workarounds in `builder-tools` will be deleted.

PR #1897, opened 2025-10-03
Better error handling on useSelectedDocumentOfType

acaldas:


packages/vetra/editors/hooks/useVetraDocument.ts
```diff
@@ -92,8 +93,14 @@ export function useVetraPackageDocument(documentId: string | undefined | null) {
...
 
 export function useSelectedVetraPackageDocument() {
+  return useSelectedDocumentOfType<VetraPackageDocument, VetraPackageAction>(
+    VETRA_PACKAGE_DOCUMENT_TYPE,
+  );
+}
+
+export function useSelectedDriveVetraPackage() {
   const [selectedDrive] = useSelectedDrive();
-  const documentId = selectedDrive?.state.global.nodes.find(
```
ryanwolhuter, line 96: isn't selected drive still possibly undefined here?

acaldas: There was an issue from the hooks change on `useSelectedDrive`/`useSelectedDriveDocument`.
They had different behaviors, where the old `useSelectedDrive` returned `DriveDocument | undefined` but now throws an error.

As a fix, I added a `useSelectedDriveSafe` hook with the old behavior.
It was a blocker so I had to merge it yesterday but please take a look as I'm not sure what's the best way to deal with it: https://github.com/powerhouse-inc/powerhouse/pull/1896

ryanwolhuter (approved):
makes sense to me

PR #1900, opened 2025-10-03
Feat/cli connect

acaldas:
Initial work on `ph connect` rework.

Some behaviors are different due to the env variable naming change and default values.

### DIFF
--- a/clis/ph-cli/src/commands/connect.ts
+++ b/clis/ph-cli/src/commands/connect.ts
@@ -8,29 +8,28 @@ import type { ConnectOptions } from "../services/connect.js";
 import type { CommandActionType } from "../types.js";
 import { setCustomHelp } from "../utils.js";
 
-async function startConnect(options: ConnectOptions) {
+async function startConnect(options?: ConnectOptions) {
   const Connect = await import("../services/connect.js");
   const { startConnect } = Connect;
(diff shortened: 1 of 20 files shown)


clis/ph-cli/src/commands/connect.ts
```diff
@@ -8,29 +8,28 @@ import type { ConnectOptions } from "../services/connect.js";
...
   .description("Starts Connect Studio (default)")
-  .option("-p, --port <port>", "Port to run the server on", "3000")
-  .option("-h, --host", "Expose the server to the network")
-  .option("--https", "Enable HTTPS")
-  .option("--open", "Open the browser")
+  // Vite dev config: https://github.com/vitejs/vite/blob/1ef57bc7700375bf4bca0edbf0a9e4517c5dd35b/packages/vite/src/node/cli.ts#L184
+  .option("--port <port>", "Port to run the server on", "3000")
+  .option("--host", "Expose the server to the network")
+  .option("--open", "Open browser on startup")
   .option("--config-file <configFile>", "Path to the powerhouse.config.js file")
```
gpuente, line 27: do we still need this option? maybe we can remove it if is not longer needed

gpuente (approved):

PR #1901, opened 2025-10-03
fix: improve build process and fix schema issues

gpuente:
This PR fixes:

- **Styles issues**: Previously, Connect was exporting the CSS file directly, which caused problems in PH projects. Importing the Connect CSS file resulted in missing styles because when running Connect in a PH project, Tailwind does not scan for styles in `node_modules`. To resolve this, Connect now builds its own `styles.css` file.

- **Switchboard crash**: Fixed crashes caused by incorrect schema types in `IDocument`.

- **Default drives not loading**: Addressed the issue where default drives were not being loaded.

- **PNPM dev in Connect**: Updated `pnpm dev` in Connect to run from `src` files instead of `dist` files.


apps/connect/package.json
```diff
@@ -16,14 +16,16 @@
...
     "assets",
     "public",
-    ".env",
-    "./src/index.css"
+    ".env"
   ],
   "exports": {
-    ".": "./lib/src/index.js",
+    ".": {
+      "development": "./src/index.ts",
```
acaldas, line 23: This export worries me as we have found these `package.json` conventions are not consistently supported.
Did you have Connect built while testing?

packages/document-drive/package.json
```diff
@@ -49,7 +49,8 @@
     "bench": "vitest bench",
     "migrate": "prisma generate && prisma db push",
     "copy-prisma": "copyfiles prisma/schema.prisma dist/ && copyfiles 'src/storage/prisma/client/**/*' dist/",
-    "prisma-generate": "prisma generate --no-hints && npm run copy-prisma"
+    "prisma-generate": "prisma generate --no-hints && npm run copy-prisma",
+    "build": "npm run copy-prisma"
```
acaldas: @ryanwolhuter This should be `build:misc` right? But it might need to run before `tsc` in this case.

Should all packages have a `build` script similar to the `build:all` on the root package.json?

apps/connect/package.json
acaldas: Ah, I guess this was the issue: `PNPM dev in Connect: Updated pnpm dev in Connect to run from src files instead of dist files.`

packages/document-drive/package.json
ryanwolhuter: @acaldas this should actually have been removed. we decided to stop re-generating the prisma stuff the whole time a while ago, so there should be no need to run this command at all

acaldas: Maybe the generated code got lost, as there was an error when running `ph vetra`

apps/connect/package.json
ryanwolhuter: this must be a misunderstanding, vite already uses the main.tsx as the entrypoint. vite outputs to dist, tsc outputs to lib. vite never uses dist as the input, the dist is vites outdir

apps/connect/package.json
```diff
@@ -50,6 +52,7 @@
     "postinstall": "cp ./package.json ./package.copy.json",
     "dev": "vite dev",
     "build:vite": "vite build",
+    "build:css": "tailwindcss -i ./src/index.css -o ./dist/styles.css",
```
ryanwolhuter: Why would we want to use the tailwind cli in connect here? connect is built with vite which already runs the tailwind vite plugin and outputs the styles. this would do the same again and output another file

apps/connect/package.json
gpuente: > this must be a misunderstanding, vite already uses the main.tsx as the entrypoint. vite outputs to dist, tsc outputs to lib. vite never uses dist as the input, the dist is vites outdir

I'm not sure about this. I added it because I noticed that when I changed something in one of the source files, the changes weren’t reflected until I ran the build. That’s why I assumed it was not using the src files (at least when using pnpm dev). Maybe I’m missing something?

packages/document-drive/package.json
gpuente: ok, I'll remove this and commit the generated prisma files instead

apps/connect/package.json
ryanwolhuter: ah i see what you mean now! hmm yeah this is interesting. ultimately the problem is that we are trying to use connect simultaneously as an app and as a library, which is inherently weird and has lots of undefined behaviors.

@acaldas I actually think @gpuente's solution here is quite nice — what do you mean when you say the conventions aren't consistently supported? the problem I'm aware of is vite's issue with the "imports" field, which is separate from the "exports" field here. it seems like this works fine with vite, since I just tested it and it works?

apps/connect/package.json
gpuente: Actually, now that I think about it, the styles we’re using here are related to the **lib**, not the **dist**.  
Since the lib is built with `tsc`, it makes sense to build its styles using the Tailwind CLI and output them in the `lib` folder, while Vite should only handle building the `dist` with its own styles.

It would be odd for Vite to output the lib’s styles when it’s not actually building the lib’s code, or for us to export the lib’s styles from the `dist` folder.  
Both should be isolated and not share styles or code across each other.

What do you think?

ryanwolhuter: @gpuente when the vite tailwind plugin runs, its doing the exact same thing that the cli does. it really just runs the same code under the hood.

when we build the monorepo, we always build both dist and lib. We only need to run tailwind once and produce the style code that ends up in the dist folder. the lib folder doesn't need its own copy of the styles in dist, it just needs to be told where to look for them. so we just want exports to have styles.css pointing to dist/styles.css. in the published package, the files array says to include both lib and dist, so all of the code in lib can reference stuff from dist with respect to the package.json exports.

gpuente: got it! 👍🏻

gpuente:
this PR is not longer needed:
https://github.com/powerhouse-inc/powerhouse/pull/1906

PR #1902, opened 2025-10-06
feat(monorepo): update to react 19

ryanwolhuter:
We are able to update to react 19.2 in all of our packages with minimal updates. i have added comments explaining these. In summary:

`useRef` now always requires an initial argument. Before if you didn't pass one it would default to `undefined`. Passing `undefined` as the initial arg achieves the same result.

`forwardRef` is no longer needed. You can pass a ref as a normal prop to any component.

`JSX` is no longer a global namespace. Instead it can either be imported as a type from `react` or accessed with `React.JSX`

`useId` now uses slightly different pre and postfixes. This doesn't change anything but does mean that I needed to update the snapshots with these new generated ids.

There were some weird type issues that cropped up, which are almost definitely things that were always wrong and are just being exposed now that I updated the type definitions for react. From what I can tell this doesn't affect anything, since these components were already working.


apps/academy/src/pages/_archive-homepage.tsx
```diff
@@ -1,10 +1,10 @@
-import clsx from "clsx";
 import useDocusaurusContext from "@docusaurus/useDocusaurusContext";
-import useBaseUrl from "@docusaurus/useBaseUrl";
-import Layout from "@theme/Layout";
 import HomepageFeatures from "@site/src/components/HomepageFeatures";
 import Heading from "@theme/Heading";
+import Layout from "@theme/Layout";
+import clsx from "clsx";
 import styles from "./index.module.css";
+import type { JSX } from "react";
```
ryanwolhuter, line 7: JSX is no longer a global namespace, functionally it works the same

apps/connect/src/services/hmr.ts
```diff
@@ -75,7 +75,7 @@ export async function handlePackageEvents(
 }
 
 export function useSubscribeToVetraPackages() {
-  const hmrRef = useRef<ViteHotContext>();
+  const hmrRef = useRef<ViteHotContext>(undefined);
```
ryanwolhuter, line 78: `useRef` now requires an initial arg, using `undefined` produces the same result as before

packages/design-system/src/connect/components/modal/__snapshots__/delete-item-modal.test.tsx.snap
```diff
@@ -2,12 +2,12 @@
 
 exports[`Modal Component > should match snapshot 1`] = `
 <div
-  aria-describedby="radix-:r2:"
-  aria-labelledby="radix-:r1:"
+  aria-describedby="radix-_r_2_"
```
ryanwolhuter, line 5: the prefix/postfix for react's `useId` is slightly changed, result is functionally the same

packages/design-system/src/connect/components/node-input/node-input.tsx
```diff
@@ -25,7 +25,7 @@ export function NodeInput(props: Props) {
 
   const ref = useRef<HTMLInputElement>(null);
 
-  useOnClickOutside(ref, handleSubmit);
+  useOnClickOutside(ref as RefObject<HTMLElement>, handleSubmit);
```
ryanwolhuter, line 28: this hook has some weird behavior now where it won't accept a ref that is null or undefined, even though that is what an html element ref is at the beginning by default. added a type cast which seems to produce the same result, but needs investigation

packages/design-system/src/rwa/components/table/base/table-base.tsx
```diff
@@ -22,12 +22,10 @@ import { twJoin, twMerge } from "tailwind-merge";
  * @param hasExpandedRow - Whether the table has an expanded row
  * @param specialFirstRow - Function to render a special first row (like the cash asset for instance), must return a React element
  */
-export const TableBase = forwardRef(function TableBase(
-  props: TableBaseProps,
-  ref: React.ForwardedRef<HTMLDivElement>,
-) {
+export function TableBase(props: TableBaseProps) {
```
ryanwolhuter, line 25: Forwardref is no longer needed, ref can be passed as a normal prop

packages/design-system/src/rwa/components/table/base/table.tsx
```diff
@@ -77,6 +77,7 @@ export function Table(props: TableProps) {
       return (
         <RWATableRow
           key={tableItem.id}
+          /* @ts-expect-error - Ref is not typed */
```
ryanwolhuter, line 80: I honestly don't even remember how this logic worked in the first place

packages/design-system/src/ui/components/value-transformer/value-transformer.tsx
```diff
@@ -141,6 +141,7 @@ export function ValueTransformer({
   }, [transformers, setValue]);
 
   return React.cloneElement(children, {
+    // @ts-expect-error - React.HTMLAttributes is not typed
```
ryanwolhuter, line 144: I think this was always wrong but is now actually showing because I updated the type definitions for react

PR #1907, opened 2025-10-06
Fix/enable ph vetra command

gpuente:


clis/ph-cli/src/services/vetra.ts
```diff
@@ -118,69 +111,22 @@ async function startLocalVetraSwitchboard(
...
-      }
-    });
-
-    child.stderr.on("data", (data: Buffer) => {
-      if (verbose) {
-        process.stderr.write(red(`[Connect]: ${data.toString()}`));
-      }
-    });
+  if (options.disableLocalPackages) {
+    process.env.DISABLE_LOCAL_PACKAGES = "true";
```
gpuente: used this to disable load the local packages in connect

acaldas: Let's use the `PH_` prefix on our env variables. If they also need to be available on the browser for Connect then they need to be `PH_CONNECT`. In this case it's only needed at runtime.

Instead of setting the env variables for Connect here, I would pass it as options to `startConnect`, where we can centralize the logic to go from options to env vars.

packages/builder-tools/connect-studio/server.ts
```diff
@@ -232,11 +232,7 @@ export async function startServer(
         exclude: ["node_modules", join(studioPath, "assets/*.js")],
       }),
       viteConnectDevStudioPlugin(true, studioPath),
-      viteLoadExternalPackages(
-        !disableDynamicLoading,
-        options.packages,
-        studioPath,
-      ),
+      viteLoadExternalPackages(!disableDynamicLoading, phPackages, studioPath),
```
acaldas: This file is no longer used, right?

packages/builder-tools/connect-utils/vite-config.ts
```diff
@@ -93,6 +93,8 @@ export function getConnectBaseViteConfig(options: IConnectOptions) {
   const fileEnv = loadEnv(mode, envDir, "PH_");
   const env = { ...fileEnv, ...process.env } as Partial<ConnectEnv>;
 
+  const disableLocalPackages = process.env.DISABLE_LOCAL_PACKAGES === "true";
```
acaldas: Here we should check variables using `env.PH_...` as it includes both the `.env` file and `process.env`, from line 94.

packages/vetra/connect.tsx
```diff
@@ -1,4 +1,3 @@
-import "@powerhousedao/connect/styles.css";
```
acaldas, line 1: Does it need to be added to the `index.html`?

acaldas (changes requested):

clis/ph-cli/src/services/vetra.ts
gpuente: fixed in [08cb25d](https://github.com/powerhouse-inc/powerhouse/pull/1907/commits/08cb25df21956371cb5e947daf2abffe067cc47d)

packages/builder-tools/connect-studio/server.ts
gpuente: my bad, I committed this file by mistake while I was testing something. removed in [08cb25d](https://github.com/powerhouse-inc/powerhouse/pull/1907/commits/08cb25df21956371cb5e947daf2abffe067cc47d)

packages/builder-tools/connect-utils/vite-config.ts
gpuente: fixed in [08cb25d](https://github.com/powerhouse-inc/powerhouse/pull/1907/commits/08cb25df21956371cb5e947daf2abffe067cc47d)

packages/vetra/connect.tsx
gpuente: it is already added in the index.html, I'm using the same pattern from the boilerplate with the connect.css file

clis/ph-cli/src/services/connect.ts
```diff
@@ -1,8 +1,29 @@
 import type { ConnectStudioOptions } from "@powerhousedao/builder-tools";
 import { startConnectStudio } from "@powerhousedao/builder-tools";
 
-export type ConnectOptions = ConnectStudioOptions["devServerOptions"];
+export type ConnectOptions = ConnectStudioOptions["devServerOptions"] & {
```
acaldas: Let's add these options to the `ConnectStudioOptions` type. So that the user can also set them on `vite.config.ts`

apps/connect/package.copy.json
```diff
@@ -20,10 +20,18 @@
     "./src/index.css"
   ],
   "exports": {
-    ".": "./lib/src/index.js",
-    "./config": "./lib/src/connect.config.js",
+    ".": {
+      "development": "./src/index.ts",
+      "import": "./lib/src/index.js",
+  "types": "./lib/src/index.d.ts"
```
acaldas, line 26: Indentation issue. Also on `package.json` I think

clis/ph-cli/src/services/connect.ts
```diff
@@ -1,8 +1,29 @@
...
 export function startConnect(options?: ConnectOptions) {
-  return startConnectStudio({ devServerOptions: options });
+  const {
+    defaultDrivesUrl,
+    drivesPreserveStrategy,
+    disableLocalPackages,
+    ...devServerOptions
+  } = options || {};
+
+  if (defaultDrivesUrl) {
```
acaldas: So the conversion to env vars would happen inside `startConnectStudio`.

packages/builder-tools/connect-utils/types.ts
```diff
@@ -58,6 +58,7 @@ export type ConnectRuntimeEnv = {
   PH_CONNECT_VERSION_CHECK_INTERVAL: string;
   PH_CONNECT_CLI_VERSION: string;
   PH_CONNECT_FILE_UPLOAD_OPERATIONS_CHUNK_SIZE: string;
+  PH_CONNECT_DISABLE_LOCAL_PACKAGES: string;
```
acaldas: I think this doesn't need to be added to the runtime. I would name it `PH_DISABLE_LOCAL_PACKAGE`.

acaldas (approved):

clis/ph-cli/src/services/connect.ts
gpuente: [13ed3e0](https://github.com/powerhouse-inc/powerhouse/pull/1907/commits/13ed3e015010dc89d0936f8aec9097ba2e30a15b)

apps/connect/package.copy.json
gpuente: [13ed3e0](https://github.com/powerhouse-inc/powerhouse/pull/1907/commits/13ed3e015010dc89d0936f8aec9097ba2e30a15b)

clis/ph-cli/src/services/connect.ts
gpuente: [13ed3e0](https://github.com/powerhouse-inc/powerhouse/pull/1907/commits/13ed3e015010dc89d0936f8aec9097ba2e30a15b)

packages/builder-tools/connect-utils/types.ts
gpuente: [13ed3e0](https://github.com/powerhouse-inc/powerhouse/pull/1907/commits/13ed3e015010dc89d0936f8aec9097ba2e30a15b)

PR #1918, opened 2025-10-07
fix: use documentModelModule when creating document in processDocumentJob

gpuente:


thegoldenmule (approved):
Best fix I've ever seen.

PR #1919, opened 2025-10-07
feat(connect, builder-tools): centralized logic to handle env vars

acaldas:


apps/connect/src/connect.config.ts
```diff
@@ -1,123 +1,76 @@
...
 import { getBasePath } from "@powerhousedao/connect";
 import { logger, setLogLevel } from "document-drive";
 
-const APP_VERSION = import.meta.env.PH_CONNECT_VERSION;
-const WARN_OUTDATED_APP =
-  import.meta.env.PH_CONNECT_WARN_OUTDATED_APP || "false";
-const PH_CONNECT_STUDIO_MODE =
-  import.meta.env.PH_CONNECT_STUDIO_MODE || "false";
+// Load environment variables with validation and defaults
+const env = loadRuntimeEnv({
```
acaldas, line 6: Env vars get validated with zod.

packages/builder-tools/connect-utils/runner.ts
```diff
@@ -64,12 +65,19 @@ export async function startConnectStudio(options?: ConnectStudioOptions) {
   };
 
   if (defaultDrivesUrl) {
-    process.env.PH_CONNECT_DEFAULT_DRIVES_URL = defaultDrivesUrl.join(",");
+    setConnectEnv({
```
acaldas, line 68: Connect env vars should be set with `setConnectEnv` instead of directly on `process.env`

packages/builder-tools/connect-utils/env-config.ts
```diff
@@ -0,0 +1,549 @@
...
+      } catch {
+        // Validation failed, log warning and try next source
+        const valueStr =
+          value === null
+            ? "null"
+            : typeof value === "object"
+              ? JSON.stringify(value)
+              : (value as string);
+        console.warn(
+          `Invalid value for ${key} from ${source.name}: ${valueStr}. Trying next source.`,
```
acaldas, line 452: Invalid env vars will log a warning and it's value ignored.

packages/vetra/index.html
```diff
@@ -19,6 +19,10 @@
   </head>
   <body>
     <div id="root"></div>
-    <script type="module" src="./connect.tsx"></script>
+    <script type="module">
```
ryanwolhuter, line 22: this is very clever, nice one

ryanwolhuter (approved):

packages/vetra/index.html
```diff
@@ -19,6 +19,10 @@
...
+      import { AppLoader } from "@powerhousedao/connect";
+      import { createRoot } from "react-dom/client";
+      createRoot(document.getElementById("root")).render(AppLoader());
```
gpuente, line 25: I'm not 100% sure about this, I think this can break some hooks according to the docs. I think we should use `React.createElement(AppLoader)` instead of `AppLoader()`. this is the recommended way to render a react component without JSX

gpuente (approved):
LGTM, just one comment

packages/vetra/index.html
acaldas: Thanks, this will change to simply `import @powerhousedao/connect/main.js` but I'll keep in mind if I see this issue.

PR #1920, opened 2025-10-07
Import design system tailwind theme separately

acaldas:
This makes the themes more consistent and should avoid those issues where the theme was not applied.

Also removes the styles warning when running Connect.


ryanwolhuter (approved):
yeah should have done this long ago lol

gpuente (approved):
LGTM

PR #1924, opened 2025-10-08
feat(vetra) package read model

froid1911:


packages/vetra/processors/vetra-read-model/factory.ts
```diff
@@ -0,0 +1,37 @@
...
+    const store =
+      await module.relationalDb.createNamespace<VetraReadModelProcessor>(
+        namespace,
+      );
+
+    // Create a filter for the processor
+    const filter: RelationalDbProcessorFilter = {
+      branch: ["main"],
+      documentId: ["*"],
+      documentType: ["*"],
```
acaldas: Should it filter for the package document type?

acaldas (approved):

PR #1931, opened 2025-10-09
feat(reactor-browser,connect,vetra): use hooks for all editor values

ryanwolhuter:
I have introduced here some abstractions to make it easier for us to manage global state in connect and external projects. With these updates, it is now possible to do away with all props for editors and instead rely solely on react hooks.

The existing global variables have been moved into an object that lives on window, `window.ph`. 

`window.ph` has this type:

```ts
export type PHGlobal = PHGlobalConfig & {
  loading?: boolean;
  reactor?: IDocumentDriveServer;
  connectCrypto?: IConnectCrypto;
  did?: DID;
  renown?: IRenown;
  user?: User;
  loginStatus?: LoginStatus;
  vetraPackages?: VetraPackage[];
  processorManager?: ProcessorManager;
  drives?: DocumentDriveDocument[];
  documents?: PHDocument[];
  selectedDriveId?: string;
  selectedNodeId?: string;
  modal?: PHModal;
  selectedTimelineRevision?: string | number | null;
};

  
export type PHGlobalKey = keyof PHGlobal;
export type PHGlobalValue = PHGlobal[PHGlobalKey];

export type PHGlobalConfig = PHCommonGlobalConfig &
  PHDrivesGlobalConfig &
  PHEditorsGlobalConfig &
  PHAnalyticsGlobalConfig &
  PHRenownGlobalConfig &
  PHSentryGlobalConfig &
  PHProcessorsGlobalConfig;

```

I have added a generic helper function which enables us to remove a lot of duplication while also being significantly more consistent. This function takes a `key` which corresponds to one of the keys in the `PHGlobal` object and creates:

a react hook for accessing the value
a setter for setting the value, not a hook and can be used anywhere
a function for registering the event listener on the window.

the function looks like this:

```ts
export function makePHEventFunctions<TKey extends PHGlobalKey>(key: TKey) {
  const setEventName = `ph:set${capitalCase(key)}` as const;
  const updateEventName = `ph:${key}Updated` as const;

  function setValue(value: PHGlobal[TKey] | undefined) {
    const event = new CustomEvent(setEventName, {
      detail: { [key]: value },
    });
    window.dispatchEvent(event);
  }

  function dispatchUpdatedEvent() {
    const event = new CustomEvent(updateEventName);
    window.dispatchEvent(event);
  }

  function handleSetValueEvent(event: SetEvent<TKey>) {
    const value = event.detail[key];
    if (!window.ph) throw new Error("ph global store is not defined");
    window.ph[key] = value;
    dispatchUpdatedEvent();
  }

  function addEventHandler() {
    window.addEventListener(setEventName, handleSetValueEvent as EventListener);
  }

  function subscribeToValue(onStoreChange: () => void) {
    window.addEventListener(updateEventName, onStoreChange);
    return () => {
      window.removeEventListener(updateEventName, onStoreChange);
    };
  }

  function getSnapshot() {
    if (!window.ph) throw new Error("ph global store is not defined");
    return window.ph[key];
  }

  function useValue() {
    return useSyncExternalStore(subscribeToValue, getSnapshot);
  }

  return {
    useValue,
    setValue,
    addEventHandler,
  };
}

```

and produces an example output like this:

```ts
export const {
  useValue: useRouterBasename,
  setValue: setRouterBasename,
  addEventHandler: addRouterBasenameEventHandler,
} = makePHEventFunctions("routerBasename");
```

these functions are all strongly typed which is inferred from the value in the key in PHGlobal.

I have also added this updated logic for registering the needed event handlers in connect:

```ts
const phGlobalEventHandlerRegisterFunctions: Record<PHGlobalKey, () => void> = {
  loading: addLoadingEventHandler,
  reactor: addReactorEventHandler,
  modal: addModalEventHandler,
  ... etc
};
export function addPHEventHandlers() {
  for (const fn of Object.values(phGlobalEventHandlerRegisterFunctions)) {
    fn();
  }
}
```

with this typing, if you add a new field to PHGlobal and forget to add the register function for the handler to the register functions object, typescript will warn you.


packages/reactor-browser/src/hooks/use-on-drop-file.ts
```diff
@@ -1,23 +1,23 @@
...
+type UseOnDropFile = () => (
   file: File,
   onProgress?: FileUploadProgressCallback,
   resolveConflict?: ConflictResolution,
 ) => Promise<FileNode | undefined>;
 
-export const useOnDropFile: UseOnDropFile = (documentTypes = []) => {
+export const useOnDropFile: UseOnDropFile = () => {
   const [selectedDrive] = useSelectedDriveSafe();
+  const supportedDocumentTypes = useSupportedDocumentTypes();
```
acaldas: This is not the same logic. The supported document types are defined for each drive, while this hooks returns all the document models in the reactor.
For example, Vetra only allows dropping vetra documents.
This hook should accept a `documentTypes` prop that defaults to `useSupportedDocumentTypes`.

packages/reactor-browser/src/hooks/user.ts
```diff
@@ -1,14 +1,17 @@
-import {
-  subscribeToLoginStatus,
-  subscribeToUser,
-} from "@powerhousedao/reactor-browser";
-import { useSyncExternalStore } from "react";
 import { useAllowList } from "./config.js";
+import { makePHEventFunctions } from "./make-ph-event-functions.js";
```
acaldas, line 1: IMO, `makePHStateFunctions` makes more sense, as it's kind of a `useState`. But I'm fine with either name.

packages/vetra/index.ts
```diff
@@ -21,7 +20,6 @@ export const documentModels: (
   | DocumentModelModule<ProcessorModulePHState>
   | DocumentModelModule<AppModulePHState>
 )[] = Object.values(documentModelsExports);
-export const editors: (EditorModule | DriveEditorModule)[] =
```
acaldas, line 24: 👍

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-editor/index.esm.t
```diff
@@ -10,8 +10,6 @@ export const module: EditorModule = {
     documentTypes: [<% if(!documentTypes.length){ %>'*'<% } else { %><% documentTypes.forEach(type => { _%>"<%= type %>", %><% }); _%> <% } %>],
     config: {
         id: '<%= editorId || 'editor-id' %>',
-        disableExternalControls: true,
```
acaldas, line 13: Can you add to `RELEASE_NOTES.md` on the new way to configure this?

acaldas (commented):
I think we should go ahead with the `<DocumentToolbar>` component and use it instead of global state for the `disableExternalControls`/`documentToolbarEnabled`/`showSwitchboardLink`

packages/vetra/document-models/app-module/app-module.json
```diff
@@ -13,8 +13,8 @@
             "changeLog": [],
             "state": {
                 "global": {
-                    "schema": "type AppModuleState {\n  name: String!\n  status: StatusType!\n  documentTypes: [DocumentTypeItem!]\n  dragAndDrop: DragAndDropSettings\n}\n\nenum StatusType {\n  DRAFT\n  CONFIRMED\n}\n\ntype DocumentTypeItem {\n  id: OID!\n  documentType: String!\n}\n\ntype DragAndDropSettings {\n  enabled: Boolean!\n}",
```
acaldas, line 16: We could still keep this setting in the Vetra App Module document, the codegen for the drive editor could automatically add the `withDropZone` with the supported `documentTypes`

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-drive-editor/index.esm.t
```diff
@@ -2,20 +2,16 @@
...
-    disableExternalControls: true,
-    documentToolbarEnabled: true,
-    showSwitchboardLink: true,<%_ if (dragAndDropEnabled) { _%>
-    documentTypes: [<%- (dragAndDropDocumentTypes ? JSON.parse(dragAndDropDocumentTypes) : []).map(type => JSON.stringify(type)).join(', ') %>],
-    dragAndDrop: {
-      enabled: true,
-    },<%_ } _%>
+    name: "<%= name || 'YOUR_APP_NAME' %>",
+    <%_ if (dragAndDropEnabled) { _%>
+    documentTypes: [<%- (dragAndDropDocumentTypes ? JSON.parse(dragAndDropDocumentTypes) : []).map(type => JSON.stringify(type)).join(', ') %>],<%_ } _%>
```
acaldas: This should be moved to config.esm.t right?

packages/codegen/src/codegen/hygen.ts
```diff
@@ -326,14 +325,11 @@ export async function hygenGenerateDriveEditor(
...
-    args.push("--drag-and-drop-enabled", "true");
-    if (editorOptions.documentTypes.length > 0) {
-      args.push(
-        "--drag-and-drop-document-types",
-        JSON.stringify(editorOptions.documentTypes),
-      );
-    }
+  if (editorOptions?.documentTypes.length) {
+    args.push(
+      "--drag-and-drop-document-types",
```
acaldas: Let's use `--allowed-document-types` instead

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-drive-editor/editor.esm.t
```diff
@@ -2,7 +2,17 @@
...
+import type { EditorProps } from "document-model";
 import { DriveExplorer } from "./components/DriveExplorer.js";
+import { editorConfig } from "./config.js";
 
-export const Editor = withDropZone(DriveExplorer);
\ No newline at end of file
+export function Editor(props: EditorProps) {
+  useSetPHGlobalEditorConfig(editorConfig);
+  return (
+    <DropZoneWrapper>
```
acaldas, line 14: Will create an issue to move this to Connect so it doesn't need to be added on every drive.
Not a blocker for this PR.

packages/builder-tools/document-model-editor/editor.tsx
```diff
@@ -36,9 +38,9 @@ import {
 } from "./utils/helpers.js";
 
 export function DocumentModelEditor() {
+  useSetPHGlobalEditorConfig(editorConfig);
```
acaldas: Should we just name this `useSetEditorConfig`?
Might make it harder to understand how it works, but we could document the method with an explanation.

packages/reactor-browser/src/hooks/make-ph-event-functions.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export function makePHEventFunctions<TKey extends PHGlobalKey>(key: TKey) {
+  const setEventName = `ph:set${capitalCase(key)}` as const;
+  const updateEventName = `ph:${key}Updated` as const;
+
+  function setValue(value: PHGlobal[TKey] | undefined) {
+    const event = new CustomEvent(setEventName, {
+      detail: { [key]: value },
+    });
+    window.dispatchEvent(event);
```
acaldas, line 17: Just a suggestion, for extra safety and encapsulation we could use our own `EventTarget`:

```ts
window.ph.target = new EventTarget();

// then
window.ph.target.addEventListener(...);
window.ph.target.dispatchEvent(event);
window.ph.target.dispatchEvent(event);
```

packages/builder-tools/document-model-editor/config.ts
```diff
@@ -0,0 +1,7 @@
+import type { PHGlobalEditorConfig } from "@powerhousedao/reactor-browser";
+
+export const editorConfig: PHGlobalEditorConfig = {
+  allowedDocumentTypes: undefined,
+  isExternalControlsEnabled: false,
+  isDragAndDropEnabled: true,
```
acaldas: I'm wondering if we need make `allowedDocumentTypes` and `isDragAndDropEnabled` only available for drive editors?

What happens if a drive editor has drag and drop disabled and a document model editor is opened? It would be possible to drop any document into the drive right?

acaldas: I think we should separate `EditorConfig` from `DriveEditorConfig`. `isExternalControlsEnabled` and the Document toolbar config also don't do anything for drive editors

packages/reactor-browser/src/hooks/config/editor.ts
```diff
@@ -0,0 +1,71 @@
...
+/** Defines the document types a drive supports.
+ *
+ * Defaults to all of the document types registered in the reactor.
+ */
+export function useAllowedDocumentTypes() {
+  const definedAllowedDocumentTypes =
+    allowedDocumentTypesEventFunctions.useValue();
+  return definedAllowedDocumentTypes;
+}
+
```
acaldas, line 58: We should have some kind of reset on selected editor change. So that when a editor is unmounted the config is reseted to the default values, then the new editor might change them

ryanwolhuter: I was thinking we can have an effect hook call in connect that watches the selected editor, and when it changes to be no selected editor then it calls the hook with the default values.

packages/reactor-browser/src/hooks/make-ph-event-functions.ts
ryanwolhuter: oh that's cool i didn't know about this one

packages/builder-tools/document-model-editor/config.ts
ryanwolhuter: @acaldas yeah the whole unified interface situation was really only an issue with editors and their props. now that we are using these hooks there is no reason not to use separate types for the configs

packages/reactor-browser/src/hooks/user.ts
ryanwolhuter: yeah i agree, can do a big rename after this pr

acaldas (approved):

PR #1935, opened 2025-10-09
fix(ph-reactor): remove test files from root index

acaldas:
Fixes error when running `ph reactor` on external project.

image


gpuente (approved):
LGTM 🚀

PR #1936, opened 2025-10-09
feat: generate unique vetra drive IDs per project

gpuente:
## Description:

- generate unique vetra drive ids per project

Screenshot 2025-10-09 at 1 38 45 PM

[Boilerplate update here](https://github.com/powerhouse-inc/document-model-boilerplate/pull/37)

### DIFF
--- a/clis/ph-cli/src/utils.ts
+++ b/clis/ph-cli/src/utils.ts
@@ -1,5 +1,6 @@
 import type { PowerhouseConfig } from "@powerhousedao/config";
 import type { Command } from "commander";
+import crypto from "node:crypto";
 import fs from "node:fs";
 import { homedir } from "node:os";
 import path, { dirname } from "node:path";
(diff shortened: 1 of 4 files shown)


clis/ph-cli/src/utils.ts
```diff
@@ -109,6 +110,22 @@ export function getProjectInfo(debug?: boolean): ProjectInfo {
...
+ * @param name - The name prefix for the drive ID (e.g., "vetra", "powerhouse")
+ * @returns A unique drive ID in the format "{name}-{hash}"
+ */
+export function generateProjectDriveId(name: string): string {
+  const projectInfo = getProjectInfo();
+  const hash = crypto
+    .createHash("sha256")
+    .update(projectInfo.path)
+    .digest("hex");
+  const shortHash = hash.substring(0, 8);
```
acaldas, line 125: Let's use generateId from document-model instead

acaldas (approved):

clis/ph-cli/src/utils.ts
gpuente: I can’t use `generateId` here because the ID I’m generating is based on the project path, it needs to be deterministic. Otherwise, every time we run Vetra, it would generate a different ID, even if a Vetra drive already exists.
We could use `generateId` and add some extra logic to handle this scenario, but I think this approach is much simpler.
What I can do is move this util to `document-model/core`.

Let me know what you think.

acaldas: Ah got it.

PR #1974, opened 2025-10-16
Handle invalid getDocument calls and clearStorage UX tweaks

### LINKED ISSUE #1865: BUG: Sync algo fails to find a number of documents upon adding a remote drive.
**Describe the bug**
Sync algo fails to find a number of documents upon adding a remote drive. The documents are there, they have not been deleted. After refresh it does find the documents without issue. 

See Video

https://github.com/user-attachments/assets/28618f8c-a895-403f-9bbf-5edb862f3135

**To Reproduce**

**Resulting behavior**
A clear and concise description of what happens, the bug itself.

**Expected behavior**

**Logs and/or Screenshots**

**Desktop (please complete the following information):**

**Additional context**

**How will regressions be avoided in the future**

- Describe tests
- Describe solutions

acaldas:
Fixes #1865.

The Document not found error was due to trying to get all documents on every update on a drive.
I was able to reproduce to issue by creating a remote drive, and creating and then deleting a document.

When syncing that drive on a fresh Connect, it would try to load that deleted document, which failed due to that document not existing since the sync protocol doesn't sync deleted documents.

By using `Promise.allSettled` instead of `Promise.all` we still get back the documents that do exist, as this could be causing document to not appear.
For now, I've changed it to log a warning as I'm hoping to get a better fix for this by debouncing `refreshReactorData` calls. 

Also added some tweaks to the Clear Storage flow to avoid flashing the HomeScreen with the drives that were just deleted.

### DIFF
--- a/apps/connect/src/components/modal/modals/ClearStorageModal.tsx
+++ b/apps/connect/src/components/modal/modals/ClearStorageModal.tsx
@@ -7,11 +7,32 @@ import {
   showPHModal,
   usePHModal,
 } from "@powerhousedao/reactor-browser";
+import { childLogger } from "document-drive";
+import { useState } from "react";
 import { useTranslation } from "react-i18next";
+
+const logger = childLogger(["ClearStorage"]);
+
 export function ClearStorageModal() {
   const phModal = usePHModal();
   const open = phModal?.type === "clearStorage";
   const { t } = useTranslation();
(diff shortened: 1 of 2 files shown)


ryanwolhuter (approved):
nice, always fun to see allSettled out in the wild

PR #1980, opened 2025-10-16
fix(vetra): update package.json to include src files

gpuente:
## Description:

This PR fixes the sourcemap warning in Vetra. Since we publish the Vetra package with sourcemaps enabled, we also need to include the source files to ensure they work correctly

Screenshot 2025-10-16 at 1 12 25 PM

### DIFF
--- a/packages/vetra/package.json
+++ b/packages/vetra/package.json
@@ -11,14 +11,7 @@
     "access": "public"
   },
   "files": [
-    "dist",
-    "document-models",
-    "editors",
-    "subgraphs",
-    "processors",
-    "utils",
-    "index.ts",
-    "style.css"
+    "dist"
   ],
   "exports": {
     ".": {
--- a/packages/vetra/package.json
+++ b/packages/vetra/package.json
@@ -11,7 +11,14 @@
     "access": "public"
   },
   "files": [
-    "dist"
+    "dist",
+    "document-models",
+    "editors",
+    "subgraphs",
+    "processors",
+    "utils",
+    "index.ts",
+    "style.css"
   ],
   "exports": {
     ".": {
--- a/packages/vetra/tsconfig.json
+++ b/packages/vetra/tsconfig.json
@@ -13,7 +13,7 @@
     "lib": ["ESNext", "dom", "dom.iterable"],
 
     // Other Outputs
-    "sourceMap": true,
+    "sourceMap": false,
     "declaration": true,
     "declarationMap": true,
 
--- a/packages/vetra/tsconfig.json
+++ b/packages/vetra/tsconfig.json
@@ -1,42 +1,10 @@
 {
   "extends": "../../tsconfig.options.json",
   "compilerOptions": {
-    "composite": true,
-    // File Layout
     "outDir": "./dist",
-
-    // Environment Settings
-    // See also https://aka.ms/tsconfig/module
-    "module": "nodenext",
-    "target": "esnext",
     "types": ["node", "vitest/globals"],
     "lib": ["ESNext", "dom", "dom.iterable"],
-
-    // Other Outputs
-    "sourceMap": false,
-    "declaration": true,
-    "declarationMap": true,
-
-    // Stricter Typechecking Options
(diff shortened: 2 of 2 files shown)


acaldas (approved):

acaldas:
The Vetra package has a different tsconfig since it extends the monorepo root tsconfig, but we should avoid making it a special case against other external packages as much as possible.
So, how should we setup sourcemaps for external packages in `document-model-boilerplate`?

IMO, it's nice to be able to look at the source code of an installed package.
Currently, it takes you to the `.d.ts` file  or to the transpiled `.js` file.
https://github.com/user-attachments/assets/8662c8bc-8bd0-4c68-a5e6-1b34c6094b46

Eventually, the packages could just publish the TS code to https://jsr.io/.

@ryanwolhuter I have feeling you have a strong opinion on this.

ryanwolhuter:
@acaldas @gpuente 

this doesn't make any sense to me unfortunately.

what even is the point of having a dist dir if we are also just going to publish the source code?

really the only reason we run tsc and emit to a dist folder is so that we can publish _optimized javascript_. 

Seeing the .d.ts file or the transpiled .js code is the normal behavior of any library. If you inspect any of our external dependencies this is exactly what you will see, because the only alternative is to publish your whole source code.

This is normal behavior and actually a good thing — generally the only thing you are looking for when inspecting code in your editor from an _external dependency_ is the type definitions, i.e. the interfaces you should use. You don't want to see the implementation details of the code, you just want to know how to use it.

This error is happening because of wrong configs and our weird build chain.

The most likely cause is vite, which is probably trying to build its own source maps and doing it wrong for some hidden reason that you would have to sacrifice your favorite goat to discover. The fact that this problem tends to appear and disappear with the wind makes me suspect vite even more.

If we really need to make source code available for the source maps in our published code, then we should at least not publish everything twice.

There is a config option for this exact situation, `inlineSources`:

https://www.typescriptlang.org/tsconfig/#inlineSources

acaldas:
> @acaldas @gpuente
> 
> this doesn't make any sense to me unfortunately.
> 
> what even is the point of having a dist dir if we are also just going to publish the source code?
> 
> really the only reason we run tsc and emit to a dist folder is so that we can publish _optimized javascript_.
> 
> Seeing the .d.ts file or the transpiled .js code is the normal behavior of any library. If you inspect any of our external dependencies this is exactly what you will see, because the only alternative is to publish your whole source code.
> 
> This is normal behavior and actually a good thing — generally the only thing you are looking for when inspecting code in your editor from an _external dependency_ is the type definitions, i.e. the interfaces you should use. You don't want to see the implementation details of the code, you just want to know how to use it.
> 
> This error is happening because of wrong configs and our weird build chain.
> 
> The most likely cause is vite, which is probably trying to build its own source maps and doing it wrong for some hidden reason that you would have to sacrifice your favorite goat to discover. The fact that this problem tends to appear and disappear with the wind makes me suspect vite even more.
> 
> If we really need to make source code available for the source maps in our published code, then we should at least not publish everything twice.
> 
> There is a config option for this exact situation, `inlineSources`:
> 
> https://www.typescriptlang.org/tsconfig/#inlineSources

Since this only happens with Vetra, I suspect it's due to the `development` export.
https://github.com/powerhouse-inc/powerhouse/blob/106abc72569dcbbe9e8055be3280d1293c778b59/packages/vetra/package.json#L16-L20
 
For `ph connect` to work I had to change this import on Connect to a custom name `source` https://github.com/powerhouse-inc/powerhouse/blob/106abc72569dcbbe9e8055be3280d1293c778b59/apps/connect/package.json#L23-L28
 
And added a custom resolve order so that running `vite dev` on `apps/connect` uses the custom import.
 https://github.com/powerhouse-inc/powerhouse/blob/106abc72569dcbbe9e8055be3280d1293c778b59/apps/connect/vite.config.ts#L13-L18

acaldas (commented):
We should try to keep the Vetra package as similar as possible to the other external packages.
Since they also have `sourcemap: true` this is likely an issue of Vite.

PR #1984, opened 2025-10-17
feat(builder-tools): add HMR support for external packages

gpuente:


acaldas:
It's working better but it still triggers a full reload sometimes. For example, if you create a document editor in the vetra drive.
Will try on my end

gpuente:
> It's working better but it still triggers a full reload sometimes. For example, if you create a document editor in the vetra drive. Will try on my end

I think this behavior is expected. When you create a new editor in Vetra, new dependencies are added, and Vite needs to optimize them, which triggers a browser refresh. If you add a second editor, you’ll notice that the refresh doesn’t happen again.
If you check the logs, you’ll see that what’s causing the browser refresh is the dependency optimization, not actual file changes.

acaldas (approved):

PR #2006, opened 2025-10-20
fix(vetra): improve logging for validation errors in document model g…

gpuente:
## Description:
- Update required fields in document model editor
- Added logs for missing fields in document model codegen

Screenshot 2025-10-20 at 2 00 05 PM
Screenshot 2025-10-20 at 1 59 51 PM


acaldas (approved):
I don't think adding an asterisk to the label is enough but better than nothing I guess.

It matches the component from `document-engineering`: https://storybook.powerhouse.academy/?path=/story/data-entry-text-input--required

PR #2017, opened 2025-10-21
feat: implement DocumentToolbar component

gpuente:
## Tasks
- [x] Migrate `DocumentToolbar` component to `document-engineering`
- [x] Refactor `DocumentToolbar` props
- [x] Update hooks usage in `DocumentToolbar` to use documentId as argument if provided
- [x] Update document model editor to use new `DocumentToolbar`
- [x] Release new version of `document-engineering`


packages/reactor-browser/src/hooks/add-ph-event-handlers.ts
```diff
@@ -81,6 +83,8 @@ const phGlobalEventHandlerRegisterFunctions: PHGlobalEventHandlerAdders = {
   selectedNodeId: addSelectedNodeIdEventHandler,
   vetraPackages: addVetraPackagesEventHandler,
   selectedTimelineRevision: addSelectedTimelineRevisionEventHandler,
```
acaldas, line 85: I guess `selectedTimelineRevision` and `selectedTimelineItem` are the same thing, right @ryanwolhuter?

packages/reactor-browser/src/hooks/index.ts
```diff
@@ -83,6 +93,8 @@ export {
   setSelectedTimelineRevision,
   useSelectedTimelineRevision,
 } from "./timeline-revision.js";
+export { useTimelineItems } from "./use-timeline-items.js";
+export type { UseTimelineItemsResult } from "./use-timeline-items.js";
```
acaldas: We want to avoid mixing type exports with normal exports. Is this export really needed?

packages/design-system/src/connect/components/document-toolbar-v2/document-toolbar-v2.tsx
```diff
@@ -0,0 +1,236 @@
...
+    onShowRevisionHistory,
+    disableRevisionHistory = false,
+    onSwitchboardLinkClick,
+    timelineItems,
+    onTimelineItemClick,
+    initialTimelineVisible = false,
+    timelineButtonVisible = false,
+  } = props;
+
+  const [document] = useSelectedDocument();
```
acaldas: It should also be possible to override this by passing a `document` as a prop.

packages/reactor-browser/src/hooks/add-ph-event-handlers.ts
ryanwolhuter: I don't actually even know, I just saw them both in various places and put them both here.

packages/reactor-browser/src/utils/export-document.ts
```diff
@@ -0,0 +1,17 @@
+import { exportFile, showPHModal } from "@powerhousedao/reactor-browser";
+import type { PHDocument } from "document-model";
+import { validateDocument } from "./validate-document.js";
+
+export const exportDocument = (document?: PHDocument) => {
```
ryanwolhuter, line 6: @acaldas @gpuente 

For stuff like this, I think if we expect to re-use it often enough to justify exporting it then it should be a util or hook in reactor browser

packages/design-system/src/connect/components/document-toolbar-v2/utils/use-document-timeline.ts
```diff
@@ -0,0 +1,13 @@
+import {
+  useSelectedDocument,
+  useTimelineItems,
+} from "@powerhousedao/reactor-browser";
+
+export function useDocumentTimeline() {
```
ryanwolhuter: @acaldas @gpuente 

same for here, if this is only intended to be used by the document toolbar then it might as well just live in that component. otherwise we should export it from reactor browser so that we and others can also use it

packages/design-system/src/connect/components/document-toolbar-v2/utils/validate-document.ts
```diff
@@ -0,0 +1,68 @@
+import type {
```
ryanwolhuter, line 1: @acaldas @gpuente 

do we not already have this function in document model? looks to me like it only uses code from there, so if not I'd say that's where it should live.

we don't want to end up with several conflicting ways of validating documents, and we don't want to force a dependency on design system when we don't need to

packages/reactor-browser/src/hooks/index.ts
ryanwolhuter: In general there is no reason to export a type like this, since it will be inferred already from the hook.

What is this type though? I don't actually see it's definition here in the PR.

In general you shouldn't need to worry about the types of these hooks.

The only exception is when they return values with inferred types that come from outside of reactor-browser. In that case, you will get the "such and such return type cannot be inferred blah blah likely not portable" which is just because the type definition is not being exported as part of this specific package.

I've added this type helpers to use in those cases, where you can just import the problem type and then type the function with this type

https://github.com/powerhouse-inc/powerhouse/blob/1b76136e05ac28a00c4ccca7648ee97c3a1f5447/packages/reactor-browser/src/types/global.ts#L33

ryanwolhuter: @acaldas @gpuente 

also the reason you see exports of the destructured result of the makePHEventFunctions in some places and then exports of the things with blahBlah.setValue blahBlah.useValue etc is just because you can't add doc comments ( the type /** */ that shows on hover) for destructured values. So for functions that need to be documented I do the individual exports, and then the lazy way for the rest.

ryanwolhuter (approved):
looks good, a few comments about where things should live but other than that looks solid

apps/connect/src/components/editors.tsx
```diff
@@ -187,32 +182,6 @@ export const DocumentEditor: React.FC<Props> = (props) => {
...
-        <DocumentToolbar
-          onClose={onClose}
-          onExport={onExport}
-          onShowRevisionHistory={() => setRevisionHistoryVisible(true)}
-          title={documentName}
-          timelineItems={timelineItems.data}
-          onTimelineItemClick={setSelectedTimelineItem}
-        />
-      )}
-      {!!isExternalControlsEnabled && (
```
gpuente: I wasn't sure if we wanted to remove this completely or not, lmk if this needs to stay so I can undo this change

gpuente:
Hey @acaldas @ryanwolhuter,
I just finished this PR, addressed the previous comments and added some annotations.
Let me know if you have any feedback or questions!

gpuente:
Btw, I wasn’t able to move the DocumentToolbar component to document-engineering because it creates a circular dependency with the design system.
To migrate this component, we’d need to move not only the DocumentToolbar but also several related components, which I believe is beyond the scope of this ticket.

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
```diff
@@ -1,56 +1,59 @@
...
-  redo?: () => void;
-  canUndo?: boolean;
-  canRedo?: boolean;
-  onExport?: () => void;
-  onClose: () => void;
-  onShowRevisionHistory?: () => void;
-  timelineItems?: Array<TimelineBarItem | TimelineDividerItem>;
+  disableRevisionHistory?: boolean;
   onSwitchboardLinkClick?: () => void;
   initialTimelineVisible?: boolean;
```
acaldas, line 53: Let's change the default value to `true` to we don't need to set it on the template.

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
```diff
@@ -1,56 +1,59 @@
...
   timelineButtonVisible?: boolean;
-  onTimelineItemClick?: DocumentTimelineProps["onItemClick"];
+  onClose?: () => void;
 };
 
 export const DocumentToolbar: React.FC<DocumentToolbarProps> = (props) => {
```
acaldas, line 89: Please add some comments here and on each prop so the dev can read about the component when hovering it.

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
```diff
@@ -1,56 +1,59 @@
...
 import { twMerge } from "tailwind-merge";
-import type {
-  DocumentTimelineProps,
-  TimelineBarItem,
-  TimelineDividerItem,
-} from "../document-timeline/document-timeline.js";
 import { DocumentTimeline } from "../document-timeline/document-timeline.js";
+import { useDocumentUndoRedo } from "./utils/use-document-undo-redo.js";
 
 export type DocumentToolbarProps = {
```
acaldas: For these components meant for external usage let's use the pattern of extending the props with `ComponentPropsWithoutRef<"div">` to allow controlling the root div, unless you guys prefer another pattern.

We should also display `{children}` somewhere on the component.

acaldas: Added a message on Discord to discuss these patterns:
https://discord.com/channels/815917281728659516/1428321344868585502/1430858687802511391

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
acaldas: Maybe `defaultTimelineVisible` to match the controlled/uncontrolled input naming?

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
```diff
@@ -1,56 +1,59 @@
...
+    disableRevisionHistory = false,
+    initialTimelineVisible = false,
   } = props;
 
+  const [selectedDocument] = useSelectedDocument();
+  const document = _document ?? selectedDocument;
+
+  const documentName = document?.header.name || undefined;
+  const parentFolder = useNodeParentFolderById(document?.header.id);
+  const handleClose = onClose ?? (() => setSelectedNode(parentFolder));
```
acaldas, line 109: If a `document` is provided I think we should not use the default navigation method of `setSelectedNode`. As it might create unexpected effects and it's easy to miss for the dev.

So `handleClose` should be a required prop if a `document` is provided. Example type:
```tsx
type DocumentToolbarBaseProps = ComponentPropsWithoutRef<"div"> & {
  disableRevisionHistory?: boolean;
  onSwitchboardLinkClick?: () => void;
  initialTimelineVisible?: boolean;
  timelineButtonVisible?: boolean;
};

export type DocumentToolbarProps =
  DocumentToolbarBaseProps &
  (
    | { document: PHDocument; onClose: () => void } 
    | { document?: undefined; onClose?: () => void } 
  );
```

Please add a comment on the `handleClose` prop explaining that by default it will select the `parentFolder` of the document.

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
```diff
@@ -1,56 +1,59 @@
...
+  const handleExport = exportDocument;
```
acaldas: Let's also provide an optional `onExport` prop.

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
```diff
@@ -1,56 +1,59 @@
...
+
+  const documentUndoRedo = useDocumentUndoRedo(document?.header.id);
+  const isUndoDisabled = !documentUndoRedo.canUndo;
```
acaldas, line 120: Before we had the `externalControlsEnabled` config to control this, we should either provide a `hideControls` prop or possibly a `controls` prop that defaults to `["undo", "redo", "export"]`.

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
acaldas: Let's move these UI elements into it's own components: `DocumentToolbar.UndoButton`, etc

acaldas, line 134: I think we should either not listen to changes or always allow it override the internal state.
```tsx
useEffect(() => {
   setShowTimeline(initialTimelineVisible);
}, [initialTimelineVisible];
```

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-editor/editor.esm.t
```diff
@@ -26,67 +27,70 @@ export function Editor() {
...
-                  <FormLabel htmlFor="name">
-                    <b className="mb-1">Change document name</b>
-                  </FormLabel>
-                  <StringField
-                    name="name"
-                    placeholder="Enter document name"
-                    className="mt-1"
-                  />
+    <div>
+      <DocumentToolbar timelineButtonVisible />
```
acaldas: As said above let's change the default value of `timelineButtonVisible` to avoid needing to add it.

apps/connect/src/components/editors.tsx
```diff
@@ -187,32 +182,6 @@ export const DocumentEditor: React.FC<Props> = (props) => {
...
-              Redo
-            </PowerhouseButton>
-          </div>
-          <div className="flex gap-4">
-            <PowerhouseButton onClick={onClose}>Close</PowerhouseButton>
-          </div>
-        </div>
-      )}
       {revisionHistoryVisible ? (
         <RevisionHistory
```
acaldas, line 186: We'll probably want to make this a component too in the future.

packages/design-system/src/connect/components/document-toolbar/utils/use-document-undo-redo.ts
```diff
@@ -0,0 +1,27 @@
+import { useDocumentById } from "@powerhousedao/reactor-browser";
+import { redo, undo } from "document-model/core";
+
+export function useDocumentUndoRedo(documentId?: string) {
+  const [document, dispatch] = useDocumentById(documentId);
+
+  const globalRevisionNumber = document?.header.revision.global ?? 0;
+  const localRevisionNumber = document?.header.revision.local ?? 0;
+
+  // TODO: check why undo and redo actions are not working
```
acaldas, line 10: I've also noticed this, created a ticket: https://github.com/powerhouse-inc/powerhouse/issues/2027

packages/design-system/src/connect/components/document-toolbar/utils/use-document-undo-redo.ts
```diff
@@ -0,0 +1,27 @@
+import { useDocumentById } from "@powerhousedao/reactor-browser";
+import { redo, undo } from "document-model/core";
+
+export function useDocumentUndoRedo(documentId?: string) {
+  const [document, dispatch] = useDocumentById(documentId);
+
+  const globalRevisionNumber = document?.header.revision.global ?? 0;
```
acaldas, line 7: I'm not sure how this will work out in the future as there might not even be a `global` scope. Let's keep it for now but we should add a test to detect regressions.

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
gpuente: I've added an `enabledControls` prop with this config [484db9b](https://github.com/powerhouse-inc/powerhouse/pull/2017/commits/484db9b02383f77c56d68b80ff72597927525235)

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
gpuente: I haven't pushed this change since we are going to use different components in the new templates

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-editor/editor.esm.t
gpuente: added fix in [484db9b](https://github.com/powerhouse-inc/powerhouse/pull/2017/commits/484db9b02383f77c56d68b80ff72597927525235)

acaldas (approved):

PR #2048, opened 2025-10-27
feat: re-enable switchboard link on document toolbar

gpuente:
## Description

- re-enable switchboard link on document toolbar


packages/reactor-browser/src/hooks/use-get-switchboard-link.ts
```diff
@@ -0,0 +1,48 @@
...
+ * The returned function generates a fresh bearer token and builds the switchboard URL
+ * with authentication when called.
+ *
+ * @param document - The document to create a switchboard URL generator for
+ * @returns An async function that returns the switchboard URL, or null if not applicable
+ */
+export function useGetSwitchboardLink(
+  document: PHDocument | undefined,
+): (() => Promise<string>) | null {
+  const drive = useDriveByDocumentId(document?.header.id);
```
acaldas: We should call useSelectedDrive instead

acaldas (changes requested):

packages/reactor-browser/src/hooks/use-get-switchboard-link.ts
gpuente: fixed in [4ca0f2a](https://github.com/powerhouse-inc/powerhouse/pull/2048/commits/4ca0f2a4106bf72f68c3af2c4c6106a313f8cb8a)

PR #2059, opened 2025-10-31
Support custom boilerplate branch on `ph init`

acaldas:
Also disables prompt to define custom directories.

### DIFF
--- a/clis/ph-cmd/src/commands/init.ts
+++ b/clis/ph-cmd/src/commands/init.ts
@@ -36,10 +36,11 @@ export const init: CommandActionType<
       name: options.project ?? projectName,
       interactive: options.interactive ?? false,
       version: parseVersion({
-        version: options.branch ?? options.tag,
+        version: options.tag,
         dev: options.dev,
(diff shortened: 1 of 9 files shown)


clis/ph-cmd/src/commands/init.ts
```diff
@@ -36,10 +36,11 @@ export const init: CommandActionType<
       name: options.project ?? projectName,
       interactive: options.interactive ?? false,
       version: parseVersion({
-        version: options.branch ?? options.tag,
+        version: options.tag,
```
ryanwolhuter: could we rename this function and its args too? using version is going to confuse people later for sure

ryanwolhuter (approved):

clis/ph-cmd/src/commands/init.ts
acaldas: I think the main confusion comes from not being able to use the `--version` arguments, as it is reserved to show the cli version.
I've updated it to --tag everywhere.

PR #2064, opened 2025-11-03
feat: pre-load local packages when building driveServer

gpuente:


apps/switchboard/src/server.ts
```diff
@@ -133,10 +136,20 @@ async function initServer(serverPort: number, options: StartServerOptions) {
       ? ".ph/read-storage"
       : dbPath;
 
-  const driveServer = new ReactorBuilder([
-    documentModelDocumentModelModule,
-    driveDocumentModelModule,
-  ] as unknown as DocumentModelModule[])
+  const packageManager = new PackageManager([new ImportPackageLoader()], {
```
acaldas: The package manager is initialized on `reactor-api`, so we need to change how Switchboard and `reactor-api` work together to only init the reactor after loading the packages.


https://github.com/powerhouse-inc/powerhouse/blob/75cfba9b55bfe6f9bbd5a3d006e7036790accfe9/packages/reactor-api/src/server.ts#L339-L352

acaldas (changes requested):
Reuse `reactor-api`'s package manager instead of creating new one on Switchboard.

apps/switchboard/src/server.ts
gpuente: I see... what about something like this:

390243522beead5fc07af2caf83ed84aeb6fa59a

let me know what do you think, this is more an exploration commit, but with this change we don't break current implementation of `startAPI`, and it allows pass `documentModels` from `packageManager` to the`initializeDriveServer` function

apps/switchboard/src/server.ts
```diff
@@ -133,44 +135,47 @@ async function initServer(serverPort: number, options: StartServerOptions) {
...
-  ] as unknown as DocumentModelModule[])
-    .withStorage(storage)
-    .withCache(cache)
-    .withOptions({
-      featureFlags: {
-        enableDualActionCreate:
-          options.reactorOptions?.enableDualActionCreate ?? false,
-      },
-    })
-    .build();
```
gpuente, line 148: @acaldas this is the main problem, we need to build the driveServer before passing it to startAPI, but at this point we are not loading local packages, so if the storage contains a document that is not a documentModel or a drive, it will not be able to handle it

acaldas (approved):

PR #2073, opened 2025-11-04
Fix Connect Studio hmr

acaldas:


apps/connect/src/store/external-packages.ts
```diff
@@ -1,14 +1,47 @@
...
+type ExternalPackagesModule = {
+  loadExternalPackages: () => Promise<DocumentModelLib[]>;
+};
 
 const { externalPackagesEnabled } = connectConfig;
 
+const callbacks: ((packages: VetraPackage[]) => unknown)[] = [];
+
+if (import.meta.hot) {
+  import.meta.hot.accept("virtual:ph:external-packages", (data) => {
```
acaldas, line 17: Set the HMR boundary here to propagate the updated external packages across the app

packages/builder-tools/connect-utils/vite-plugins/ph-external-packages.ts
```diff
@@ -1,119 +1,147 @@
...
+          (pkg, index) =>
+            `try {
+          const module = await import('${pkg}');
+          await import('${pkg}/style.css');
+          modules.push({
+              id: 'module${index}',
+              ...module,
+          });
+      } catch (error) {
+          console.error("Error loading package", pkg, error);
```
acaldas, line 22: Load the external packages separately to prevent an invalid package from causing all other packages to not be loaded.

packages/builder-tools/connect-utils/vite-plugins/ph-external-packages.ts
```diff
@@ -1,119 +1,147 @@
...
+    }
+    return allModules;
+  }
 
   const plugin: PluginOption = {
     name: "ph-external-packages",
     enforce: "pre",
     resolveId(id) {
       if (id === virtualModuleId) {
-        return resolvedVirtualModuleId;
```
acaldas, line 56: I'm not sure why Vite has this example in their docs but the extra name just complicated things.

packages/builder-tools/connect-utils/vite-plugins/ph-external-packages.ts
```diff
@@ -1,119 +1,147 @@
...
-          );
+      // if the changed file is not part of any local package, do not override HMR
+      if (!localPackage) {
+        return undefined;
+      }
 
-          // For React components (.tsx, .jsx), let Fast Refresh handle them
-          const isReactComponent =
-            file.endsWith(".tsx") || file.endsWith(".jsx");
+      // iterate through all local package imports to build set of imported modules, if not already built
```
acaldas, line 111: If it's not built yet, builds a set of all the imported modules, starting from the `index.ts` file of the package.
This allows us to ignore stuff that is in the project but is not imported by Connect, like the subgraphs.

packages/builder-tools/connect-utils/vite-plugins/ph-external-packages.ts
```diff
@@ -1,119 +1,147 @@
...
+      for (const module of modules) {
+        if (
+          module.file &&
+          !importedModules.has(module.file) &&
+          !shouldSkipModule(module, localPackage, importedModules) &&
+          module.importers
+            .values()
+            .some((m) => m.file && importedModules.has(m.file))
+        ) {
+          buildImportedModulesSet(module, localPackage, importedModules);
```
acaldas, line 136: If a new module is imported then adds it to the set.

packages/vetra/processors/codegen/document-handlers/document-codegen-manager.ts
```diff
@@ -5,7 +5,7 @@ import { logger } from "../logger.js";
 import type { BaseDocumentGen } from "./base-document-gen.js";
 import type { Config } from "./types.js";
 
-const DEFAULT_DEBOUNCE_TIME = 1000; // 1 second
+const DEFAULT_DEBOUNCE_TIME = 3000; // 3 seconds
```
acaldas, line 8: I think the extra 2 seconds delay is worth it over having outdated (and potentially app breaking) codegen.
Ideally on app startup the codegen processors would ignore all operations and just run on the final state of the document.

PR #2077, opened 2025-11-05
feat: implement vetra document backup functionality in generators

gpuente:


packages/document-drive/src/server/listener/transmitter/types.ts
```diff
@@ -40,6 +40,7 @@ export type InternalOperationUpdate<TState extends PHBaseState = PHBaseState> =
   };
 
 export type InternalTransmitterUpdate = {
+  document: PHDocument;
```
acaldas, line 43: This breaks the tests on document-drive

gpuente: fixed in [668cb9d](https://github.com/powerhouse-inc/powerhouse/pull/2077/commits/668cb9dd0d04304328e04c16370b0baf25aa575d)

PR #2079, opened 2025-11-05
feat(connect): improve bundling

acaldas:
### Removes circular imports on Connect
There are still a lot of them on `design-system` and `document-drive` so there is no significant improvement in the bundling.
Connect circular imports can be tested by running `npx dpdm -T src/main.tsx --tree=false --warning=false --exclude=".*\/packages|node_modules/\.*"` on `apps/connect`.

### Isolate analytics/pglite code
Commented out the pglite code on Connect an isolated it on `common` and `reactor-browser`.
For now this makes the bundle significantly smaller and avoids all the build warnings related to bundling `knex` and `pglite`.

### Small fixes
- Fixed retrieval of the Connect package.json to set the app version and list the dependencies on the Settings modal


apps/connect/src/components/app-loader.tsx
```diff
@@ -1,14 +1,11 @@
...
-  useSubscribeToVetraPackages,
-  useSetSentryUser,
-  createReactor,
-} from "@powerhousedao/connect";
-import "../i18n/index.js";
+import { App } from "@powerhousedao/connect/components/app";
+import { AppSkeleton } from "@powerhousedao/connect/components/app-skeleton";
+import { CookieBanner } from "@powerhousedao/connect/components/cookie-banner";
+import { ModalsContainer } from "@powerhousedao/connect/components/modal/modals-container";
+import { useCheckLatestVersion } from "@powerhousedao/connect/hooks";
```
acaldas, line 5: When importing from the same "barrel" file, explicit imports must be used. In this case `app-loader` is a component that is importing other components.

`useCheckLatestVersion` can be imported from the hooks barrel file since it won't cause a circular import.
I'm still note sure `vite` is able to tree shake barrel files, so these might have to change to use the full path too. To be tested when all circular imports are fixed.

apps/connect/src/context/processor-manager.tsx
```diff
@@ -1,235 +1,241 @@
...
-    <AnalyticsProvider
-      options={{
-        databaseName: connectConfig.analytics.databaseName,
-        pgLiteFactory,
-      }}
-    >
-      {content}
-    </AnalyticsProvider>
-  );
+  return <>{children}</>;
```
acaldas, line 240: I've commented all the pglite stuff out in Connect until we get it fixed. This makes it noticeable when PgLite is added to the bundle because of an external dependency so we can isolate it.

apps/connect/src/hooks/useCheckLatestVersion.ts
```diff
@@ -25,7 +23,7 @@ export const useCheckLatestVersion = () => {
         `Connect is outdated: \nCurrent: ${result.currentVersion}\nLatest: ${result.latestVersion}`,
       );
     } else {
-      toast(<ReloadConnectToast />, {
+      toast(createElement(ReloadConnectToast), {
```
acaldas, line 26: Changed this to make the files under a "barrel" folder have consistent file extensions. The import works without this change anyway.

packages/common/package.json
```diff
@@ -18,9 +18,13 @@
     "./utils": {
       "import": "./dist/utils/index.js",
       "types": "./dist/utils/index.d.ts"
+    },
+    "./analytics": {
```
acaldas, line 22: Separated analytics stuff to it's own export

packages/design-system/src/connect/components/document-toolbar/document-toolbar.tsx
```diff
@@ -15,6 +14,12 @@ import { twMerge } from "tailwind-merge";
 import { DocumentTimeline } from "../document-timeline/document-timeline.js";
 import { useDocumentUndoRedo } from "./utils/use-document-undo-redo.js";
 
+// TODO: Remove this when timeline analytics is available
+// import { useDocumentTimeline } from "@powerhousedao/reactor-browser/analytics";
```
acaldas, line 18: Importing `useDocumentTimeline` causes pglite to be added to the bundle. Commenting out for a later fix, as the feature itself is not functioning anyway

packages/design-system/src/connect/components/modal/settings-modal/dependency-versions/dependency-versions.tsx
```diff
@@ -2,16 +2,11 @@ import { useState } from "react";
 import { object, record, string } from "zod";
 import { Disclosure } from "../../../disclosure/index.js";
 
-const REQUIRED_DEPENDENCIES = [
-  "@powerhousedao/design-system",
+const PH_DEPENDENCIES = [
```
acaldas, line 5: Show all PH dependencies

apps/connect/package.copy.json
```diff
@@ -3,7 +3,7 @@
   "productName": "Powerhouse-Connect",
   "version": "4.1.0-dev.100",
   "description": "Powerhouse Connect",
-  "main": "./dist/index.html",
+  "main": "dist/index.html",
```
ryanwolhuter, line 6: I've never really understood what the deal is with the paths here. what is the difference with vs without the ./ ? Not a problem, just curious

ryanwolhuter (approved):
nice

PR #2080, opened 2025-11-05
test: vetra e2e

gpuente:


.github/workflows/vetra-e2e-tests.yml
```diff
@@ -0,0 +1,81 @@
+name: Vetra E2E Tests
```
acaldas: Let's run these tests on the e2e action that already exists instead

test/vetra-e2e/tests/helpers/folder.ts
```diff
@@ -0,0 +1,126 @@
...
+import { expect } from "@playwright/test";
+import { navigateIntoFolder } from "./navigation.js";
+
+/**
+ * Helper function to create a new folder in the current drive
+ * @param page Playwright Page object
+ * @param folderName Name of the folder to create
+ * @param parentFolderName Optional parent folder to navigate to first
+ *
+ * NOTE: Adapted from connect-e2e. Verify selectors match Vetra UI.
```
acaldas: If possible it would be nice to have a reusable set of helpers across both projects.

acaldas (commented):
Would be nice to separate the helpers into their own library and reuse them in both e2e projects. This would also make fixes easier when the helpers break.

Note: I had to do some fixes to `connect-e2e` after the boilerplate changes merged yesterday, might be needed here too:
https://github.com/powerhouse-inc/powerhouse/commit/d3c3d2ac8a38be4d81e541036af043a51cd31972
https://github.com/powerhouse-inc/powerhouse/commit/e96e16c65b290f90bc2ccdb9993c59f6bf880928

PR #2088, opened 2025-11-10
feat(monorepo): fix circular imports

ryanwolhuter:
NOTE: BREAKING CHANGE IN EXTERNAL PROJECTS using drive editor boilerplate

`FileItem`, `FolderItem`, and `Breadcrumbs` are now imported from `@powerhousedao/design-system/connect`, previously was just `@powerhousedao/design-system`

I have managed to fix all of the circular imports in the monorepo, without making too many updates.

A key thing to note that caused us to miss a lot of these the first time is that we must also check the dist dirs for these circular imports. It can happen that an import does not form a cycle from the perspective of typescript files, but then once they have been transpiled and used as dependencies via dist, they do actually form cycles.

Another important point is that `import type` is a special case since these are removed from transpiled code. This means that we can export all of our types in a package from the root and import them from the root as well, which is quite convenient since types are more often used throughout a package than actual code.

I have added a new script at the monorepo root package.json file called `check-circular-imports` which analyzes the entire repo and takes the above into account. I have added this to our `check-commit` github workflow.

My steps to fix this situation were:

1. remove all imports from @powerhousedao/ and index.js paths
2. remove all index.ts files
3. use custom ts morph scripts to re-add missing imports
4. confirm no cycles
5. work backwards to have the minimal changes while also keeping relatively neat import paths

In this process I was keeping an eye on vite's build of connect, and I posted some results last week. I added some manual chunking that specifically lets us see which deps are going where.


apps/connect/src/components/drive-editor-container.tsx
```diff
@@ -1,5 +1,5 @@
-import { GenericDriveExplorer } from "@powerhousedao/common";
-import { DropZoneWrapper } from "@powerhousedao/design-system";
+import { GenericDriveExplorer } from "@powerhousedao/common/generic-drive-explorer";
```
ryanwolhuter, line 1: this causes us to get a warning from vite saying that this component is dynanmically exported by the module.js file, but statically imported here. this means that it will not be put into a separate chunk.

apps/connect/src/components/modal/index.ts
```diff
@@ -1,2 +1,18 @@
 export * from "./modals-container.js";
```
ryanwolhuter, line 1: we should avoid using many nested folders with index files wherever possible — it makes it much easier to accidentally create cycles and it makes our file paths much more complicated. Ultimately the text found in all of these index files is the same amount of code as it would have been if we put it in one index file.

apps/connect/src/feature-flags.ts
ryanwolhuter, line 1: we should keep source code that we intend to bundle inside the src dir

apps/connect/vite.config.ts
```diff
@@ -17,6 +17,29 @@ export default defineConfig(({ mode }) => {
       conditions: ["source", "browser", "module", "jsnext:main", "jsnext"],
     },
     plugins: [process.env.PH_DEBUG_BUILD === "true" ? analyzer() : undefined],
+    build: {
```
ryanwolhuter, line 20: doing this makes things slightly better in terms of bundle size, but the major improvement is being able to see which chunks are coming from where even when minified

packages/design-system/scripts/create-icon-components.ts
```diff
@@ -66,7 +66,7 @@ readdir(iconsDir, (err, files) => {
           return `${camelCase(attrName)}="${attrValue}"`;
```
ryanwolhuter, line 66: afaik we dont use this script anymore, can we delete?

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-drive-editor/components/Files.esm.t
```diff
@@ -2,7 +2,7 @@
 to: "<%= rootDir %>/<%= h.changeCase.param(name) %>/components/Files.tsx"
 unless_exists: true
 ---
-import { FileItem } from "@powerhousedao/design-system";
+import { FileItem } from "@powerhousedao/design-system/connect";
```
ryanwolhuter, line 5: these changed import paths are the only thing that will affect existing projects. I wanted to move these files to avoid this change, but doing so would end up resulting in even more changes

PR #2093, opened 2025-11-11
feat(connect,common,builder-tools): optimize connect bundle chunks

ryanwolhuter:
Now that we can see what's going on in our modules when building connect, I was able to track down and code-split the things that were causing the last few big chunks in our bundle.

The biggest thing was just to separate the contents of a file utils.ts in the document-model-editor which contained way too much stuff. Behavior is exactly the same but can now be split nicely.

Also added a bit of lazy loading inside that editor component itself.

Finally, I saw warnings in vite saying that we were both dynamically and statically importing the drive and document model editor components. This happens because of how we use them as defaults in connect. This means that the lazy loading was not working in that case.  I updated this to just do static imports instead, which actually makes it produce a significantly smaller bundle.


apps/connect/src/store/editor.ts
```diff
@@ -1,13 +1,14 @@
-import { documentModelEditorModule } from "@powerhousedao/builder-tools/editor";
 import "@powerhousedao/builder-tools/style.css";
-import { GenericDriveExplorer } from "@powerhousedao/common/generic-drive-explorer";
+import DriveExplorerEditor from "@powerhousedao/common/generic-drive-explorer/editor-component";
```
ryanwolhuter, line 2: i tested various things and found that for this special case of default editors in connect, this is by far the most performant

acaldas (approved):

PR #2094, opened 2025-11-11
fix(reactor-browser): handle node and drive navigation on window popstate event

ryanwolhuter:
There were actually two bugs here:

1. We were calling `window.history.pushState` every time the user called `setSelectedDrive` or `setSelectedNode`, without checking if the pathname was already the same. This meant that multiple clicks on the same item produced multiple duplicate entries in the history stack.

2. The `window.popState` event is a special case that requires its own event handlers. With the bug above fixed, we can simply add listeners to this event which update the selected node and / or drive on this event based on the url.


acaldas (approved):
I think we will eventually move this logic to the host app.

ryanwolhuter:
> I think we will eventually move this logic to the host app.

yeah this is really still just an awkward workaround because we need to update to the new react-router

PR #2099, opened 2025-11-12
feat(ph-cli): ph migrate command

ryanwolhuter:
Add a `migrate` command to the ph-cli which makes the various updates to bring an existing project in sync with the current result of generating code and using the boilerplate.

I have intentionally not attempted to fix every possible problem, just ones that are a direct result of legacy codegen. Things like imports from the design system which now use different paths should rather be handled by individual devs.

In the process of doing this I also found some bugs in the codegen. They are relatively minor and easily fixed so I just did them here too.


clis/ph-cli/src/services/generate.ts
```diff
@@ -84,16 +92,20 @@ export async function startGenerate(
...
     if (!command.editorName) {
       throw new Error("Editor name is required (--editor or -e)");
     }
-    await generateEditor(
-      command.editorName,
-      command.documentTypes?.split(/[|,;]/g) ?? [],
-      config,
-    );
+    await generateEditor({
+      name: command.editorName,
```
ryanwolhuter, line 102: I've updated these to use an object arg because there were way too many similar args and they were getting mixed up in some places

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-drive-editor/components/CreateDocument.esm.t
```diff
@@ -1,5 +1,5 @@
 ---
-to: "<%= rootDir %>/<%= h.changeCase.param(name) %>/components/CreateDocument.tsx"
+to: "<%= driveEditorDir %>/components/CreateDocument.tsx"
```
ryanwolhuter, line 2: prefer to use the existing dir name for a given editor, so like editors/some-editor. Otherwise we can accidentally end up creating a new one if the dir name is different from the editor name.

packages/codegen/src/codegen/hygen.ts
```diff
@@ -45,7 +45,12 @@ export async function run(
 
     const filesToFormat = actions
       .filter((action) => ["added", "inject"].includes(action.status))
-      .map((action) => action.subject.replace(".", process.cwd()));
+      .map((action) => action.subject.replace("./", `${process.cwd()}/`))
```
ryanwolhuter, line 48: this caused a bug because replacing "." can end up doing this in unintended places like in files with an extension

acaldas (approved):

PR #2101, opened 2025-11-13
feat(ph-cmd): create vetra document on init

gpuente:


clis/ph-cmd/src/utils/validate-remote-drive.ts
```diff
@@ -20,13 +24,32 @@ export async function validateRemoteDrive(
...
+        console.log("✅ Vetra package document created successfully");
+
+        // Re-fetch documents after creation
+        documents = await getVetraDocuments(graphqlEndpoint, driveId!);
+
+        if (documents.length === 0) {
+          console.error(
+            "❌ Failed to create vetra package document in the remote drive",
+          );
+          return false;
```
acaldas, line 43: If `getVetraDocument` returns no documents then the logs will be:
```
> ✅ Vetra package document created successfully
> ❌ Failed to create vetra package document in the remote drive
```

Let's simplify it to be a single success or error message.

PR #2107, opened 2025-11-18
feat(codegen): use ts-morph for document models and editors and index files

ryanwolhuter:
In our codegen, any hygen logic which needs knowledge of files and their contents outside of the arguments directly passed to hygen is problematic. Hygen has inconsistent async behavior, which means that it can easily have stale data.

The only two places where we actually need this extra information is when we create the files which export the collections of document models and editors, i.e.

`editors/editors.ts`
`document-models/document-models.ts`

These are to be removed in the future but are currently needed for backwards compatibility

`editors/index.ts`
`document-models/index.ts`

We can avoid this problem while also making our code much safer and easier to work with by using ts-morph instead.

We can initialize a ts-morph `Project` using just the files needed for generating this code, and then use ts-morph's knowledge of the whole file system to accurately get the names of each of the module variables exported by each `module.ts` file.

See the code and comments in `packages/codegen/src/ts-morph-utils.ts` for detailed explanations of how this is achieved.


packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-document-model/document-models.esm.t
ryanwolhuter, line 1: let ts-morph handle this instead

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-document-model/lib.esm.t
ryanwolhuter, line 1: let ts-morph handle this instead

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-drive-editor/editors.esm.t
ryanwolhuter, line 1: let ts-morph handle this instead

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-drive-editor/lib.esm.t
ryanwolhuter, line 1: let ts-morph handle this instead

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-editor/editors.esm.t
ryanwolhuter, line 1: let ts-morph handle this instead

packages/codegen/src/codegen/.hygen/templates/powerhouse/generate-editor/lib.esm.t
ryanwolhuter, line 1: let ts-morph handle this instead

packages/codegen/src/codegen/.hygen/templates/powerhouse/utils.js
ryanwolhuter, line 1: this logic was always problematic because of race conditions — we are much better off without it

packages/codegen/src/codegen/__tests__/generate-doc-model.test.ts
ryanwolhuter, line 1: I had to make these timeouts super long because of how slow the internet here is. We can change it back when the world's wires are fixed

acaldas (approved):

PR #2135, opened 2025-12-12
feat(codegen,vetra): use ts morph codegen by default

ryanwolhuter:
Currently we have been setting the flag for using the ts-morph based codegen to false as default, with the user needing to explicitly opt in to use it with the `--ts-morph` flag.

With this update, we will instead use the the ts-morph codegen by default, with the user being able to opt-in to the legacy hygen codegen with the `--use-hygen` flag.

The `--ts-morph` flag is now removed and replaced with the `--use-hygen` flag.

The various codegen test suites for the legacy code still passes, but I have changed the default test command to now use the ts-morph tests instead for the sake of saving time in the ci flow. You can use the `test:hygen` test command to test the legacy codegen.

I have run the new codegen on the vetra code and confirmed that it produces valid output. The only difference it made was to add some test cases for the modules that cover operations that had no corresponding cases.


clis/ph-cli/src/services/generate.ts
```diff
@@ -46,7 +46,7 @@ export async function startGenerate(
 ) {
   const baseConfig = getConfig();
   const useVersioning = !!options.useVersioning;
-  const legacy = useVersioning || !options.tsMorph;
+  const useTsMorph = useVersioning || !options.useHygen;
```
ryanwolhuter, line 49: the versioning logic is still opt-in with the `--use-versioning` flag. But since the versioning only works with ts-morph, I have set this to not allow using the versioning with the legacy codegen.

PR #2138, opened 2025-12-15
feat: add support for operations with no inputs

gpuente:
This PR adds support for empty operation inputs.

Since GraphQL does not allow empty input types (e.g. `input MyInput {}`), I added a checkbox in the document model editor that enables creating empty operation inputs. Under the hood, this generates an input with an optional `_empty` field:

```graphql
input MyInput {
  _empty: Boolean
}
```

This approach prevents breaking the `schema.graphql` file and the codegen pipeline.

During code generation, empty operations default to an empty object in the action creator:

```ts
export const setMyDocEmptyOperation = (
  input: SetMyDocEmptyOperation = {}
) => {
  ...
}
```

This allows the action creator to be called without arguments:
```ts
actions.setMyDocEmptyOperation()
```

### DIFF
--- a/packages/builder-tools/document-model-editor/components/module.tsx
+++ b/packages/builder-tools/document-model-editor/components/module.tsx
@@ -27,6 +27,7 @@ type Props = {
     errorId: string,
     name: string,
   ) => void;
+  toggleNoInputRequired: (id: string, noInputRequired: boolean) => void;
 };
 export function Module(props: Props) {
   const {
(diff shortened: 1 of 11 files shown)


acaldas (approved):

acaldas:
Would be nice to improve the UI/UX a little bit, @CallmeT-ty.
I'd change the text to "Operation with no input".
Maybe we could disable the input text area instead of hiding it? The issue is that then the `_empty` field will show up there right?

CallmeT-ty:
- tbh, the whole document model document model probably needs a big review and update.
I think it could slightly better looking if it was a gray on/off toggle instead of blue checkbox,
With the toggle being integrated in the code box. In case of no inputs the code box content would become low opacity grayed out? 

But i'd maybe deprioritise optimizing the document-model document further until we can focus on the whole UX and integrate a document upgrade functionality aswell?

PR #2145, opened 2025-12-17
feat: integrate doc model versioning in reactor

gpuente:
## Summary

This PR implements document model versioning support in the reactor, enabling documents to be upgraded from one version to another via `UPGRADE_DOCUMENT` operations.

### Key Changes

**Codegen**
- Set `version` field on generated `DocumentModelModule` objects

**Registry**
- Changed storage from `Map<string, Module>` to flat array for version-aware lookup
- Added `getModule(documentType, version?)` - returns specific version or latest
- Added `getSupportedVersions()`, `getLatestVersion()` methods
- Added upgrade manifest support: `registerUpgradeManifests()`, `getUpgradeManifest()`, `computeUpgradePath()`, `getUpgradeReducer()`
- New error classes: `DuplicateModuleError`, `ModuleNotFoundError`, `DuplicateManifestError`, `ManifestNotFoundError`, `MissingUpgradeTransitionError`

**Executor**
- Version-aware module selection using `docMeta.state.version` from DocumentMetaCache
- When `version === 0`: uses latest module (document not yet upgraded)
- When `version >= 1`: uses specific version module
- Upgrade chain execution: computes upgrade path via registry, applies reducers sequentially

**Utilities (util.ts)**
- `parseVersionNumber(versionString)`: Parses version strings to integers
- `getUpgradeTargetVersion(action)`: Extracts toVersion from action with fallback
- Refactored `applyUpgradeDocumentAction()`:
  - Returns `null` for no-op (same version)
  - Throws `DowngradeNotSupportedError` for downgrade attempts
  - Accepts optional `upgradePath` for multi-step upgrades
  - Always applies `initialState` from action after transitions

**ReactorBuilder**
- Added `withUpgradeManifests(manifests)` method

### Tests
- 28 new registry versioning tests (`versioned.test.ts`)
- Unit tests for `applyUpgradeDocumentAction` scenarios
- Test fixtures with mock v1, v2, v3 modules and upgrade manifest

## Test Plan

- [x] Registry unit tests pass (28 tests)
- [x] Executor util unit tests pass
- [x] Existing tests pass (backward compatibility)
- [ ] Executor integration tests
- [ ] E2E tests

### DIFF
--- a/packages/reactor/src/executor/util.ts
+++ b/packages/reactor/src/executor/util.ts
@@ -4,13 +4,60 @@ import type {
   DeleteDocumentAction,
   PHDocument,
   UpgradeDocumentAction,
+  UpgradeTransition,
 } from "document-model";
 import { createPresignedHeader, defaultBaseState } from "document-model/core";
+import { DowngradeNotSupportedError } from "../shared/errors.js";
 import type {
   ConsistencyCoordinate,
   ConsistencyToken,
 } from "../shared/types.js";
 import type { OperationWithContext } from "../storage/interfaces.js";
+
+/**
+ * Parses a version string to extract the major version number.
+ * Handles formats like "0", "1", "2", "1.0.0", "2.0", etc.
+ * Returns 0 for empty/invalid strings or "0"/"0.0.0".
+ *
+ * @param versionString - The version string to parse
+ * @returns The extracted major version number
+ */
+export function parseVersionNumber(versionString: string): number {
+  if (!versionString) return 0;
+
+  const parsed = parseInt(versionString, 10);
+  if (!isNaN(parsed)) return parsed;
+
+  const parts = versionString.split(".");
+  if (parts.length > 0) {
+    const major = parseInt(parts[0], 10);
+    if (!isNaN(major)) return major;
+  }
+
+  return 0;
+}
+
+/**
+ * Extracts the target version from an UPGRADE_DOCUMENT action.
+ * First tries action.input.toVersion, then falls back to
+ * initialState.document.version for legacy operations.
+ *
+ * @param action - The UPGRADE_DOCUMENT action
+ * @returns The target version number
+ */
+export function getUpgradeTargetVersion(action: UpgradeDocumentAction): number {
+  let toVersion = action.input.toVersion;
+  if (toVersion === 0) {
+    const initialState = action.input.initialState as
+      | { document?: { version?: number } }
+      | undefined;
+    const stateVersion = initialState?.document?.version;
+    if (typeof stateVersion === "number") {
+      toVersion = stateVersion;
+    }
+  }
+  return toVersion;
+}
 
 /**
  * Creates a PHDocument from a CREATE_DOCUMENT action input.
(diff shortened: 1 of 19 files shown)


packages/reactor/src/executor/util.ts
```diff
@@ -4,14 +4,61 @@ import type {
...
 
+/**
+ * Parses a version string to extract the major version number.
+ * Handles formats like "0", "1", "2", "1.0.0", "2.0", etc.
+ * Returns 0 for empty/invalid strings or "0"/"0.0.0".
+ *
+ * @param versionString - The version string to parse
+ * @returns The extracted major version number
+ */
+export function parseVersionNumber(versionString: string): number {
```
thegoldenmule: I don't think this is needed anymore?

packages/reactor/src/registry/interfaces.ts
```diff
@@ -1,43 +1,126 @@
...
+    ...manifests: UpgradeManifest<readonly number[]>[]
+  ): void;
+
+  /**
+   * Get the upgrade manifest for a document type.
+   *
+   * @param documentType The document type identifier
+   * @returns The upgrade manifest, or undefined if not found
+   */
+  getUpgradeManifest(
```
thegoldenmule: We're throwing an error in getModule() if the module is found, which allows us to leave off the `| undefined` (and cleans up a lot of the code that uses it). I think we should do the same here to keep it all the same, and so we can leave of the `undefined`.

packages/reactor/src/executor/util.ts
```diff
@@ -4,14 +4,61 @@ import type {
...
+
+/**
+ * Extracts the target version from an UPGRADE_DOCUMENT action.
+ * First tries action.input.toVersion, then falls back to
+ * initialState.document.version for legacy operations.
+ *
+ * @param action - The UPGRADE_DOCUMENT action
+ * @returns The target version number
+ */
+export function getUpgradeTargetVersion(action: UpgradeDocumentAction): number {
```
thegoldenmule: Same here -- this is `number` now, and it should always be defined or it's an invalid upgrade.

packages/reactor/src/executor/util.ts
```diff
@@ -72,32 +119,71 @@ export function createDocumentFromAction(
...
+ * @param upgradePath - Optional pre-computed upgrade path for multi-step upgrades
+ * @returns The upgraded document, or null if no-op (same version)
+ * @throws DowngradeNotSupportedError if attempting to downgrade
  */
 export function applyUpgradeDocumentAction(
   document: PHDocument,
   action: UpgradeDocumentAction,
-): PHDocument {
+  upgradePath?: UpgradeTransition[],
+): PHDocument | null {
```
thegoldenmule: Returning `| null` makes every caller a bit messier (lots of null checks). There's no real reason to, as this already throws exceptions. Just throw more so that you can always return a `PHDocument`.

packages/reactor/src/executor/util.ts
gpuente: added fix in 3bd7f553c3ab3c30e58dad35bedc98baca28d836

packages/reactor/src/executor/util.ts
gpuente: added fix in 3bd7f553c3ab3c30e58dad35bedc98baca28d836

packages/reactor/src/executor/util.ts
gpuente: added fix in 3bd7f553c3ab3c30e58dad35bedc98baca28d836

packages/reactor/src/registry/interfaces.ts
gpuente: added fix in 3bd7f553c3ab3c30e58dad35bedc98baca28d836

PR #2147, opened 2025-12-18
Perf/e2e todo profiling

Samyoul:


Samyoul:
With #2228 closing now

PR #2149, opened 2025-12-18
perf: add performance profiling to two-reactor GQL sync tests

Samyoul:
- Add performance marks and measures to all test cases to isolate actual sync operation timing from setup/teardown
- Add logPerformanceMarks() helper to output performance data to console for correlation with CPU profiles
- Add profile-two-reactor-gql-tests.sh script for CPU profiling with Node.js --cpu-prof flag


Samyoul:
with #2228 closing

PR #2173, opened 2025-12-31
feat(codegen): restore ph-default-styles wrapper in editor templates

gpuente:


gpuente:
this PR is not longer needed, this changes were already added to main

PR #2216, opened 2026-01-15
Implement cursor pagination in GraphQL adapters

Samyoul:
## What's Changed?

- Use `result.nextCursor` instead of `result.options.cursor` for cursor field
- Fix `hasNextPage` to check if `nextCursor` exists (was always false)
- Fix `hasPreviousPage` to check if `options.cursor` exists (was always false)
- Export `redis` cache module in document-drive cache index

## Why Make the Change?

I was having an issue with counting the total number of documents stored by Switchboard. Switchboard (or at lease my usage of it) seemed to have a few problems with `findDocuments` pagination.

The `findDocuments` GraphQL query looks like it had broken / unimplemented pagination, making it impossible to retrieve more than the limit of documents per type.

### Issues Found

  1. `hasNextPage` always returns false
    - Even when there are 1000+ documents, the first page returns `hasNextPage: false`
  2. `cursor` always returns null
    - No cursor is provided to fetch subsequent pages
  3. offset parameter is ignored
    - paging: `{ limit: 100, offset: 100 }` returns the same documents as offset: 0
  4. `totalCount` is incorrect
    - Returns the count of items in the current response, not the total matching documents
    - With `limit: 1`, it returns `totalCount: 1` even if there are 1000 documents

### Impact

  - Can't count total documents accurately
  - Can't retrieve documents beyond the first 100 per type
  - My profiling/testing scripts can't verify document creation at scale

### Reproduction

```gql
  # Returns same results regardless of offset
  query {
    findDocuments(
      search: { type: "powerhouse/document-model" }
      paging: { limit: 5, offset: 0 }
    ) {
      items { id name }
      hasNextPage  # always false
      cursor       # always null
      totalCount   # equals items.length, not actual total
    }
  }
```

### `totalCount` Still Incomplete (Medium Priority)

As the AI reviewer has pointed out, I'm aware that there is slightly more to do to fix all issues with cursor and pagination.

I'm happy to implement here or in a follow on PR.

### DIFF
--- a/packages/document-drive/src/cache/index.ts
+++ b/packages/document-drive/src/cache/index.ts
@@ -1,2 +1,3 @@
 export * from "./memory.js";
+export * from "./redis.js";
 export * from "./util.js";
--- a/packages/reactor-api/src/graphql/reactor/adapters.ts
+++ b/packages/reactor-api/src/graphql/reactor/adapters.ts
@@ -30,9 +30,9 @@ export function toDocumentModelResultPage(
 ): DocumentModelResultPage {
   const models = result.results.map((module) => module.documentModel);
   return {
-    cursor: result.options.cursor || null,
-    hasNextPage: false,
-    hasPreviousPage: false,
+    cursor: result.nextCursor ?? null,
+    hasNextPage: !!result.nextCursor,
+    hasPreviousPage: !!result.options.cursor,
     items: models.map(toGqlDocumentModelState),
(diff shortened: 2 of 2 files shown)


Samyoul:
I'm also mindful that there maybe other reasons the cursor / pagination wasn't implemented in the adapter. So I'm open to being gently shepherded to this understanding :) .

thegoldenmule (approved):

PR #2219, opened 2026-01-16
Update editor boilerplate

acaldas:


acaldas:
@claude fix the following issues:
Fix Tailwind important modifier syntax in json-viewer.tsx
Add useMemo for scopes array to prevent unnecessary re-renders
Add onKeyDown handler for Enter key
Fix state synchronization with Tabs component
Change the className array to string type
Rename ScopeLabel to follow conventions

Issue #2177, opened 2026-01-02
BUG: Field \"id\" of required type \"OID!\" was not provided.",


CallmeT-ty:
**Describe the bug**
Invalid input for AddTodoItemInput 

<img width="1250" height="886" alt="Image" src="https://github.com/user-attachments/assets/f77ca876-886c-493c-8536-d618b90ef035" />

<img width="845" height="1028" alt="Image" src="https://github.com/user-attachments/assets/f7ffe7f3-93c8-4565-befd-f5ce26557a6f" />

**To Reproduce**

**Resulting behavior**
A clear and concise description of what happens, the bug itself.

**Expected behavior**

**Logs and/or Screenshots**

**Desktop (please complete the following information):**

**Additional context**

**How will regressions be avoided in the future**

- Describe tests
- Describe solutions

CallmeT-ty:
I was still able to reproduce this issue.

<img width="1705" height="1011" alt="Image" src="https://github.com/user-attachments/assets/12c242ba-ff94-4114-9c12-6c51cafe8b26" />

PR #2241, opened 2026-01-21
Work in progress on the switchboard-e2e tests

CallmeT-ty:
Still working on the search subgraph function


CallmeT-ty:
Search function through the processor is now added to the test. But there is an issue with the mutations. 
I'm not fully understanding it but it sounds similar to this issue we tried to solve. https://github.com/powerhouse-inc/powerhouse/issues/2177

> > The subgraph schema at subgraphs/todo-list/schema.ts:58-61 defines TodoList_AddTodoItemInput with both id: OID! and text: String! as required fields, but the document model at document-models/todo-list/gen/schema/types.ts:62-64 only defines text as required.
> 
>   Field "id" of required type "OID!" was not provided.
> 
>   This means there are two different schemas:
>   1. Supergraph schema: Only requires text: String!
>   2. TodoList subgraph schema: Requires id: OID! AND text: String!
> 
>   The error is happening at the subgraph level
>   (http://localhost:4001/graphql/todo-list), not the supergraph level.

gpuente:
The tests look good, but you should remove `drives.test.ts` and `processor.test.ts` since they duplicate the tests already covered in `reactor-core.test.ts`.

PR #2247, opened 2026-01-22
docs(academy): add CSS customization guide for Connect

gpuente:


CallmeT-ty (approved):
lgtm!

apps/academy/docs/academy/02-MasteryTrack/03-BuildingUserExperiences/04-CSSCustomization.md
```diff
@@ -0,0 +1,217 @@
...
+}
+```
+
+:::tip
+Using `height: "100%"` ensures your editor fills the available vertical space within Connect's container hierarchy.
+:::
+
+### Method 2: CSS File with Container Selectors
+
+For more complex customizations or when you need to override Connect's default container styles, you can target the container IDs directly in a CSS file:
```
acaldas, line 55: We don't want users to do this as their changes will affect all editors. At least until we somehow make it so a package's css only applies to their editor.

One solution would be to add `data-editor=[editorId]` to allow setting rules only for a specific editor.

gpuente: I added the `data-editor=[editorId]` and `data-document-type=[documentType]` attributes for specific style targeting [755e636](https://github.com/powerhouse-inc/powerhouse/pull/2247/commits/755e636e982eb779bdd572387004350f87e0b497)

PR #2261, opened 2026-01-27
fix(reactor-api): enable document-model subgraph generation

gpuente:
## Description

The issue was caused by `powerhouse/document-model` being explicitly excluded when the subgraph feature was introduced, along with a schema generation bug where mutations were generated for operations with `schema: null`, causing Apollo Federation "Unknown type" errors. The exclusion was likely a workaround for another bug

Screenshot 2026-01-27 at 4 12 21 PM
Screenshot 2026-01-27 at 4 12 37 PM


acaldas (commented):
Nice.
Can you add a test, maybe in reactor-api, that uses the graphql api to create a document model and uses the mutations?
Then it will be easier to check for regressions in the future when we update the document model spec

acaldas (approved):

PR #2278, opened 2026-01-30
feat(profiling): add reactor-direct.ts for direct reactor performance profiling

Samyoul:
Bypass GraphQL API to isolate reactor performance. Supports:
  - Creating N documents with M operations per document
  - Targeting existing documents via --doc-id flag
  - Multiple database backends (in-memory PGlite, PostgreSQL, file-based)
  - Operation loop iterations and percentile statistics
  
## Usage

```
powerhouse % tsx ./scripts/profiling/reactor-direct.ts 1 -o 25 -l 10 --db "postgresql://postgres:postgres@localhost:5432/postgres" --pyroscope http://localhost:4040
    
Initializing Pyroscope profiler at: http://localhost:4040
  Wall and CPU profiling enabled
Initializing reactor directly (no GraphQL API)...
  Connecting to PostgreSQL: postgresql://postgres:***@localhost:5432/postgres
  Running database migrations...
Reactor initialized in 7.46s

Initial memory: heap: 112.7MB/211.5MB, rss: 282.1MB

Phase 1: Creating 1 documents...
  Progress: 1/1
  Created 1 documents in 0.20s (avg: 196ms/doc)
  Memory: heap: 113.9MB/211.8MB, rss: 283.4MB

Phase 2: Performing 25 operations x 10 loops on each document...
  [1/1] 55d2975d-ac18-4dbc-8fd8-00d8fc9e22c1: loop 1/10: 25/25 ops (1.24s, 49ms/op, min: 32ms, max: 167ms)
  [1/1] 55d2975d-ac18-4dbc-8fd8-00d8fc9e22c1: loop 2/10: 25/25 ops (1.27s, 51ms/op, min: 28ms, max: 102ms)
  [1/1] 55d2975d-ac18-4dbc-8fd8-00d8fc9e22c1: loop 3/10: 25/25 ops (0.82s, 33ms/op, min: 20ms, max: 43ms)
  [1/1] 55d2975d-ac18-4dbc-8fd8-00d8fc9e22c1: loop 4/10: 25/25 ops (0.89s, 36ms/op, min: 21ms, max: 67ms)
  [1/1] 55d2975d-ac18-4dbc-8fd8-00d8fc9e22c1: loop 5/10: 25/25 ops (1.17s, 47ms/op, min: 21ms, max: 85ms)
[...]

PR #2296, opened 2026-02-04
Perf/get doc last

Samyoul:
## Summary

  - Optimize `getDocumentLastModified` from O(n log n) to O(n) by replacing sorting with a linear scan
  - Further optimize `updateHeaderRevision` hot path from O(n) to O(1) by passing the timestamp directly from call sites
  - Add missing `@types/node` dependency to profiling scripts

## Problem

See https://github.com/powerhouse-inc/powerhouse/pull/2278#issuecomment-3847081667

Pyroscope profiling identified `getDocumentLastModified` as a performance bottleneck. The function was:
1. Creating 3 intermediate arrays (`Object.values`, `flatMap`, `slice`)
2. Sorting all operations twice (O(n log n)) just to get the last one's timestamp
3. Called on every action dispatch via `updateHeaderRevision`

## Solution

**First optimisation:** Replace the sort with a linear scan to find the operation with the highest index/skip:           
  ```typescript                                                                                                            
  // Before: O(n log n) with 3 array allocations                                                                           
  const sortedOperations = sortOperations(                                                                                 
    Object.values(document.operations).flatMap((ops) => ops || []),                                                        
  );                                                                                                                       
  return sortedOperations.at(-1)?.timestampUtcMs;                                                                          
```

  ```ts
  // After: O(n) with zero allocations                                                                                     
  for (const ops of Object.values(document.operations)) {                                                                  
    for (const op of ops) {                                                                                                
      if (!latest || op.index > latest.index || ...) latest = op;                                                          
    }                                                                                                                      
  }                                                                                                                        
  return latest?.timestampUtcMs;       
  ```                                                                                    
                                                                                                                           
Second optimisation: Pass the already-known timestamp from call sites, avoiding the scan entirely in the hot path:
`reducer.ts`: `action.timestampUtcMs` is already available
```ts
updateHeaderRevision(newDocument, action.scope, action.timestampUtcMs);
```

`documents.ts`: `lastOperation.timestampUtcMs` is already available   
```ts                                                   
updateHeaderRevision(result, lastOperation.action.scope, lastOperation.timestampUtcMs);
```

---

Pyroscope shows no hotspots coming from `getDocumentLastModified()`

<img width="2820" height="1034" alt="image" src="https://github.com/user-attachments/assets/44eaabf9-796c-4e59-832e-c9b23035c355" />
<img width="1404" height="607" alt="Screenshot 2026-02-05 at 13 23 27" src="https://github.com/user-attachments/assets/c39a3bd5-8596-4d99-977e-807393bb2d57" />

### DIFF
--- a/packages/document-model/src/core/documents.ts
+++ b/packages/document-model/src/core/documents.ts
@@ -344,6 +344,7 @@ export function replayDocument<TState extends PHBaseState = PHBaseState>(
         result = updateHeaderRevision(
           result,
           lastOperation.action.scope,
+          lastOperation.timestampUtcMs,
         ) as PHDocument<TState>;
       }
     }
@@ -1147,13 +1148,15 @@ function getNextRevision(document: PHDocument, scope: string) {
  * Updates the document header with the latest revision number and
  * date of last modification.
  *
- * @param state The current state of the document.
- * @param operation The action being applied to the document.
+ * @param document The current state of the document.
+ * @param scope The scope of the operation.
+ * @param lastModifiedTimestamp Optional timestamp to use directly, avoiding a scan of all operations.
  * @returns The updated document state.
  */
 export function updateHeaderRevision(
   document: PHDocument,
   scope: string,
+  lastModifiedTimestamp?: string,
 ): PHDocument {
   const header: PHDocumentHeader = {
     ...document.header,
@@ -1161,7 +1164,8 @@ export function updateHeaderRevision(
       ...document.header.revision,
       [scope]: getNextRevision(document, scope),
     },
-    lastModifiedAtUtcIso: getDocumentLastModified(document),
+    lastModifiedAtUtcIso:
+      lastModifiedTimestamp ?? getDocumentLastModified(document),
   };
 
   return {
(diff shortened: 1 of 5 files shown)


packages/document-model/src/core/documents.ts
```diff
@@ -1138,21 +1148,24 @@ function getNextRevision(document: PHDocument, scope: string) {
...
+  lastModifiedTimestamp?: string,
 ): PHDocument {
   const header: PHDocumentHeader = {
     ...document.header,
     revision: {
       ...document.header.revision,
       [scope]: getNextRevision(document, scope),
     },
-    lastModifiedAtUtcIso: getDocumentLastModified(document),
+    lastModifiedAtUtcIso:
```
acaldas, line 1175: Looks good to me.
Not sure if it can happen, but to be safe we should only update the `lastModified` header if it's actually before `lastModifiedTimestamp ?? getDocumentLastModified(document)`.

Samyoul: That's a good point, I'll make that change.

acaldas (approved):

PR #2302, opened 2026-02-05
feat: clean drive subgraphs

gpuente:
## Description:

- added subgraphs auto generated for document drive (using new reactor)
- removed old drive subgraphs
- fixed some bugs related to subgraphs
- added GET REST endpoint for drives
- updated connect to work with the new drive endpoint


apps/connect/src/store/reactor.ts
```diff
@@ -137,16 +137,20 @@ export async function createReactor() {
...
-      {
-        type: "gql",
-        parameters: { url: remoteUrl },
-      },
-    );
+    try {
+      const response = await fetch(remoteUrl);
+      if (response.ok) {
+        const driveInfo = (await response.json()) as { id: string };
+        const { graphqlEndpoint } = parseDriveUrl(remoteUrl);
```
acaldas: `graphqlEndpoint` should be returned from the `/d/driveId` endpoint

apps/connect/src/utils/reactor.ts
```diff
@@ -96,12 +96,16 @@ export async function addDefaultDrivesForNewReactor(
...
+      const remoteName = `default-drive-${driveInfo.id}`;
       if (existingRemoteNames.has(remoteName)) {
-        // Remote already exists, skip adding it
         continue;
       }
-      await sync.add(remoteName, driveCollectionId("main", config.driveId), {
+
+      await sync.add(remoteName, driveCollectionId("main", driveInfo.id), {
         type: "gql",
         parameters: { url: config.graphqlEndpoint },
```
acaldas: If the `graphqlEndpoint` is returned from the endpoint then it no longer needs to be part of the config

packages/reactor-api/src/utils/create-schema.ts
```diff
@@ -14,6 +14,17 @@ import { GraphQLJSONObject } from "graphql-type-json";
 
 const logger = childLogger(["reactor-api", "create-schema"]);
 
+/**
+ * Operations whose input schemas reference output/union types in input position,
+ * which is invalid GraphQL. These operations are excluded from schema generation.
+ * - ADD_LISTENER: `input AddListenerInput { listener: Listener! }` — Listener is an output type
+ * - ADD_TRIGGER: `input AddTriggerInput { trigger: Trigger! }` — Trigger is a union type
+ */
+export const OPERATIONS_WITH_INVALID_INPUT_SCHEMAS = new Set([
```
acaldas: Let's update the graphql schema to fix this instead

packages/reactor-browser/src/actions/drive.ts
```diff
@@ -44,6 +44,7 @@ export async function addDrive(input: DriveInput, preferredEditor?: string) {
 export async function addRemoteDrive(
```
acaldas, line 44: Let's reuse this method in Connect to avoid code duplication

acaldas (changes requested):

apps/connect/src/store/reactor.ts
gpuente: fix here: [7fcc6f4](https://github.com/powerhouse-inc/powerhouse/pull/2302/commits/7fcc6f48dcf4f0e6842760e95150996d2d0c29b4)

apps/connect/src/utils/reactor.ts
gpuente: fix here: [7fcc6f4](https://github.com/powerhouse-inc/powerhouse/pull/2302/commits/7fcc6f48dcf4f0e6842760e95150996d2d0c29b4)

packages/reactor-api/src/utils/create-schema.ts
gpuente: fix here: [1fa1f65](https://github.com/powerhouse-inc/powerhouse/pull/2302/commits/1fa1f6500b5d98f6c5491e6369cc7ee4ba30e799)

packages/reactor-browser/src/actions/drive.ts
gpuente: fix here:

[9f09092](https://github.com/powerhouse-inc/powerhouse/pull/2302/commits/9f09092bc96d99454bd2a9943b2435d9d33175fd)

gpuente:
I also added an additional commit ([db9cfb9](https://github.com/powerhouse-inc/powerhouse/pull/2302/commits/db9cfb9b82c8f501a4cd30bcef16440793877404)) to fix an issue with the `Action` type in GraphQL.

The schema defines `input: JSONObject!`, but `JSONObject` only accepts objects, not primitives. As a result, actions like `SET_NAME` (string input) and `UNDO` / `REDO` (number inputs) could not be properly serialized through GraphQL, causing sync failures between reactors.

Updated action inputs to use object wrappers instead of primitives:

| Action     | Before            | After                       |
|------------|-------------------|-----------------------------|
| `SET_NAME` | `input: "my-doc"` | `input: { name: "my-doc" }` |
| `UNDO`     | `input: 1`        | `input: { count: 1 }`       |
| `REDO`     | `input: 1`        | `input: { count: 1 }`       |

## Notes

- **Public API remains unchanged**, function signatures are still `setName(name)`, `undo(count)`, and `redo(count)`
- Conversion to object format is handled internally within the action creators

acaldas:
> I also added an additional commit ([db9cfb9](https://github.com/powerhouse-inc/powerhouse/pull/2302/commits/db9cfb9b82c8f501a4cd30bcef16440793877404)) to fix an issue with the `Action` type in GraphQL.
> 
> The schema defines `input: JSONObject!`, but `JSONObject` only accepts objects, not primitives. As a result, actions like `SET_NAME` (string input) and `UNDO` / `REDO` (number inputs) could not be properly serialized through GraphQL, causing sync failures between reactors.
> 
> Updated action inputs to use object wrappers instead of primitives:
> 
> Action	Before	After
> `SET_NAME`	`input: "my-doc"`	`input: { name: "my-doc" }`
> `UNDO`	`input: 1`	`input: { count: 1 }`
> `REDO`	`input: 1`	`input: { count: 1 }`
> ## Notes
> * **Public API remains unchanged**, function signatures are still `setName(name)`, `undo(count)`, and `redo(count)`
> * Conversion to object format is handled internally within the action creators

This change has long been needed. But won't this change break existing documents?
What happens if an old document with a SET_NAME operation is replayed?

gpuente:
> > I also added an additional commit ([db9cfb9](https://github.com/powerhouse-inc/powerhouse/pull/2302/commits/db9cfb9b82c8f501a4cd30bcef16440793877404)) to fix an issue with the `Action` type in GraphQL.
> > The schema defines `input: JSONObject!`, but `JSONObject` only accepts objects, not primitives. As a result, actions like `SET_NAME` (string input) and `UNDO` / `REDO` (number inputs) could not be properly serialized through GraphQL, causing sync failures between reactors.
> > Updated action inputs to use object wrappers instead of primitives:
> > Action	Before	After
> > `SET_NAME`	`input: "my-doc"`	`input: { name: "my-doc" }`
> > `UNDO`	`input: 1`	`input: { count: 1 }`
> > `REDO`	`input: 1`	`input: { count: 1 }`
> > ## Notes
> > 
> > * **Public API remains unchanged**, function signatures are still `setName(name)`, `undo(count)`, and `redo(count)`
> > * Conversion to object format is handled internally within the action creators
> 
> This change has long been needed. But won't this change break existing documents? What happens if an old document with a SET_NAME operation is replayed?

@acaldas  Yes, this change will break existing documents with old SET_NAME operations... do we want to add backwards compatibility?? if so, I can find a way to make it work with old documents and add additional tests

PR #2304, opened 2026-02-06
Added nested operations query on phdocument and Graphql client update

acaldas:


acaldas:
@claude Implement the permission checks and add test coverage for new features

PR #2322, opened 2026-02-10
feat: preserve existing signatures and support resulting state hash

gpuente:
## Description:

- Preserve pre-signed actions when submitted through `ReactorClient`
- Add `signActionWithResultingState()` method for including resulting state hash in signatures                                                
- Add helper functions for parsing signature hash fields

### DIFF
--- a/packages/renown/test/script.test.ts
+++ b/packages/renown/test/script.test.ts
@@ -1,11 +1,22 @@
 import { ReactorBuilder, ReactorClientBuilder } from "@powerhousedao/reactor";
-import { RenownBuilder } from "@renown/sdk/node";
 import {
+  MemoryKeyStorage,
+  RenownBuilder,
+  RenownCryptoBuilder,
+  RenownCryptoSigner,
+} from "@renown/sdk/node";
+import {
+  actions,
+  type Action,
   type DocumentModelDocument,
   documentModelDocumentModelModule,
   setName,
+  type UserActionSigner,
 } from "document-model";
 import { describe, expect, it } from "vitest";
+
+// Note: RenownBuilder is used by the first test (requires network access)
+// The new test uses RenownCryptoBuilder + RenownCryptoSigner directly (no network access)
 
 describe("Renown on script", () => {
   it("should create a document and add a signed SET_NAME action", async () => {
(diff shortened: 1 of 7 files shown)


packages/renown/test/script.test.ts
```diff
@@ -60,4 +71,126 @@ describe("Renown on script", () => {
         .filter((sig) => sig.length > 0).length,
     ).toBeGreaterThan(0);
   });
+
+  it("should NOT overwrite pre-signed actions with resulting-state-hash format", async () => {
```
gpuente, line 74: Although the test passes, there's something that doesn't feel right about how this works...

Using `signActionWithResultingState` directly in the script feels weird... it breaks the regular flow. Right now, if I want to sign with the resulting state, I have to do something like:

```typescript
const signatureA = await signerA.signActionWithResultingState(
  actionWithContext,
  resultingStateHash,
);
```

And you also have to remember to get the latest operation hash yourself, build the action with the right context, etc... it's a lot of manual work that could go wrong.

I think a better approach would be for `reactorClient.execute` to know when to use `signAction` vs `signActionWithResultingState`...

Now, to do this we need to define at which level we want to set this strictness (signature level). Is it per document? Per drive? Per action?

I think the more granular, the better... so per action seems like the right candidate to me.

In that case, I would do something like:

```typescript
async execute<TDocument extends PHDocument>(
  documentIdentifier: string,
  branch: string,
  actions: Action[],
  signal?: AbortSignal,
  signWithResultingState?: boolean,
)
```

Or even better, change the execute signature to use an object... so we don't add an extra argument and we can support new properties in the future if needed:

```typescript
execute<TDocument extends PHDocument>(options: {
  documentIdentifier: string;
  branch: string;
  actions: Action[];
  signal?: AbortSignal;
  options?: {
    signWithResultingState?: boolean;
  };
}): Promise<TDocument>
```

This way, `execute` would take care of getting the latest operation for the document and branch, extract the resulting hash, and use it when signing...

What do you think guys?? @thegoldenmule @acaldas

thegoldenmule: @gpuente @acaldas I don't think we need to make the reactor interface more complicated, as we could accomplish this with signer and validator interface changes. Right now the signer and validator have context about only the current action, but the reactor could also pass the previous operation, which would let the signer decide if it wants to include previous hashes in the signature.

acaldas: @gpuente The unit test is just a straightforward way to test the feature, it's not meant to be the way devs will use this.

The use case is for example, creating a document locally on a script, add actions to it, sign them with a renown user and use the graphql api to push the document to a remote Switchboard instance.

The client doesn't need to have a reactor running, it can run the actions through a reducer to get the resulting hash.

Example code:
```ts
const todo = new TodoDocument();
todo.setSigner(signer);
todo.addTodo({ ... });

gqlClient.createDocument(...);
qglClient.mutate({ id: todo.id, actions: todo.operations.global} );
```

gpuente: ok, changed this PR to ready for review, fixed some issues and rebased with main.

packages/renown/test/script.test.ts
```diff
@@ -1,12 +1,23 @@
...
+  actions,
+  type Action,
   type DocumentModelDocument,
   documentModelDocumentModelModule,
   setName,
+  type UserActionSigner,
 } from "document-model";
 import { describe, expect, it } from "vitest";
 
+// Note: RenownBuilder is used by the first test (requires network access)
```
acaldas: Can you change the first test so that it also doesn't need network access? It was more of an experimental e2e test.

acaldas (approved):

PR #2345, opened 2026-02-16
feat: add document model versioning documentation

gpuente:


CallmeT-ty (approved):
Hi @gpuente Looks good to me. 
Only questions i have. 
- Generate function: Users won't make use of this a lot in the future, but yeah, we need to make it available i understand. 
- Vetra allows for upgrades once it's configured in the versions.ts file. 
- I imagine in the future we want to make all of this a bit more automated somehow as it seems like a rather manual process? 

Lets add this to the documentation so we have it available for v 6 and we can get some feedback from Prometheus.

PR #2367, opened 2026-02-19
Pre-allocate operation arrays to avoid resize overhead

Samyoul:
## Why Make the Change?

Profiling the reactor under load revealed two sources of O(n²) overhead in the write path that
compound as operation history grows:

1. Reducer (`updateOperationsForAction` / `updateOperationsForOperation`), each action called
`.slice()` on the full operations array to copy it, then `.sort()`, then `.push()`. With n operations
this means O(n) copy + O(n log n) sort on every write.
2. Write-cache (`KyselyWriteCache.putState`), the full document (including complete operation
history) was stored in the ring buffer. Every downstream reducer call then received this full
array and repeated the copy/sort pattern, accumulating O(n²) total work across n operations.

## What's Changed?

1: reducer pre-allocation:
Replace `slice()` + `sort()` + `push()` with a pre-allocated fixed-size array. Since the write-cache now
guarantees operations arrive sorted (last op is always `at(-1)`), the sort is unnecessary. Array
size is known upfront so no resize occurs.

2: write-cache operation slicing:
Store only the last operation per scope in the ring buffer. The reducer only needs `at(-1).index` to
determine the next index, so accumulating the full history in the cache was driving the O(n²)
pattern. UNDO/REDO actions are exempt, the executor invalidates the cache before loading to force
a cold-miss rebuild with full history from the operation store.

Also simplifies `getNextRevision` in `document-model` from a linear scan to `at(-1)` since operations
are stored in sorted order.

### Profiling Context

Top hot spots from a 120s wall-time profile before these changes:
- `baseReducer`: 7.4% wall (9.4× wall:CPU ratio)
- `updateOperationsForAction`: 4.3% wall
- GC: 3.7% wall (driven by short-lived array allocations)
- `document-model` module total: 13.8% wall

### Test changes

Cache integration tests that compared full document equality across a cold-miss and a cache-hit
were updated to compare `.state` and `.header` only, reflecting the new contract that cached documents
carry only the last operation per scope.

Screenshot 2026-02-20 at 11 28 38

### DIFF
--- a/packages/document-model/src/core/documents.ts
+++ b/packages/document-model/src/core/documents.ts
@@ -1139,12 +1139,7 @@ export function getDocumentLastModified(document: PHDocument) {
  */
 function getNextRevision(document: PHDocument, scope: string) {
   const scopeOperations = document.operations[scope];
-  let maxIndex = -1;
-  if (scopeOperations) {
-    for (const op of scopeOperations) {
-      if (op.index > maxIndex) maxIndex = op.index;
-    }
-  }
+  const maxIndex = scopeOperations?.at(-1)?.index ?? -1;
   return maxIndex + 1;
 }
 
--- a/packages/document-model/src/core/reducer.ts
+++ b/packages/document-model/src/core/reducer.ts
@@ -98,24 +98,23 @@ function updateOperationsForAction<TDocument extends PHDocument>(
   }
 
   const scope = action.scope;
-
-  let operations: Operation[] = [];
-  if (document.operations[scope]) {
-    operations = document.operations[scope].slice();
-  }
-
-  const latestOperation = operations.sort((a, b) => a.index - b.index).at(-1);
-  const lastOperationIndex = latestOperation?.index ?? -1;
+  const existing = document.operations[scope];
+  const lastOperationIndex = existing?.at(-1)?.index ?? -1;
 
   const index = reuseLastOperationIndex
     ? lastOperationIndex
     : lastOperationIndex + 1;
 
   const newOperation = operationFromAction(action, index, skip, context);
-  operations.push(newOperation);
 
-  // adds the action to the operations history with
-  // the latest index and current timestamp
+  // Pre-allocate array with exact size to avoid resize overhead
+  const existingLen = existing?.length ?? 0;
+  const operations = new Array<Operation>(existingLen + 1);
+  for (let i = 0; i < existingLen; i++) {
+    operations[i] = existing![i];
+  }
+  operations[existingLen] = newOperation;
+
   return {
     ...document,
     operations: { ...document.operations, [scope]: operations },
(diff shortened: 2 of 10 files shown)


Samyoul:
## Profiling: Reactor Direct 1 doc × 25k ops (from 2026-02-20)

**Command:** `tsx reactor-direct.ts 1 -o 25 -b 5 -l 1000 --pyroscope http://localhost:4040 --file`
**Workload:** 1 document × 25 ops × 1000 loops = 25,000 operations · batch size 5

### Results vs baseline (2026-02-16)

| Metric | 2026-02-20 | 2026-02-16 baseline | Change |
|:--|--:|--:|--:|
| Total time | 487.99s | 495.37s | -1.5% |
| Avg ms/op | 19ms | 20ms | -5.0% |
| Min ms/op | 7.2ms | 7.4ms | -0.2ms |
| Max ms/op | 129.2ms | 112.2ms | +15.2% |
| Heap delta | +23.8MB | +240.2MB | -90.1% |
| RSS delta | +57.6MB | +201.6MB | -71.4% |

Memory behaviour improved dramatically. Throughput is roughly flat vs baseline with a modestly higher tail ceiling.

### Degradation pattern

| Segment | Avg ms/op |
|:--|--:|
| First half (loops 1–500) | 13.29ms |
| Second half (loops 501–1000) | 25.70ms |
| Change | +93.4% |

Still degrades across the run, but the profile is well-behaved, only 3/1000 loops (0.3%) include an operation ≥ 100ms.

### Pyroscope wall profile (33.01s total)

| Module | Self time | % |
|:--|--:|--:|
| other (infra) | 11.05s | 33.5% |
| pg (driver) | 6.97s | 21.1% |
| kysely (query builder) | 6.77s | 20.5% |
| I/O (writev/writeBuffer) | 2.23s | 6.7% |
| zod (validation) | 2.20s | 6.7% |
| GC | 1.26s | 3.8% |
| node:buffer | 1.17s | 3.6% |
| document-view.js | 0.89s | 2.7% |
| **reducer.js** | **0.10s** | **0.3%** |

**`reducer.js` is at 0.3% self time**, down from 20.7% on the pre-fix baseline and 38.4% in the 50k run. The O(n²)
reducer bottleneck is gone at this operation count.

The dominant cost is now the query stack: `kysely` + `pg` + `pg-protocol` combined account for ~49% of wall self time.

Notable hotspot: `kysely/object-utils.js:freeze` at 903ms (2.7%), `Object.freeze` on every query result, candidate for disabling in production.

### Degradation root cause

The remaining degradation follows a **~100-op sawtooth cycle**, floor rises, spikes, then partially resets, consistent
with a PostgreSQL autovacuum or checkpoint cycle aligning to accumulated keyframe writes. The floor drifts upward
monotonically as keyframe table size grows, accounting for the 2× first-to-second-half slowdown. This could be a DB-side effect, not a code-path issue.

### Wall:CPU ratio: 0.94x

The process is essentially CPU-bound during the profile window, the async query path itself (event loop coordination,
buffer decoding, protocol parsing) is the dominant cost, not raw I/O wait.

thegoldenmule (approved):

PR #2386, opened 2026-03-03
Add --async flag to docs-create and auto-migrate in switchboard-pyroscope

Samyoul:
## Summary

This PR adds two independent improvements to the profiling toolset: an `--async` mode for `docs-create.ts` that saturates the switchboard queue as fast as possible, and automatic database migrations in `switchboard-pyroscope.sh` when running against Postgres.

---

## `docs-create.ts`: `--async` flag

Adds a two-phase async dispatch mode for high-throughput profiling runs.

**Phase 1 - dispatch:** Requests are sent sequentially without awaiting job completion. Every Nth job (configurable via `--async [N]`, default `N=100`) immediately starts a concurrent poll in the background as it's dispatched.

**Phase 2 - drain:** Once all requests are dispatched, remaining in-flight polls are awaited. Results stream to stdout as each poll resolves.

**Additional details:**
- `*Async` mutation variants return `String!` (a job ID) - the GraphQL selection set is omitted accordingly
- Compatible with `--batch-size`
- `--async-timeout <ms>` (default 30s) caps how long each poll will wait before printing a `TIMED OUT` warning to stderr and moving on
- Timed-out jobs are excluded from min/max/percentile aggregation and counted separately in the summary line
- `JOB_POLL_INTERVAL_MS` reduced from 200ms to 5ms to reflect actual job latency rather than poll overhead
- Per-loop dispatch timing (duration + ms/op) is logged to help identify bottlenecks
- Summary line notes when stats are sampled (e.g. `"stats from N sampled jobs"`) to avoid misleading aggregates

**Validation:**
- `--async N` must be a positive integer (rejects `0` and floats)
- `--async-timeout` validated as an integer
- Warns if `--async` is combined with `--operations 0`
- Job ID validated as a non-empty string before polling begins

---

## `switchboard-pyroscope.sh` - auto-migrate on Postgres

When `--postgres` is provided and `DATABASE_URL` is set, database migrations now run automatically before switchboard starts. This removes a common manual step that caused silent failures.

---

## README

- Added `pnpm --filter @powerhousedao/reactor-api run build:misc` as a required prerequisite (the missing `schema.graphql` in `dist/` caused the reactor subgraph to fail silently)
- Updated `--postgres` flag description to reflect the `DATABASE_URL` condition
- Added `--async` and `--async-timeout` to the flag reference table

 ---  
 
 # Next Steps
 
See https://github.com/powerhouse-inc/powerhouse/pull/2386#issuecomment-4031938786

### DIFF
--- a/scripts/profiling/README.md
+++ b/scripts/profiling/README.md
@@ -182,20 +182,21 @@ tsx docs-create.ts 1 -o 25 --async
 tsx docs-create.ts 1 -o 100 -b 10 --async
 ```
 
-| Flag                  | Short | Description                                                          |
-| --------------------- | ----- | -------------------------------------------------------------------- |
-| `N` (positional)      |       | Number of documents to create (default: 10)                          |
-| `--operations`        | `-o`  | Operations per loop (default: 0)                                     |
-| `--op-loops`          | `-l`  | Loops per document (default: 1)                                      |
-| `--batch-size`        | `-b`  | Operations per `mutateDocument` call (default: 1)                    |
-| `--doc-id`            | `-d`  | Use existing document(s), can be repeated                            |
-| `--endpoint`          |       | GraphQL endpoint (default: `http://localhost:4001/graphql`)          |
-| `--async`             |       | Use `*Async` mutation variants (fire-and-forget; returns job ID)     |
-| `--file`              |       | Write output to a timestamped file (default name: `docs-create.txt`) |
-| `--output`            | `-O`  | Write output to a specific file (no timestamp prefix)                |
-| `--verbose`           | `-v`  | Show detailed operation timings                                      |
-| `--percentiles`       | `-p`  | Show p50/p90/p95/p99 stats                                           |
-| `--show-action-types` | `-a`  | Show action names in min/max timings                                 |
+| Flag                  | Short | Description                                                                 |
+| --------------------- | ----- | --------------------------------------------------------------------------- |
+| `N` (positional)      |       | Number of documents to create (default: 10)                                 |
+| `--operations`        | `-o`  | Operations per loop (default: 0)                                            |
+| `--op-loops`          | `-l`  | Loops per document (default: 1)                                             |
+| `--batch-size`        | `-b`  | Operations per `mutateDocument` call (default: 1)                           |
+| `--doc-id`            | `-d`  | Use existing document(s), can be repeated                                   |
+| `--endpoint`          |       | GraphQL endpoint (default: `http://localhost:4001/graphql`)                 |
+| `--async`             |       | Use `*Async` mutation variants (fire-and-forget; returns job ID)            |
+| `--async-timeout`     |       | Max ms to wait for a polled job to reach terminal status (default: `30000`) |
+| `--file`              |       | Write output to a timestamped file (default name: `docs-create.txt`)        |
+| `--output`            | `-O`  | Write output to a specific file (no timestamp prefix)                       |
+| `--verbose`           | `-v`  | Show detailed operation timings                                             |
+| `--percentiles`       | `-p`  | Show p50/p90/p95/p99 stats                                                  |
+| `--show-action-types` | `-a`  | Show action names in min/max timings                                        |
 
 ### `docs-count.ts` — Count documents (fast)
 
--- a/scripts/profiling/README.md
+++ b/scripts/profiling/README.md
@@ -247,11 +247,11 @@ Starts the switchboard with [Pyroscope](https://pyroscope.io/) continuous profil
 ./scripts/profiling/switchboard-pyroscope.sh -r bun -m legacy --postgres "postgresql://postgres:postgres@localhost:5432/reactor"
 ```
 
-| Flag         | Short | Description                              |
-| ------------ | ----- | ---------------------------------------- |
-| `--runtime`  | `-r`  | Runtime: `node` (default) or `bun`       |
-| `--mode`     | `-m`  | Storage mode: `v2` (default) or `legacy` |
-| `--postgres` | `-p`  | PostgreSQL database URL                  |
+| Flag         | Short | Description                                                                                                                     |
+| ------------ | ----- | ------------------------------------------------------------------------------------------------------------------------------- |
+| `--runtime`  | `-r`  | Runtime: `node` (default) or `bun`                                                                                              |
+| `--mode`     | `-m`  | Storage mode: `v2` (default) or `legacy`                                                                                        |
+| `--postgres` | `-p`  | PostgreSQL database URL; sets `DATABASE_URL` — migrations run automatically before the server starts when `DATABASE_URL` is set |
 
 ## Infrastructure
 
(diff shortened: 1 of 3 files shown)


Samyoul:
# Next Steps

## Plan: Add per-status timestamps to JobInfo

### Goal

Record and expose the wall-clock time of every job status transition (`PENDING`, `RUNNING`, `WRITE_READY`, `READ_READY`, `FAILED`) so external consumers can compute precise durations like queue wait time (`PENDING`→`RUNNING`) or processing time (`RUNNING`→`WRITE_READY`).

### Changes by layer

#### 1. `packages/reactor/src/shared/types.ts`: Add `timestamps` type and field

Add a `JobTimestamps` type and a `timestamps` field on `JobInfo`:

```ts
type JobTimestamps = {
pendingAtUtcIso: string;
runningAtUtcIso?: string;
writeReadyAtUtcIso?: string;
readReadyAtUtcIso?: string;
failedAtUtcIso?: string;
};

// On JobInfo:
timestamps: JobTimestamps;
```

`pendingAtUtcIso` is required (always set on creation). The rest are optional since a job may not reach every
state. This replaces the need for separate `createdAtUtcIso`/`completedAtUtcIso`, those become derivable from
`timestamps.pendingAtUtcIso` and the terminal timestamp, but keep them for backwards compatibility.

#### 2. `packages/reactor/src/core/reactor.ts`: Populate `pendingAtUtcIso` on job creation

All six registerJob call sites (lines 528, 584, 641, 707, 759, 855) already construct a `JobInfo` with
`createdAtUtcIso`. Add a `timestamps` field seeded with `{ pendingAtUtcIso: createdAtUtcIso }` at each site.

#### 3. `packages/reactor/src/executor/simple-job-executor-manager.ts`: Record `runningAtUtcIso`

At line 143 where markRunning is called, capture `new Date().toISOString()` and pass it through so the tracker can
store it.

#### 4. `packages/reactor/src/job-tracker/in-memory-job-tracker.ts`: Record all transition timestamps

- `registerJob` (line 102): already receives the `JobInfo` with `timestamps.pendingAtUtcIso` set, no change needed.
- `markRunning` (line 106): set `timestamps.runningAtUtcIso = new Date().toISOString()`.
- `handleWriteReady` (line 60): set `timestamps.writeReadyAtUtcIso = new Date().toISOString()`.
- `handleReadReady` (line 73): set `timestamps.readReadyAtUtcIso = new Date().toISOString()`.
- `markFailed` (line 128): set `timestamps.failedAtUtcIso = new Date().toISOString()`.

Update the `IJobTracker` interface if `markRunning`/`markFailed` signatures change.

#### 5. `packages/reactor-api/src/graphql/reactor/schema.graphql`: Extend GraphQL schema

```gql
type JobTimestamps {
  pendingAt: DateTime!
  runningAt: DateTime
  writeReadyAt: DateTime
  readReadyAt: DateTime
  failedAt: DateTime
}

type JobInfo {
  id: String!
  status: String!
  result: JSONObject!
  error: String
  createdAt: DateTime!        # keep for backwards compat
  completedAt: DateTime       # keep for backwards compat
  timestamps: JobTimestamps!  # new
}
```

#### `6. packages/reactor-api/src/graphql/reactor/adapters.ts`: Map timestamps in `toGqlJobInfo`

Add a `timestamps` field to the return of `toGqlJobInfo` (line 111), mapping each `*UtcIso` field to the GraphQL
`DateTime` scalar.

#### 7. `packages/reactor-api/src/graphql/reactor/gen/graphql.ts`: Regenerate

Run the codegen to update generated types from the schema changes.

Files touched (8 files)

| File | Change |
|------|--------|
| `packages/reactor/src/shared/types.ts` | Add `JobTimestamps` type, add `timestamps` field to `JobInfo` |
| `packages/reactor/src/core/reactor.ts` | Seed timestamps at all 6 `registerJob` call sites |
| `packages/reactor/src/job-tracker/interfaces.ts` | Update `markRunning` signature if needed |
| `packages/reactor/src/job-tracker/in-memory-job-tracker.ts` | Record timestamps on every transition |
| `packages/reactor/src/executor/simple-job-executor-manager.ts` | Pass timestamp through `markRunning` |
| `packages/reactor-api/src/graphql/reactor/schema.graphql` | Add `JobTimestamps` type and field |
| `packages/reactor-api/src/graphql/reactor/adapters.ts` | Map timestamps in `toGqlJobInfo` |
| `packages/reactor-api/src/graphql/reactor/gen/graphql.ts` | Regenerate |

### Not in scope

- Removing `createdAtUtcIso`/`completedAtUtcIso` from `JobInfo`, keep for backwards compatibility.
- Persisting timestamps to a database, `InMemoryJobTracker` is the only implementation today, so in-memory
storage is sufficient.
- Modifying the `jobChanges` subscription, it already returns `JobInfo`, so it gets timestamps for free.

### Test impact

- `packages/reactor/test/job-tracker/unit.test.ts`, update to verify timestamps are set on each transition.
- Any tests constructing `JobInfo` objects will need the new required timestamps field.

Samyoul:
## Assessment: `opentelemetry-instrumentation-reactor` for per-job timing

### What the package does

`ReactorInstrumentation` runs inside switchboard, subscribes to the internal event bus, maintains three timestamp maps to correlate events across the job lifecycle, and publishes aggregated OTLP metrics to an external endpoint (`METRICS_ENDPOINT` env var).

The durations it measures are exactly what you want:

| Metric | Transition |
|--------|------------|
| `reactor.executor.job.duration` | `RUNNING` -> `WRITE_READY` |
| `reactor.job.total.duration` | `PENDING` -> `READ_READY`/`FAILED` |
| `reactor.readmodel.index.duration` | `WRITE_READY` -> `READ_READY` |

### Why it's not suitable for the profiling script's use case

1. **Histograms, not per-job data.** All timing instruments are histograms. They give aggregate distributions across many jobs (p50, p95, bucket counts) - there's no way to retrieve the timing for a specific job ID.
2. **No `job.id` attribute.** The only attribute on duration histograms is `job.success: true/false`. A specific dispatch cannot be correlated to a specific measurement.
3. **Infrastructure dependency.** You need an OTLP collector (Prometheus, Grafana, etc.) running to receive and query the data. The profiling script currently needs zero additional infrastructure.
4. **Export interval is 60 seconds by default.** A profiling run of 50 operations might complete before the first export fires. Even with `METRICS_EXPORT_INTERVAL_MS` lowered, the data wouldn't be queryable in real time.

### Where it IS a good fit

The package is well-suited for production monitoring, understanding queue depth trends, aggregate executor throughput, and failure rates over time in a deployed environment.

### Conclusion

For what the profiling script needs, the precise start time of a specific job from outside switchboard, the `JobInfo.startedAtUtcIso` plan is still the right approach. The OTEL package and the `startedAt` field are complementary: the OTEL package records the same event internally for aggregate metrics, while `startedAt` on `jobStatus` exposes the server-side timestamp per job to any HTTP client. The `InMemoryJobTracker.markRunning` call is the single point where both would be populated.

Samyoul:
see https://github.com/powerhouse-inc/powerhouse/pull/2405

Samyoul:
Also see https://github.com/powerhouse-inc/powerhouse/pull/2406

PR #2397, opened 2026-03-06
feat: add ph build command

ryanwolhuter:
Things to note:

We want the type declarations from typescript, the type declarations from tsdown, and the actual js exported code always to be in sync. This becomes more difficult when dealing with outdirs and root dirs. To make this easy, I have moved the index.ts files to the root of each project instead of having it in the src dir. This doesn't change anything from the dev side, but it makes it so that our bundled exports and typescript builds are always the same paths.

For packages intended to be bundled for the node platform, I have also renamed the index files to be index.mts. This is to make sure that we compile es-module code in a way that node will always recognize and process correctly. It makes no difference in the functionality of the file and works the same as other ts files.

Similarly, I have also removed a bunch of export routes from places like reactor browser. This was meant to help us divide what we ship, but we haven't really stuck to this and it makes things significantly messier in the bundler. We now have good code splitting in place, so the benefit from separate exports is also negligible anyway.

We have two modular pieces of state that are decoupled:

1. Package manager and its store — in charge of loading packages from the registry, node_modules, and local. First tries to load from node_modules, then hits the registry

2. Registry packages — keeps an up to date list of the packages available from the registry endpoint, and stores their installation statuses in local storage.


.github/actions/setup/action.yml
```diff
@@ -26,19 +26,18 @@ runs:
...
 
-    - name: Run bundle scripts
+    - name: Run misc build scripts
       shell: bash
-      run: pnpm build-bundle
+      run: pnpm build-misc
 
     - name: Run css build scripts
       shell: bash
       run: pnpm build-css
```
ryanwolhuter: bundle scripts might need stuff from the misc build. vite build might need stuff from the bundle build

apps/connect/src/components/modal/modals/MissingPackageModal.tsx
```diff
@@ -0,0 +1,47 @@
+import { useRegistryPackages } from "@powerhousedao/connect/hooks";
```
ryanwolhuter, line 1: This is the modal we can use for prompting the installation of missing packages. It works the same as the other modals.

apps/connect/src/components/app-loader.tsx
```diff
@@ -1,23 +1,27 @@
...
-    </Suspense>
-    <Suspense name="CookieBanner">
-      <CookieBanner />
-    </Suspense>
-    <Suspense name="ModalsContainer">
-      <ModalsContainer />
-    </Suspense>
-  </StrictMode>
-);
+export const AppLoader = (props: { localPackage?: VetraPackage }) => {
```
ryanwolhuter, line 8: All we need to do is pass the local package as a prop to the AppLoader. vite handles the rest in the dev server.

apps/connect/src/components/load.tsx
```diff
@@ -1,10 +1,11 @@
 import { useCheckLatestVersion } from "@powerhousedao/connect/hooks";
 import "@powerhousedao/connect/i18n";
 import { createReactor, useSetSentryUser } from "@powerhousedao/connect/store";
+import type { VetraPackage } from "@powerhousedao/reactor-browser";
 import { type ReactNode } from "react";
 
-export async function loadComponent() {
-  await createReactor();
+export async function loadComponent(localPackage?: VetraPackage) {
```
ryanwolhuter, line 7: We can then pass the local package to createReactor. it's already available when the vite dev server starts so this is instant. This also means that vite is aware of the source code for the local package and the hot module reloading sees it

apps/connect/src/hooks/useRegistryPackages.ts
```diff
@@ -0,0 +1,173 @@
...
+        for (const packageInfo of packageInfos) {
+          const existingPackage = newRegistryPackages[packageInfo.name];
+
+          if (!existingPackage) {
+            const packageSource = packageManager.getPackageSource(
+              packageInfo.name,
+            );
+            const status = getPackageStatusFromPackageSource(packageSource);
+            newRegistryPackages[packageInfo.name] =
+              makeRegistryPackageFromPackageInfo(packageInfo, status);
```
ryanwolhuter, line 51: It would be great if we could get rid of all of these redundant types and interfaces so that we don't have to do all of this converting. We could just use one type with the same fields.

apps/connect/src/store/processor-host-module.ts
```diff
@@ -1,18 +1,14 @@
...
-} from "@powerhousedao/shared/processors";
+import { createAnalyticsStore } from "@powerhousedao/reactor-browser";
+import { type IProcessorHostModule } from "@powerhousedao/shared/processors";
+import { getDb } from "../pglite.db.js";
 
 export async function createProcessorHostModule(): Promise<
   IProcessorHostModule | undefined
 > {
   try {
-    const { getDb } = await import("../pglite.db.js");
```
ryanwolhuter, line 10: this type of async import is almost always a bad idea, because vite is often unable to analyze them. This means that for some minor code splitting, we lose all other optimization. This also is just simply ignored in cases where a path is already statically imported elsewhere, which is the case for reactor-browser in connect like here.

apps/connect/src/store/reactor.ts
```diff
@@ -94,35 +89,13 @@ export async function createReactor() {
     .build();
 
   // initialize package manager
-  const registryCdnUrl = getDefaultRegistryCdnUrl();
   const packageManager = new BrowserPackageManager(
     phGlobalConfigFromEnv.routerBasename ?? "",
-    registryCdnUrl,
+    PH_PACKAGE_REGISTRY_URL,
```
ryanwolhuter, line 94: this value is defined as a global variable in the vite dev server, derived from first the env var and then the config file

apps/connect/package.json
```diff
@@ -19,15 +19,20 @@
...
   ],
   "exports": {
+    ".": {
+      "types": "./dist/start-connect.d.ts",
+      "import": "./dist/start-connect.js"
+    },
     "./main.js": {
-      "types": "./dist/src/main.d.ts",
-      "import": "./dist/src/main.js"
+      "types": "./dist/main.d.ts",
```
ryanwolhuter, line 27: I've left this here for now but it should be removed

apps/connect/tsconfig.json
```diff
@@ -4,19 +4,38 @@
     "outDir": "./dist",
     "emitDeclarationOnly": true,
     "paths": {
-      "@powerhousedao/connect/config": ["./src/connect.config.ts"],
-      "@powerhousedao/connect/assets/*": ["./assets/*"],
-      "@powerhousedao/connect/*": ["./src/*/index.ts"]
+      "@powerhousedao/connect/config": [
```
ryanwolhuter, line 7: tsconfig paths works now

apps/connect/vite.config.ts
```diff
@@ -1,37 +1,25 @@
...
-        external: ["/index.ts", "/style.css", /^react(-dom)?(\/.*)?$/],
-      },
-    },
-    server: {
-      watch: {},
-    },
-  };
-
-  return mergeConfig(baseConnectViteConfig, additionalViteConfig);
+export default defineConfig({
```
ryanwolhuter, line 7: Using a separate vite config for connect since while it does have some similarities to the vite config used by ph-cli, its fundamentally different in some important ways.

packages/analytics-engine/browser/src/BrowserAnalyticsStore.ts
```diff
@@ -1,25 +1,596 @@
+import type { PGlite } from "@electric-sql/pglite";
```
ryanwolhuter, line 1: I re-implemented this with raw sql because knex in browser and vite do not play nice.

PR #2399, opened 2026-03-09
feat: replace reactor dropdown with registry selector in package manager

gpuente:
## Description

- Replace reactor dropdown with registry selector in package manager

https://github.com/user-attachments/assets/76195214-a6ca-4c79-94df-786a3ecb7808

### DIFF
--- a/apps/connect/src/hooks/use-registry.ts
+++ b/apps/connect/src/hooks/use-registry.ts
@@ -1,0 +1,200 @@
+import { connectConfig } from "../connect.config.js";
+import type {
+  RegistryOption,
+  RegistryPackageInfo,
+  RegistryStatus,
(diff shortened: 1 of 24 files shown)


apps/connect/src/hooks/use-registry.ts
```diff
@@ -0,0 +1,200 @@
...
+    if (!apiUrl) {
+      setRegistryStatus("idle");
+      setAvailablePackages([]);
+      return;
+    }
+
+    let cancelled = false;
+    setRegistryStatus("connecting");
+
+    fetch(`${apiUrl}/packages`)
```
acaldas: Can we add a simple registry api client to reactor-browser? So it can be used on vetra.io.

acaldas (approved):

PR #2401, opened 2026-03-10
feat: auto-discover and install packages for unknown document types

gpuente:
## Description:

A `BrowserPackageLoader` (implements `IDocumentModelLoader`) is created during reactor startup and wired in via `ReactorClientBuilder.withDocumentModelLoader()`. When the `DocumentModelResolver` encounters an unknown document type during sync, it calls `loader.load(documentType)` which queries the registry API (`RegistryClient.getPackagesByDocumentType()`) to find the matching package, adds it to a pending list, and returns a held-open Promise.

React hooks (`usePendingInstallations`) drive a `PackageInstallModal` (design-system) that shows the pending packages. When the user clicks Install, the loader calls `packageManager.addPackage()`, resolves the Promise, and the sync job retries automatically. On Dismiss, the Promise rejects and the package is saved to localStorage. A `DismissedPackagesList` component in Settings shows these dismissed packages so the user can install them later.


apps/connect/src/components/modal/modals/settings/package-manager.tsx
```diff
@@ -63,40 +72,92 @@ export const ConnectPackageManager: React.FC = () => {
...
   const handleInstall = useCallback(
-    (packageName: string) => {
+    async (packageName: string) => {
       if (!effectiveRegistryUrl) {
         throw new Error("No registry selected");
       }
-      return packageManager?.addPackage(packageName, effectiveRegistryUrl);
+      try {
+        await packageManager?.addPackage(packageName, effectiveRegistryUrl);
+        loader?.removeDismissed(packageName);
```
acaldas: Really cool!

Instead of creating a separate package loader entity, can we make this feature part of the package manager?
It should make it easier to manage this logic internally instead of the consumer having to sync them.

gpuente: makes sense! added the changes here: [d65191a](https://github.com/powerhouse-inc/powerhouse/pull/2401/commits/d65191a392ff8fb236540f9af08c886e5ea751d4)

PR #2402, opened 2026-03-11
feat: add ph publish command and update ph install

gpuente:


clis/ph-cli/src/commands/publish.ts
```diff
@@ -0,0 +1,90 @@
...
+import { join } from "path";
+
+export const publish = command({
+  name: "publish",
+  description: `
+Publish a package to the Powerhouse registry. This is a thin wrapper around npm publish
+that automatically sets the registry URL.
+
+This command:
+1. Resolves the registry URL (--registry flag > powerhouse.config.json > PH_REGISTRY_URL env > default)
```
acaldas, line 18: Let's make `PH_REGISTRY_URL` have priority over the config file

clis/ph-cli/src/commands/install.ts
```diff
@@ -84,13 +73,30 @@ This command:
...
+        });
+        console.log("Dependency installed successfully 🎉");
+      } catch (error) {
+        console.error("❌ Failed to install dependencies");
+        throw error;
+      }
+
+      try {
+        console.log("⚙️ Updating styles.css file...");
+        updateStylesFile(dependenciesWithVersions, projectPath);
```
acaldas, line 119: I noticed we are import each package style on style.css but is it really needed?
We already import their compiled css.

gpuente:
this PR si not longer needed

PR #2405, opened 2026-03-13
Implemented Profiling Open Telemetry Metrics

Samyoul:
Adds OpenTelemetry metrics support to the profiling scripts, enabling real-time visibility into reactor performance during benchmarks via Prometheus. Building on the work introduced with https://github.com/powerhouse-inc/powerhouse/commit/67d5c31e59417a2c5ca32090a342d656885d1aa1 also see https://github.com/powerhouse-inc/powerhouse/blob/main/packages/opentelemetry-instrumentation-reactor/docs/taxonomy.md

## What's Been Added?                                                                                                      

- `reactor-direct.ts`: new `--otel [endpoint]` flag that initialises an OTel MeterProvider with OTLP HTTP export, hooks up ReactorInstrumentation, and tears down cleanly after the benchmark
- Docker Compose: adds otel-collector and prometheus services alongside the existing postgres and pyroscope containers
- `otel-collector-config.yaml`: new config wiring OTLP receivers (gRPC + HTTP) to a Prometheus scrape endpoint
- `prometheus.yml`: new config scraping the collector every 5s
- `run-reactor-direct.sh`: updated build sequence to include the opentelemetry-instrumentation-reactor package
- `switchboard-pyroscope.sh`: same build sequence update
- `README.md`: full metrics reference (counters, gauges, histograms), PromQL query examples, and new workflow sections for OTel and combined Pyroscope+OTel profiling

The PR screenshots show Prometheus graphs of average ms/op, fastest operations (P1), and slowest operations (P99) over time during a live benchmark run.

### Average ms/op(job)
Screenshot 2026-03-13 at 09 38 28

### Fastest Operations Over Time
Screenshot 2026-03-13 at 09 38 53

### Slowest Operations Over Time
Screenshot 2026-03-13 at 09 39 17

### DIFF
--- a/packages/opentelemetry-instrumentation-reactor/src/instrumentation.ts
+++ b/packages/opentelemetry-instrumentation-reactor/src/instrumentation.ts
@@ -203,16 +203,26 @@ export class ReactorInstrumentation {
       // window the observation is silently dropped, making the gauge appear
       // to drop to zero under load. The timeout makes the failure explicit.
       const TIMEOUT_MS = 2_000;
-      const depth = await Promise.race([
-        queue.totalSize(),
-        new Promise<never>((_, reject) =>
-          setTimeout(
-            () => reject(new Error("queue.totalSize() timed out")),
-            TIMEOUT_MS,
-          ),
-        ),
-      ]);
-      result.observe(depth);
+      let timeoutId: ReturnType<typeof setTimeout> | undefined;
+      try {
+        const depth = await Promise.race([
+          queue.totalSize(),
+          new Promise<never>((_, reject) => {
+            timeoutId = setTimeout(
+              () => reject(new Error("queue.totalSize() timed out")),
+              TIMEOUT_MS,
+            );
+          }),
+        ]);
+        result.observe(depth);
+      } catch (err) {
+        console.warn(
+          "[ReactorInstrumentation] queueDepth observation failed:",
+          err,
+        );
+      } finally {
+        clearTimeout(timeoutId);
+      }
     };
     this.metrics.queueDepth.addCallback(depthCb);
     this.observableCallbacks.push([this.metrics.queueDepth, depthCb]);
--- a/packages/opentelemetry-instrumentation-reactor/src/instrumentation.ts
+++ b/packages/opentelemetry-instrumentation-reactor/src/instrumentation.ts
@@ -199,10 +199,19 @@ export class ReactorInstrumentation {
 
     const depthCb: ObservableCallback = async (result) => {
       if (!this.metrics) return;
-      // queue.totalSize() is async (DB query). The OTel SDK expects observable
-      // callbacks to complete within the collection window; if this is slow
-      // under DB load the observation may be silently dropped for that scrape.
-      const depth = await queue.totalSize();
+      // queue.totalSize() is a DB query. If it exceeds the OTel collection
+      // window the observation is silently dropped, making the gauge appear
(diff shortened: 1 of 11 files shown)


Samyoul:
@claude

The `otel/opentelemetry-collector-contrib` image is distroless, no shell, no `wget`, no `curl`, just the collector binary. I tested both and neither is present, so the `wget` healthcheck won't work inside that container.

The health extension on 13133 is reachable, but there's nothing inside the image to hit it. `--version` is the pragmatic workaround for this setup since exact readiness timing isn't critical for a profiling script. For a production deployment a custom image with curl added would be the right fix, but this is a dev tool.

Samyoul:
# Plan: OTel metrics support in switchboard

## Overview

Switchboard already calls `ReactorInstrumentation.start()` but has no `MeterProvider` or OTLP exporter, so metrics are collected but never exported. The fix mirrors the pattern used in `reactor-direct.ts` and follows the Pyroscope initialisation model already in place.

## Step 1: Add OTel dependencies to switchboard's `package.json`
```
@opentelemetry/api
@opentelemetry/sdk-metrics
@opentelemetry/exporter-metrics-otlp-http
```

These are already in the workspace catalog (used by `scripts/profiling`), so no new versions need pinning.

## Step 2: Create `apps/switchboard/src/metrics.ts`

Mirror the pattern of `profiler.ts`. The function reads `OTEL_EXPORTER_OTLP_ENDPOINT` from env and returns a `MeterProvider` or `undefined`:
```ts
export function initMetricsFromEnv(env: NodeJS.ProcessEnv): MeterProvider | undefined {
  const endpoint = env.OTEL_EXPORTER_OTLP_ENDPOINT;
  if (!endpoint) return undefined;

  const meterProvider = new MeterProvider({
    readers: [
      new PeriodicExportingMetricReader({
        exporter: new OTLPMetricExporter({ url: `${endpoint}/v1/metrics` }),
        exportIntervalMillis: 5_000,
      }),
    ],
  });
  metrics.setGlobalMeterProvider(meterProvider);
  return meterProvider;
}
```

## Step 3: Call `initMetricsFromEnv()` in `apps/switchboard/src/index.ts`

After Pyroscope init, before `startSwitchboard()`:
```ts
const meterProvider = initMetricsFromEnv(process.env);
```

## Step 4: Shut down `MeterProvider` on exit

In the existing signal/shutdown handling in `index.ts`, add:
```ts
await meterProvider?.shutdown();
```

This ensures a final flush is exported before the process exits, same ordering as `reactor-direct.ts` (`meterProvider.shutdown()` before `instrumentation.stop()`).

## Step 5: Remove the TODO comment from `switchboard-pyroscope.sh`

Once the above is in place, the `OTEL_EXPORTER_OTLP_ENDPOINT` export in the script will actually take effect.

## Step 6: Update the README

Document `OTEL_EXPORTER_OTLP_ENDPOINT` in the `switchboard-pyroscope.sh` section alongside the existing Pyroscope env vars.

---

## Notes

- No changes needed to `ReactorInstrumentation`: it already uses `metrics.getMeter()` from the global provider, so setting the global provider before `instrumentation.start()` is sufficient.
- The `ReactorInstrumentation` init in `server.ts` can stay as-is, since `initMetricsFromEnv()` sets the global provider before `startSwitchboard()` is called.

Samyoul:
See https://github.com/powerhouse-inc/powerhouse/pull/2406

PR #2416, opened 2026-03-20
Introduce Pluggable Gateway and HTTP Adapter Interfaces

Samyoul:
## What's Changed?
 
Refactors `GraphQLManager` in `reactor-api` to introduce two clean adapter interfaces, separating concerns that were previously mixed together:
 
- **`IGatewayAdapter`** (`gateway/types.ts`) abstracts the GraphQL server layer, turning a `GraphQLSchema` into a Fetch API `FetchHandler`. Implemented by `ApolloGatewayAdapter` using `ApolloServer.executeHTTPGraphQLRequest()`.
- **`IHttpAdapter`** (`gateway/types.ts`) abstracts the HTTP framework layer, handling route mounting, handler registration, and middleware setup. Implemented by `ExpressHttpAdapter`, which encapsulates the subgraph prefix-dispatch map (formerly the `subgraphHandlers` Map and `path-to-regexp` loop in `GraphQLManager`).
 
Both adapters are injected into `GraphQLManager` via constructor with sensible defaults, leaving `server.ts` unchanged. The federation gateway server is also mounted through `httpAdapter` via a new `createApolloFetchHandler()` utility.
 
Three bugs uncovered during integration testing are also fixed:
 
- **OOM / infinite routing loop**: `ExpressHttpAdapter` was matching against `req.originalUrl` (includes query string) with `{ end: false }`, causing the `/graphql` gateway handler to intercept subgraph requests to `/graphql/r`, `/graphql/document-model`, etc., routing them back to the gateway indefinitely. Fixed by switching to `req.path` with exact matching.
- **Double `createSchema` on cache miss**: `subgraphHandlerCache` was skipping `createHandler` on repeat `_updateRouter()` calls but still re-running `createSchema`. The guard now covers both.
- **Infinite wait on server startup failure**: `ApolloGatewayAdapter#waitForServer` now throws after 100 attempts (~10s) rather than looping indefinitely.
 
Unit tests for `ExpressHttpAdapter` are included.
 
## Why Make the Changes?
 
`GraphQLManager` was conflating three distinct responsibilities: GraphQL server execution, HTTP routing, and subgraph management. The adapter interfaces make each concern independently testable and replaceable, reducing coupling and making future changes to either the GraphQL server (e.g. swapping Apollo) or the HTTP layer (e.g. swapping Express) localised to a single adapter implementation.
 
This is also a necessary foundation for upcoming work to evaluate different gateway and HTTP routing implementations. Having clean adapter boundaries means each alternative can be swapped in and tested in isolation, without requiring broad changes across the codebase each time.
 
The bug fixes address real failure modes exposed during integration: the OOM crash was a silent routing cycle, the schema double-creation was a latent performance bug, and the infinite wait could have caused deployments to hang silently on a failed server start.


Samyoul:
# Express Audit Report: `packages/reactor-api`

I've abstracted the vast majority of Express away from the application via the `IGatewayAdapter` and `IHttpAdapter` interfaces, however there are a small number of remaining Express-specific references in `packages/reactor-api` that couldn't be cleanly removed in this pass. This report catalogues each one, documenting whether the coupling is intentional, a temporary workaround, or something that needs active removal, along with a concrete roadmap for reaching a fully framework-agnostic `reactor-api`.
  
## 1. `src/graphql/gateway/adapter-http-express.ts`: Intentional
 
This file is the Express implementation of `IHttpAdapter`. All Express imports here are correct and expected. When a `FastifyHttpAdapter` is added, this file stays unchanged.
 
No action needed.
  
## 2. `src/graphql/gateway/factory.ts`: Intentional (thin coupling)
 
```ts
import type { Express } from "express";
// ...
case "express":
  return createExpressHttpAdapter(existingApp as Express | undefined);
```
 
The Express type import is needed only to cast the `existingApp?: unknown` parameter before passing it to `createExpressHttpAdapter`. This is the correct place for that coupling, the factory is the one spot that knows which concrete adapter maps to which type.
 
No action needed unless the `options.express` escape hatch in `server.ts` is removed (see item 5).
  
## 3. `src/graphql/sse.ts`: Needs replacement
 
```ts
import { createHandler } from "graphql-sse/lib/use/express";
import type { Request as ExpressRequest } from "express";
```
 
`graphql-sse` ships both an Express handler (`graphql-sse/lib/use/express`) and a framework-agnostic Fetch API handler (`graphql-sse/lib/use/fetch`). The current implementation uses the Express flavour, which means:
 
- The `contextFactory` receives an Express `Request` and reads `req.params` and `req.headers` from it.
- The SSE handler cannot be mounted via `httpAdapter.mount()`, it requires `app.use()`, so `graphql-manager.ts` drops through the `handle` escape hatch (see item 4).
- Auth context on the SSE request currently reads `req.user` set by the old Express middleware, which is no longer how auth flows (the `WeakMap` pattern replaced that). The `contextFactory` in `graphql-manager.ts` calls `this.getAdditionalContextFields()` rather than `getAuthContext(request)`, meaning SSE subscriptions currently bypass the new auth context threading.
 
**How to remove:** Replace with `graphql-sse/lib/use/fetch` (or the Node.js HTTP handler). The handler signature becomes `(request: Request) => Promise<Response>`, which plugs directly into `httpAdapter.mount(ssePath, handler)`. The `contextFactory` would receive the Fetch `Request` and call `getAuthContext(request)` like all other handlers.
  
## 4. `src/graphql/graphql-manager.ts`: Temporary workaround
 
```ts
// SSE handler is Express-specific; mount directly on the underlying app.
// TODO: replace sse.ts with a Fetch API handler so this can use httpAdapter.mount()
(
  this.httpAdapter.handle as { use: (path: string, h: unknown) => void }
).use(ssePath, sseHandler);
```
 
This is a deliberate escape hatch, already marked with a `TODO`. It exists solely because `sse.ts` produces an Express-compatible handler. Fixing item 3 removes this entirely, `#setupSSEHandler` becomes a standard `httpAdapter.mount()` call.
 
One comment in the method JSDoc also refers to "Express middleware", that would be updated as part of the same pass.
  
## 5. `src/server.ts`: Two independent references
 
### a) `options.express?: Express` (line 72)
 
```ts
import type { Express } from "express";
// ...
type Options = { express?: Express; ... }
// ...
const { adapter: httpAdapter } = createHttpAdapter("express", options.express);
```
 
This escape hatch lets callers pass an existing Express app (e.g. Switchboard, which wraps the app in Sentry). Once those callers migrate to the `IHttpAdapter`-first pattern, constructing `new ExpressHttpAdapter(theirApp)` themselves and passing it as `httpAdapter` rather than `express`, this option and the import can be removed. The `Options` type would gain `httpAdapter?: IHttpAdapter` instead.
 
### b) MCP server (lines 640-644)
 
```ts
// TODO: decouple reactor-mcp from Express
await setupMcpServer(
  { client: reactorClient, syncManager },
  httpAdapter.handle as Express,
);
```
 
`@powerhousedao/reactor-mcp` currently takes an Express app and calls `app.use('/mcp', ...)` on it internally. Until that package exposes a framework-agnostic API (e.g. a `FetchHandler` or an `IHttpAdapter`), the `handle` cast is the only option. This is a cross-package concern and requires changes in `reactor-mcp`, not `reactor-api`.
  
## 6. `src/services/auth.service.ts`: Deprecated method retained for SSE
 
```ts
import type { NextFunction, Request as ExpressRequest, Response as ExpressResponse } from "express";
 
export interface AuthenticatedRequest extends ExpressRequest { ... }
 
/** @deprecated Use authenticateRequest with the Fetch API instead. */
async authenticate(req: AuthenticatedRequest, res: ExpressResponse, next: NextFunction): Promise<void>
```
 
The `authenticate()` Express middleware and `AuthenticatedRequest` interface are kept because `sse.ts` still threads auth via the old `req.user` mutation pattern. Once `sse.ts` is replaced (item 3), nothing calls `authenticate()` anymore. At that point:
 
- Remove `authenticate()` and `AuthenticatedRequest`
- Remove all three Express imports from this file
 
The `authenticateRequest(request: Request)` Fetch API method (already present) is the permanent replacement.
  
## Summary
 
| Location | Why it exists | Prerequisite to remove |
|---|---|---|
| `adapter-http-express.ts` | IS the Express adapter | Never remove, add alternative adapters alongside it |
| `factory.ts` | Type cast for `existingApp` | Remove when `options.express` is dropped from `Options` |
| `sse.ts` | Uses `graphql-sse/lib/use/express` | Migrate to `graphql-sse/lib/use/fetch` |
| `graphql-manager.ts` (SSE mount) | `handle` escape hatch for Express SSE handler | Fixed automatically when `sse.ts` is migrated |
| `server.ts`: `options.express` | Allows callers to pass an existing app | Replace with `options.httpAdapter?: IHttpAdapter`; update callers |
| `server.ts`: MCP cast | `reactor-mcp` depends on Express app | Decouple `reactor-mcp` from Express (cross-package) |
| `auth.service.ts` | `authenticate()` still needed for SSE auth path | Fixed when `sse.ts` is migrated and `authenticate()` is deleted |
  
## Recommended order of work
 
1. Migrate `sse.ts` to `graphql-sse/lib/use/fetch`. This is self-contained and unblocks three other removals (items 3, 4, 6 above).
2. Delete `authenticate()` and `AuthenticatedRequest` from `auth.service.ts` (and the three Express imports).
3. Replace `options.express` with `options.httpAdapter` in `server.ts` `Options`; update callers (`reactor-local`, `switchboard`).
4. Decouple `reactor-mcp` from Express, this is the largest item and spans another package.
 
After steps 1-3, the only remaining Express reference outside `adapter-http-express.ts` and `factory.ts` will be the MCP cast.

Samyoul:
Completion here https://github.com/powerhouse-inc/powerhouse/pull/2431

Samyoul:
see https://github.com/powerhouse-inc/powerhouse/pull/2443 and https://github.com/powerhouse-inc/powerhouse/pull/2444

PR #2431, opened 2026-03-25
Remove Outstanding Express Coupling

Samyoul:
## What's Changed?
 
This is the follow-up to #2416 (pluggable gateway and HTTP adapter interfaces). It eliminates every remaining Express import from `reactor-api`, `reactor-mcp`, and `switchboard` outside of `adapter-http-express.ts`.
 
**SSE migration (`sse.ts` + `graphql-manager.ts`)**
 
`sse.ts` is migrated from `graphql-sse/lib/use/express` to `graphql-sse/lib/use/fetch`. `createGraphQLSSEHandler` now returns a typed `(req: Request) => Promise<Response>` and the `contextFactory` receives a Fetch API `Request`. The `SSEHandlerOptions` docs are updated to reflect that auth flows through the `WeakMap` pattern via `authFetchMiddleware`, consistent with all other handlers.
 
In `graphql-manager.ts`, the Express `handle` escape hatch and its `TODO` comment are removed. `#setupSSEHandler` now calls `httpAdapter.mount(ssePath, handler, { exact: true })` like any other handler. Auth middleware is correctly applied to the SSE handler before mounting, fixing a pre-existing bug where SSE subscriptions were bypassing authentication entirely.
 
**Auth service cleanup (`auth.service.ts`)**
 
The deprecated `authenticate()` Express middleware (~80 lines) and the `AuthenticatedRequest` interface are deleted, along with the three Express type imports. With the SSE migration complete there are no remaining callers. The `authenticateRequest(request: Request)` Fetch API method is the sole auth path.
 
**`IHttpAdapter` interface additions (`gateway/types.ts`)**
 
Two new methods are added to `IHttpAdapter`:
 
- `mountNodeRoute(method, path, handler)`: for Node.js-style `IncomingMessage`/`ServerResponse` route handlers, needed by the MCP SDK which cannot use the Fetch API path. Documented as an escape hatch for streaming protocols that require direct access to Node.js HTTP types.
- `setupSentryErrorHandler(sentry)`: registers the framework-specific Sentry error-capturing middleware after all routes are mounted. Each adapter calls the appropriate Sentry setup function for its framework. No-op if not relevant.
 
**`ExpressHttpAdapter` (`adapter-http-express.ts`)**
 
Implements both new interface methods. `mountNodeRoute()` delegates to `app.get/post/delete` passing `req.body` as the third argument. `setupSentryErrorHandler()` casts the `sentry` object and calls `setupExpressErrorHandler(this.#app)`.
 
**`factory.ts`**
 
`createHttpAdapter` drops the `existingApp` parameter and the Express type import. `createExpressHttpAdapter()` is called with no arguments.
 
**`server.ts`**
 
`options.express` and its Express import are removed from `Options`. `createHttpAdapter("express")` is called without a second argument. `setupMcpServer` now receives `httpAdapter` directly instead of `httpAdapter.handle as Express`, removing the last Express cast. The return value renames `app` to `httpAdapter` to reflect what it actually is.
 
**`reactor-mcp`**
 
`src/express.ts` is renamed to `src/mcp-routes.ts`. The Express app parameter is replaced with a minimal local `NodeRouteAdapter` interface using `node:http` types throughout. Routes are mounted via `httpAdapter.mountNodeRoute()`. `express` and `@types/express` are removed as package dependencies. `index.ts` exports are updated to reflect the rename.
 
**Switchboard (`switchboard/server.ts`)**
 
The monolith `express()` app creation and the early `Sentry.setupExpressErrorHandler(app)` call are removed. Sentry setup is replaced with `api.httpAdapter.setupSentryErrorHandler(Sentry)` after `initializeAndStartAPI` returns, which is the correct registration order. `api.app.mountRawMiddleware` is updated to `api.httpAdapter.mountRawMiddleware`.
 
**`reactor-local` (`server.ts`)**
 
Updated to use `api.httpAdapter` instead of `api.app`.
 
**Profiling script**
 
`scripts/profiling/switchboard-pyroscope.sh` removes the now-deleted `document-drive` build step.
 
**Tests**
 
- `graphql-manager.test.ts`: 4 SSE handler tests verifying correct mount path and auth middleware wrapping (confirming the auth-bypass fix); `makeDriveModule()` mock updated to include `state.global.schema` to prevent `buildSubgraphSchemaModule` crashing.
- `http-adapter-contract.ts`: 4 `mountNodeRoute()` contract tests covering POST/GET/DELETE invocation and method-based routing.
- `reactor-mcp/test/setup-mcp-server.test.ts`: expanded to cover POST handler registration and 405 responses for unsupported methods.
- `auth-chain.test.ts`: new end-to-end integration tests wiring real `AuthService` through `createAuthFetchMiddleware`, covering invalid token (401, handler not called), valid token (handler called with correct `AuthContext`), OPTIONS/GET/anonymous POST bypass, auth disabled passthrough, and no `WeakMap` leak on failure.
 
**Lockfile**
 
`pnpm-lock.yaml` updated after removing Express dependencies from `reactor-mcp`.
 
## Why Make the Changes?
 
PR #2416 introduced the adapter interfaces but left several Express references in place as deliberate workarounds, each documented with a `TODO`. This PR works through all of them. The result is that Express is now entirely contained within `adapter-http-express.ts` and `gateway/factory.ts`. Everything else in the stack is framework-agnostic.
 
The SSE auth bypass fix is a correctness issue that fell out of the migration: SSE subscriptions were previously calling `getAdditionalContextFields()` rather than routing auth through `getAuthContext(request)`, meaning authenticated context was never applied to subscription connections. The migration to the Fetch API handler path closes that gap.
 
The `setupSentryErrorHandler` addition to `IHttpAdapter` means Sentry registration is no longer a caller concern. Switchboard no longer needs to know which framework is underneath or when to call the right Sentry setup function; it just calls `httpAdapter.setupSentryErrorHandler(Sentry)` after startup.
 
With this merged, swapping Express for a different HTTP framework requires only adding a new adapter implementation. No changes are needed in `reactor-api`, `reactor-mcp`, `switchboard`, `reactor-local`, or any other package.


packages/reactor-api/src/graphql/gateway/adapter-http-express.ts
acaldas: `<typeof expressLib>` can be replaced with the Express type:`import type { Express } from "express"`;

acaldas (commented):
LGTM!

packages/reactor-api/src/graphql/gateway/adapter-http-express.ts
```diff
@@ -27,6 +27,13 @@ export class ExpressHttpAdapter implements IHttpAdapter {
     this.#app.use(this.#router);
   }
 
+  setupSentryErrorHandler(sentry: object): void {
+    const s = sentry as {
+      setupExpressErrorHandler(app: ReturnType<typeof expressLib>): void;
+    };
```
thegoldenmule, line 34: typescript is so great (shiver)

thegoldenmule (approved):

acaldas (approved):

PR #2438, opened 2026-03-26
refactor: reconfigure the build pipeline

thegoldenmule:
Goal:

- Run `pnpm build` from whatever package you're in and it does "the right stuff" to incrementally build that package.
- Instead of using `pnpm -r`, explicitly supply what you're doing in root build scripts.


apps/switchboard/package.json
```diff
@@ -2,26 +2,26 @@
...
   "type": "module",
   "version": "6.0.0-dev.129",
-  "main": "dist/src/index.js",
+  "main": "dist/index.mjs",
   "exports": {
     ".": {
-      "import": "./dist/src/index.js",
-      "types": "./dist/src/index.d.ts"
+      "types": "./dist/src/index.d.ts",
+      "import": "./dist/index.mjs"
```
ryanwolhuter, line 9: I'd recommend changing these files to .mts in the source code too, this helps typescript stay consistent with the type definitions that tsdown outputs

apps/switchboard/tsdown.config.ts
```diff
@@ -0,0 +1,16 @@
...
+export default defineConfig({
+  entry: [
+    "src/index.ts",
+    "src/server.ts",
+    "src/utils.ts",
+    "src/install-packages.ts",
+    "src/migrate.ts",
+  ],
+  platform: "node",
+  outDir: "dist",
```
ryanwolhuter, line 12: If you rename the entry files to .mts, then you can instead set clean to true and dts to true. This will make the bundle much smaller, since otherwise it will contain both the untreeshaken transpiled javascript and type definitions that tsc -b emits, along with the code that tsdown bundles

packages/analytics-engine/benchmarks/package.json
```diff
@@ -11,6 +11,7 @@
     "dist"
   ],
   "scripts": {
+    "build": "tsc --build",
```
ryanwolhuter: This is what caused the build issues before. If the build script does `tsc --build`, and we are already running `tsc --build` at the top level, then this command will either produce redundant output because it will be the same as what the top level invocation did, or it will produce a slightly different result and break things later on.

If `tsc --build` is all a package needs to build, then it doesn't need a build script at all. It will already be built in the top level step.

packages/common/package.json
```diff
@@ -31,6 +31,7 @@
   "sideEffects": false,
   "description": "The Powerhouse Common package contains the basic document model and drive UI (App) that is required to get started with Connect and visualize private & public apps.",
   "scripts": {
+    "build": "tsc --build",
```
ryanwolhuter: This package should definitely also be bundled

packages/config/package.json
```diff
@@ -24,6 +24,7 @@
     "access": "public"
   },
   "scripts": {
+    "build": "tsc --build",
```
ryanwolhuter: same here, should be bundled

packages/design-system/package.json
```diff
@@ -61,6 +61,7 @@
     "tsc": "tsc",
     "lint": "eslint",
     "create-icon-components": "tsx scripts/create-icon-components.ts",
+    "build": "tsc --build && pnpm run build:css && pnpm run build:misc",
```
ryanwolhuter: This package also needs to be bundled, and you can use the `copy` field in tsdown's config to do the build:misc step in the bundle process instead

packages/powerhouse-vetra-packages/tsdown.config.ts
```diff
@@ -0,0 +1,17 @@
...
+    "document-models/*/index.ts",
+    "editors/index.ts",
+    "editors/*/index.ts",
+  ],
+  platform: "browser",
+  outDir: "dist",
+  clean: true,
+  dts: true,
+  sourcemap: true,
+  external: [/^[^./]/],
```
ryanwolhuter, line 16: what's up with the external field here?

packages/reactor-api/package.json
```diff
@@ -9,20 +9,12 @@
...
-    "./graphql/*": {
-      "import": "./dist/src/graphql/*.js",
-      "types": "./dist/src/graphql/*.d.ts"
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.mjs"
     },
     "./vite": {
-      "import": "./dist/src/packages/vite-loader.js",
-      "types": "./dist/src/packages/vite-loader.d.ts"
+      "types": "./dist/src/packages/vite-loader.d.ts",
```
ryanwolhuter: also recommend using .mts here

packages/reactor-api/tsconfig.json
```diff
@@ -1,10 +1,12 @@
 {
   "extends": "../../tsconfig.options.json",
   "compilerOptions": {
+    "emitDeclarationOnly": true,
     "outDir": "./dist",
     "types": ["node"]
   },
   "include": ["**/*"],
+  "exclude": ["dist", "node_modules"],
```
ryanwolhuter: you don't need to add an exclude field for just dist and node_modules, typescript excludes those by default

packages/reactor-api/tsdown.config.ts
```diff
@@ -0,0 +1,12 @@
+import { defineConfig } from "tsdown";
+
+export default defineConfig({
+  entry: ["index.ts", "src/packages/vite-loader.ts"],
+  platform: "node",
+  outDir: "dist",
+  clean: false,
```
ryanwolhuter: same as above here, better to clean and let tsdown do the dts part

packages/registry/package.json
```diff
@@ -18,8 +18,8 @@
   "scripts": {
     "start": "node ./dist/cli.js",
     "dev": "bun run ./cli.ts",
-    "build:cli": "tsdown",
-    "postbuild:cli": "bun run ./copy-dirs.ts",
+    "build": "tsdown",
```
ryanwolhuter, line 21: This postbuild can also be done more easily and cleanly with the tsdown `copy` field

packages/renown/package.json
```diff
@@ -37,6 +37,7 @@
     "dist"
   ],
   "scripts": {
```
ryanwolhuter, line 39: this package should also be bundled

packages/shared/clis/file-system/version-info.ts
```diff
@@ -30,7 +30,10 @@ export async function getPhCliVersionInfo() {
   if (!projectPackageJsonExists) {
     return noProjectWarningMessage;
   }
-  const projectPackageJson = await readPackage({ cwd: projectPath });
+  const projectPackageJson = await readPackage({
```
ryanwolhuter, line 33: what's up with this normalize: false?

scripts/check-wrong-tsc-usage.ts
```diff
@@ -1,94 +0,0 @@
-#!/usr/bin/env tsx
```
ryanwolhuter, line 1: I see this was deleted, but its whole purpose was to prevent the usage of `tsc --build` where it should not be used

test/versioned-documents/tsdown.config.ts
```diff
@@ -0,0 +1,20 @@
+import { defineConfig } from "tsdown";
```
ryanwolhuter, line 1: I would caution against using a tsdown config like this in packages that are meant to simulate external packages — they are supposed to be build with `ph build`, so doing this will make the build output different from what would happen in a real world case

ryanwolhuter (changes requested):
This looks good 👍 

Primary concerns are:

- usage of tsc --build where it should not be used
- inconsistent source vs output file names (.ts vs .mts)
- inconsistent use of the clean and dts fields in tsdown configs
- some packages which are not meant to be built in this build step
- some packages should be bundled, this expands the scope of this PR but I think its worth doing as part of this to make sure it works with this new flow

packages/analytics-engine/benchmarks/package.json
thegoldenmule: Right, that is a main focus of this PR. We should not be treating tsc like our build system. We should be able to run `pnpm build` in any package and have it build what is necessary for that package. In packages, this can be achieved with tsdown or tsc --build. At the top level build, the script `pnpm build` does not use `tsc --build` directly.

scripts/check-wrong-tsc-usage.ts
thegoldenmule: See above reply.

packages/analytics-engine/benchmarks/package.json
ryanwolhuter: tsc --build is already the TypeScript project-reference build orchestrator. Running it at the repo root is sufficient: it computes dependency order and incrementally builds only what is stale. Putting "build": "tsc --build" in every package and then running pnpm -r build means multiple independent tsc -b processes are traversing overlapping parts of the same reference graph in parallel. That is redundant at best and race-prone at worst, especially around shared outputs and .tsbuildinfo. If a package only needs TypeScript project-reference compilation, the root tsc -b step already covers it. Package build scripts should be reserved for package-specific artifact builds, not for re-invoking the workspace TypeScript build coordinator.

thegoldenmule: I think the main difference, is that you should not run `tsc` directly anymore. Just use the build scripts, which are uniform everywhere (`pnpm build`). They do not double build.

apps/switchboard/package.json
thegoldenmule: done

apps/switchboard/tsdown.config.ts
thegoldenmule: done

packages/analytics-engine/benchmarks/package.json
thegoldenmule: Moving everything to tsdown

packages/renown/package.json
thegoldenmule: done

scripts/check-wrong-tsc-usage.ts
thegoldenmule: tsdown'd

packages/powerhouse-vetra-packages/tsdown.config.ts
thegoldenmule: This is to keep out stuff like `react`, `document-model`, etc which were getting bundled. Since this package is sorta special case (in that these are all loaded dynamically at runtime), the dedupe happens at runtime.

packages/registry/package.json
thegoldenmule: There are situations where some of the folders did not exist (like cdn-cache/) which caused tsdown's copy to throw. The one in the script has try/catches which "fix" this, though it's unclear to me why this step is needed at all-- I don't think any of the directories exist at all in CI.

packages/shared/clis/file-system/version-info.ts
thegoldenmule: `normalize: false` gets the raw values in the project.json. Apparently `workspace:*` prefixes are stripped if you don't use this.

packages/powerhouse-vetra-packages/tsdown.config.ts
ryanwolhuter: they shouldn't be getting bundled, externalizing all of the deps is the default behavior 

https://tsdown.dev/options/dependencies#default-behavior

If you see warnings about deps in the bundle when you run the build, its probably because there are things in devDependencies that should actually be in dependencies

packages/registry/package.json
ryanwolhuter: Whoops this is my fault — I put this script in there when I was fixing the original false positives in the end to end tests here. At the time the server expected those dirs to exist in the dir that its invoked from, so for it to work the same in dev and in build I copied over the folders. This can be deleted.

packages/shared/clis/file-system/version-info.ts
ryanwolhuter: ah I see, interesting

apps/switchboard/package.json
```diff
@@ -1,27 +1,27 @@
...
     },
     "./server": {
-      "import": "./dist/src/server.js",
-      "types": "./dist/src/server.d.ts"
+      "types": "./dist/src/server.d.ts",
```
ryanwolhuter: Sorry I forgot to say — when declaration files are emitted for .mts files, the output file extension becomes .d.mts instead of .d.ts

ryanwolhuter (approved):
nice looks solid, a few minor comments but I'm approving so long anyway.

@acaldas we need to sort out how this works with the release flow. I also think we should delete all of the unused flows because maintaining them all when things like this change is becoming a huge pain and people are still using various old ones sometimes.

packages/powerhouse-vetra-packages/tsdown.config.ts
ryanwolhuter: I actually need to push an update here anyway — this package needs to be bundled using the same logic as the `ph build` command uses so that its compatible with the other package loading stuff.

I think I will already have merged that by the time you come online, so this package's bundling will already be sorted 👍

PR #2443, opened 2026-03-27
Add FastifyHttpAdapter implementing IHttpAdapter

Samyoul:
Introduces adapter-http-fastify.ts with full IHttpAdapter parity:
- CORS via @fastify/cors, URL-encoded bodies via @fastify/formbody
- Connect/Express middleware support via @fastify/middie (mountRawMiddleware works both before and after listen(), enabling Vite dev-server integration)
- TLS handled via serverFactory so the pre-built http/https server is injected at listen() time, matching the same TlsOptions contract as ExpressHttpAdapter
- mountNodeRoute uses reply.hijack() to hand raw IncomingMessage/ServerResponse directly to the handler (used by reactor-mcp StreamableHTTP transport)
- factory.ts now accepts "fastify" as an HttpAdapterType


Samyoul:
builds on work introduced with https://github.com/powerhouse-inc/powerhouse/pull/2416

PR #2444, opened 2026-03-27
Add MercuriusGatewayAdapter implementing IGatewayAdapter

Samyoul:
Uses Mercurius + @mercuriusjs/gateway as an alternative to Apollo Server:

- createHandler: registers Mercurius on an isolated Fastify instance and bridges requests via fastify.inject(), returning a standard FetchHandler
- createSupergraphHandler: uses @mercuriusjs/gateway (Apollo Federation compatible) to compose subgraph services; subgraph URLs are fetched at startup via the _service { sdl } protocol
- updateSupergraph: atomically swaps the gateway Fastify instance so in-flight requests on the old instance drain normally
- attachWebSocket: reuses graphql-ws directly (same as Apollo adapter)
- AsyncLocalStorage threads the original Fetch Request through inject() so contextFactory receives the same Request that auth middleware populated in the WeakMap — no auth context is lost across the bridge
- factory.ts: GatewayAdapterType now includes "mercurius"

### DIFF
--- a/packages/reactor-api/package.json
+++ b/packages/reactor-api/package.json
@@ -70,6 +70,8 @@
     "@powerhousedao/shared": "workspace:*",
     "@renown/sdk": "workspace:*",
     "@fastify/cors": "^11.0.1",
+    "@mercuriusjs/gateway": "^5.2.0",
+    "mercurius": "^16.8.0",
     "@fastify/formbody": "^8.0.1",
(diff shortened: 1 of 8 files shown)


Samyoul:
builds on work introduced with https://github.com/powerhouse-inc/powerhouse/pull/2416

PR #2451, opened 2026-03-30
feat: add separate node and browser processor bundles

ryanwolhuter:
Add separate node and browser build steps in the `ph build` command.

When running `ph build`, we will now output to two directories in the outDir:

- `{outDir}/browser`
- `{outDir}/node`

These are built with tsdown's platform set to "browser" and "node" respectively.

We can then provide separate exports in package.json for each of these, like so:

```json
    ".": {
      "types": "./dist/index.d.ts",
      "browser": "./dist/browser/index.js",
      "node": "./dist/node/index.mjs"
    },
    "./document-models": {
      "types": "./dist/document-models/index.d.ts",
      "browser": "./dist/browser/document-models/index.js",
      "node": "./dist/node/document-models/index.mjs"
    },
    ... etc
```

By default, vite already supports the "browser" export condition as seen above, so this works when doing `ph connect build` out of the box.

Most of the codegen code does not need to change here. The only place where updates were needed is in the processors, which have additional logic for determining the "processor app" in which they should run.

I have added a new type definition for `ProcessorFactoryBuilder`, which is the type of the function which takes an `IProcessorHostModule` as a parameter and returns a `ProcessorFactory`.

We can then have two files in the processors dir:

- `processors/connect.ts`
- `processors/switchboard.ts`

The code in each of these files is identical:

```ts
import type { ProcessorFactoryBuilder } from "@powerhousedao/reactor";

export const processorFactoryBuilders: ProcessorFactoryBuilder[] = [];
```

The reason for the duplication is that it significantly simplifies codegen and actually reduces overall duplication.

Previously, we had a lot of brittle code which would try to dynamically import the appropriate code for each processor app, and this was done inside function bodies. This is _significantly_ more difficult to work with in the codegen, because the values we need to update are nested several layers deep in the source file.

With this change, we never need to update the contents of `processors/factory.ts` when generating processors. We can simply add this file when creating a new project and leave it alone. When we generate a new processor, we add it's factory builder to the array corresponding to the processor app. If both connect and switchboard are passed, we just add it to both arrays.

```ts
export const processorFactory = async (module: IProcessorHostModule) => {
// <---- both of the files export a variable of the same name, so this works for either
  const { processorFactoryBuilders } =
    module.processorApp === "connect"
      ? await import("./connect.js")
      : await import("./switchboard.js");

  const factories = await Promise.all(
    processorFactoryBuilders.map(
      async (buildFactory) => await buildFactory(module),
    ),
  );

  // Return the inner function that will be called for each drive
  return async (driveHeader: PHDocumentHeader): Promise<ProcessorRecord[]> => {
    const processors: ProcessorRecord[] = [];

    // Call each cached factory with the driveHeader
    for (const factory of factories) {
      const factoryProcessors = await factory(driveHeader, module.processorApp);
      processors.push(...factoryProcessors);
    }

    return processors;
  };
};
```

I have also updated some of the naming in the processor codegen — previously we were appending "Processor" to the end of the passed in processor's name. Many people use names like "something-something-processor", which lead to many instances of code generated with names like SomethingSomethingProcessorProcessor. This is undesirable. We don't need to enforce having the word "processor" in a processor's name, so we can instead leave that to the user's discretion.

Finally, I have changed what was previously `index.ts` in a generated processor to instead be called `processor.ts`. I then export both the factory and processor in the `index.ts` file instead. 

I did this because:

- putting the processor in the index file is not how index files are normally used.
- the previous usage of the index file made it so that you could only import the processor from a path like `processors/my-processor` and not the factory.


apps/connect/package.json
```diff
@@ -67,7 +67,7 @@
     "@powerhousedao/design-system": "workspace:*",
     "@powerhousedao/powerhouse-vetra-packages": "workspace:*",
     "@powerhousedao/reactor-browser": "workspace:*",
-    "@powerhousedao/shared": "workspace:",
```
ryanwolhuter, line 70: I found many instances of this in the codebase which was causing things to break, so I did a global find and replace. The * is important for proper dependency resolution in the workspace.

packages/codegen/src/codegen/generate.ts
```diff
@@ -171,100 +178,72 @@ export async function generateDocumentModel(args: GenerateDocumentModelArgs) {
     useVersioning,
     migrateLegacy,
   });
-  ensurePackageExportsWildcards();
-  ensureTsconfigPaths(packageName);
+  // await ensurePackageExportsWildcards();
```
ryanwolhuter, line 181: I have commented this out because of several issues it was causing:

- doing `package-name/document-models/` etc. is the exact opposite of what we agreed to do going forward

- used hand-coded implementations for getting and updating tsconfig and package.json when we already have robust libraries installed for doing this

- the functions take an argument for the dir, but this was not actually passed at invocation. This caused it to overwrite the tsconfig and package.json files in the tests/codegen package

- only ensured paths for document-models and editors when this would need to be done for the other exports as well

- arbitrarily doing this on some but not all of the generate commands creates inconsistency and is a waste

- making updates like this should be done in a migration not a normal generate call

acaldas (approved):

PR #2452, opened 2026-03-30
feat(connect,docker): load PH_PACKAGES from runtime JSON instead of build-time define

acaldas:
## Summary
- Move `PH_PACKAGES` from Vite `define` (baked into JS bundle) to a `ph-packages.json` static asset that can be replaced post-build
- Add `phPackagesPlugin` Vite plugin that emits the JSON at build time and serves it via middleware during dev
- Add `packages.config.ts` with async loader and runtime validation
- Update Docker entrypoint to write `ph-packages.json` from `PH_REGISTRY_PACKAGES` env var
- Add `PH_REGISTRY_URL` and `PH_PACKAGE_REGISTRY_URL` env vars to both Docker targets

Changing the package list no longer requires a full `ph connect build` — just restart the container with a new `PH_REGISTRY_PACKAGES` value.

## Test plan
- [ ] Build Connect and verify `ph-packages.json` exists in `dist/`
- [ ] Run Docker with `PH_REGISTRY_PACKAGES="@powerhousedao/versioned-documents"` and verify the JSON is written
- [ ] Run without `PH_REGISTRY_PACKAGES` and verify build-time defaults are used
- [ ] Dev server serves correct `ph-packages.json` via middleware


apps/connect/public/ph-packages.json
```diff
@@ -0,0 +1,3 @@
+{
+  "packages": ["∑"]
```
ryanwolhuter: what's up with this "∑" 😂 sigma male energy?

acaldas: fat finger!

PR #2463, opened 2026-04-08
fix(connect): prevent HMR page reloads and support hot module re-injection

acaldas:
## Summary
- Prevent full page reloads during development when codegen updates the manifest or backup documents
- Support hot module re-injection so updated document models and editors are applied without reloading
- Add e2e test to verify no page reload occurs when editing a document in the Vetra drive

### HMR reload prevention
- **ph-packages plugin**: Add `hotUpdate` hook with `order: "pre"` that filters modules imported only by `style.css`, preventing tailwind from triggering full-reload
- **vite-config**: Register `phPackagesPlugin` before tailwind so the hook runs first. Add `server.watch.ignored` for `backup-documents` and `.ph` directories

### HMR module re-injection
- **startConnect**: Returns `{ updateLocalPackage }` for consumers to call from `import.meta.hot.accept`
- **BrowserPackageManager**: Add `updateLocalPackage` method that re-registers the local package and notifies subscribers
- **IPackageManager**: Add `updateLocalPackage` to the interface
- **vetra-packages hook**: Replace duplicate document models in the reactor registry instead of silently skipping them

### Tests
- New `hmr-no-reload.spec.ts` e2e test that creates a package manifest, changes category, and verifies no page reload via Playwright's `load` event
- Fix `vetra-drive.spec.ts` and `navigateToVetraDrive` helper to use `exact: true` for heading match (avoids ambiguity with "Vetra Preview" drive in `--watch` mode)
- Update playwright config to use `pnpm vetra --watch` for codegen support

## Test plan
- [x] `hmr-no-reload.spec.ts` passes with fix, fails without
- [x] All other vetra-e2e tests pass (9/9, excluding pre-existing registry test failure)
- [x] Manual verification: editing vetra-package category in Connect does not reload the page

### DIFF
--- a/apps/connect/start-connect.tsx
+++ b/apps/connect/start-connect.tsx
@@ -8,8 +8,26 @@ import { AppLoader } from "./src/components/index.js";
  *
  * // main.tsx
  * import * as localPackage from "./index.js";
- * startConnect(localPackage);
+ * const { updateLocalPackage } = startConnect(localPackage);
+ *
+ * if (import.meta.hot) {
+ *   import.meta.hot.accept(["./index.js"], ([newModule]) => {
+ *     if (newModule) {
+ *       updateLocalPackage(newModule);
+ *     }
+ *   });
+ * }
  */
+
+// Type for Vite HMR modules
+export type ModuleNamespace = Record<string, any> & {
+  [Symbol.toStringTag]: "Module";
+};
+
+function updateLocalPackage(pkg: DocumentModelLib | ModuleNamespace) {
+  window.ph?.vetraPackageManager?.updateLocalPackage(pkg as DocumentModelLib);
+}
+
 export function startConnect(localPackage: Partial<DocumentModelLib>) {
   if (!window.ph) {
     window.ph = {};
@@ -18,4 +36,8 @@ export function startConnect(localPackage: Partial<DocumentModelLib>) {
   createRoot(document.getElementById("root")!).render(
     <AppLoader localPackage={localPackage as DocumentModelLib} />,
   );
+
+  return {
+    updateLocalPackage,
+  };
 }
--- a/packages/builder-tools/connect-utils/vite-config.ts
+++ b/packages/builder-tools/connect-utils/vite-config.ts
@@ -259,6 +259,11 @@ export function getConnectBaseViteConfig(options: IConnectOptions) {
   const config: InlineConfig = {
     configFile: false,
     mode,
+    server: {
+      watch: {
+        ignored: ["**/backup-documents/**", "**/.ph/**"],
+      },
+    },
     resolve: {
       tsconfigPaths: true,
     },
@@ -271,6 +276,12 @@ export function getConnectBaseViteConfig(options: IConnectOptions) {
       exclude: ["@electric-sql/pglite", "@electric-sql/pglite-tools"],
     },
     plugins: [
+      // phPackagesPlugin must be registered before tailwind so its hotUpdate
+      // hook runs first and can suppress HMR updates for codegen-generated
+      // files, preventing tailwind from triggering full page reloads.
+      phPackagesPlugin({
+        packages: phPackages,
+      }),
       ...plugins,
       // Externalize React so both Connect and dynamically loaded registry
       // packages share the same React instance via the import map in index.html.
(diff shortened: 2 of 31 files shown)


packages/builder-tools/connect-utils/vite-config.ts
```diff
@@ -259,6 +259,11 @@ export function getConnectBaseViteConfig(options: IConnectOptions) {
   const config: InlineConfig = {
     configFile: false,
     mode,
+    server: {
+      watch: {
+        ignored: ["**/backup-documents/**", "**/.ph/**"],
```
acaldas, line 262: Not really needed but might as well ignore them to reduce load on watcher

packages/reactor-browser/src/hooks/vetra-packages.ts
```diff
@@ -45,6 +45,9 @@ function updateReactorClientDocumentModels(packages: DocumentModelLib[]) {
       registry.registerModules(module);
     } catch (error) {
       if (DuplicateModuleError.isError(error)) {
+        const documentType = module.documentModel.global.id;
+        registry.unregisterModules(documentType);
+        registry.registerModules(module);
```
acaldas, line 50: Will reload the local package when it's changed.
This means that if a reducer is updated then Connect will run the updated code without needing to reload the page

packages/vetra/processors/codegen/document-handlers/document-codegen-manager.ts
```diff
@@ -4,7 +4,7 @@ import { logger } from "../logger.js";
 import type { BaseDocumentGen } from "./base-document-gen.js";
 import type { CodegenInput, Config } from "./types.js";
 
-const DEFAULT_DEBOUNCE_TIME = 3000; // 3 seconds
+const DEFAULT_DEBOUNCE_TIME = 1000; // wait 1 second between codegen calls
```
acaldas, line 7: We should replace this with some kind of locking mechanism or queue. For now I've reduced the debounce time to account for the faster codegen and vite server.

test/test-consumer-project/main.tsx
```diff
@@ -1,4 +1,12 @@
 import { startConnect } from "@powerhousedao/connect";
 import * as localPackage from "./index.js";
 
-startConnect(localPackage);
+const { updateLocalPackage } = startConnect(localPackage);
+
+if (import.meta.hot) {
+  import.meta.hot.accept(["./index.js"], ([newModule]) => {
```
acaldas, line 7: This `accept` prevents the page reload and allows us to propagate the updated local package.

ryanwolhuter: Unfortunate that we have to make this file more complicated, but I don't think there's another way

apps/connect/start-connect.tsx
```diff
@@ -8,8 +8,26 @@ import { AppLoader } from "./src/components/index.js";
...
+ *   import.meta.hot.accept(["./index.js"], ([newModule]) => {
+ *     if (newModule) {
+ *       updateLocalPackage(newModule);
+ *     }
+ *   });
+ * }
  */
+
+// Type for Vite HMR modules
+export type ModuleNamespace = Record<string, any> & {
```
ryanwolhuter, line 23: classic vite using some weird thing I've never even seen before

packages/builder-tools/connect-utils/vite-plugins/ph-packages.ts
```diff
@@ -25,6 +25,20 @@ export function phPackagesPlugin(options: PhPackagesPluginOptions): Plugin {
         next();
       });
     },
+    hotUpdate: {
+      order: "pre",
+      handler(ctx) {
```
ryanwolhuter, line 30: I recently renamed the main styles file in connect from index.css to style.css for consistency with the other stuff — might be a good idea to check for the .css extension instead of the file name, because the user might add more style files

packages/vetra/tsconfig.json
```diff
@@ -16,7 +16,7 @@
     }
   },
   "include": ["**/*", "./powerhouse.manifest.json"],
-  "exclude": ["dist", "node_modules", ".ph", "main.tsx", "index.html"],
+  "exclude": ["dist", "node_modules", ".ph", "index.html"],
```
ryanwolhuter, line 19: how did index.html end up in here? afaik typescript doesn't include html files like this

ryanwolhuter (approved):
nice!

PR #2476, opened 2026-04-16
feat: add offline preview for installed packages

gpuente:
## Summary

Adds a `--local` flag to `ph install` that installs packages into  
`node_modules` and marks them as `provider: "local"` in  
`powerhouse.config.json`.

`ph connect build` bundles every `provider: "local"` package into the  
Connect output, so `ph connect preview` can run offline without the registry.

---

### Install

`ph install --local <pkg>` runs `<pm> add <pkg@version>` with per-scope  
registry routing (no project `.npmrc` needed) and writes  
`provider: "local"` to the config.

### Build

- Reads `provider: "local"` entries from `powerhouse.config.json`
- Fails fast if any are missing from `node_modules`
- A new `phBundledPackagesPlugin` exposes a `virtual:ph-bundled-packages` module with literal imports
- Rolldown bundles them into Connect's chunks (React stays externalized)

### Runtime

Bundled packages are registered via the same `addLocalPackage` path as  
Common/Vetra, so they are indistinguishable at runtime.


clis/ph-cli/src/services/connect-build.ts
```diff
@@ -1,25 +1,40 @@
...
-  const { outDir, debug } = args;
+  const { outDir, debug, bundlePackages } = args;
   assignEnvVars(args);
 
+  const mode = "production";
+  const dirname = process.cwd();
+
+  // Install bundled packages first so they're resolvable by both the
+  // project tsdown build and the Vite build that follow.
+  if (bundlePackages) {
```
acaldas: I think it's safer to require the packages to be installed by the user. The builds should throw an error if he tries to bundle packages that aren't installed.
For example, `ph install/add` could have a "--local" flag that would install the package.
This also makes it more flexible, as the user is able to install from the registry, from npm, or just link a package in the filesystem.

packages/registry/src/verdaccio-config.ts
```diff
@@ -12,12 +12,18 @@ export function buildVerdaccioConfig(config: RegistryConfig) {
         file: htpasswdPath,
       },
     },
-    uplinks: undefined,
+    uplinks: {
```
acaldas: Let's avoid the uplink if possible

apps/connect/src/package-manager.ts
acaldas: There is already a way to load local packages, like Common and Vetra. It would be good to handle it the same way instead of adding the concept of bundledPackages

apps/connect/src/packages.config.ts
```diff
@@ -1,6 +1,7 @@
 export type PackagesConfig = {
   packages: string[];
   localPackage?: { name: string; version: string } | null;
+  bundledPackages?: string[];
```
acaldas: I don't think this is needed.
Connect loads the packages to install from `/ph-packages.json`, this JSON is generated at build time:
https://github.com/powerhouse-inc/powerhouse/blob/65c7140dc76d1ad9c0a4f05ba93366beed6ef8a8/packages/builder-tools/connect-utils/vite-plugins/ph-packages.ts#L67-L72

This was implemented to allow changing the packages without needing to rebuild Connect. You just need to update the json file.
Here's how it's done in the dockerfile:
https://github.com/powerhouse-inc/powerhouse/blob/65c7140dc76d1ad9c0a4f05ba93366beed6ef8a8/docker/connect-entrypoint.sh#L11-L16

Since the local packages are known at build time, they don't need to be added here. Actually, I believe they will be part the `packages` array in the json, but the package manager should ignore them as they will already be loaded:
https://github.com/powerhouse-inc/powerhouse/blob/65c7140dc76d1ad9c0a4f05ba93366beed6ef8a8/apps/connect/src/package-manager.ts#L270-L275

clis/ph-cli/src/services/connect-build.ts
gpuente: fixed in [4ea3570](https://github.com/powerhouse-inc/powerhouse/commit/4ea3570da886877773c629e78fcd325a34ab5c1a)

packages/registry/src/verdaccio-config.ts
gpuente: fixed in [4ea3570](https://github.com/powerhouse-inc/powerhouse/commit/4ea3570da886877773c629e78fcd325a34ab5c1a)

apps/connect/src/package-manager.ts
gpuente: fixed in [4ea3570](https://github.com/powerhouse-inc/powerhouse/commit/4ea3570da886877773c629e78fcd325a34ab5c1a)

apps/connect/src/packages.config.ts
gpuente: fixed in [4ea3570](https://github.com/powerhouse-inc/powerhouse/commit/4ea3570da886877773c629e78fcd325a34ab5c1a)

apps/academy/docs/academy/04-APIReferences/00-PowerhouseCLI.md
```diff
@@ -826,14 +826,17 @@ the CLI to act on behalf of your Ethereum identity for authenticated operations.
...
-- 4. Updates style.css with CSS imports if applicable
+By default it only registers the package in powerhouse.config.json with
+provider "registry" — Connect will load it from the registry CDN at runtime.
+
+With --local, the package is also installed into node_modules and marked
+as provider "local" — it will be bundled into ph connect build so the
+preview works without the registry being reachable.
+
+Resolution order for the registry URL:
+  --registry flag > powerhouse.config.json > PH_REGISTRY_URL env > default
```
acaldas: PH_REGISTRY_URL should be higher than config.json

clis/ph-cli/src/commands/install.ts
```diff
@@ -2,27 +2,30 @@ import { DEFAULT_REGISTRY_URL } from "@powerhousedao/config";
...
+
+By default it only registers the package in powerhouse.config.json with
+provider "registry" — Connect will load it from the registry CDN at runtime.
+
+With --local, the package is also installed into node_modules and marked
+as provider "local" — it will be bundled into ph connect build so the
+preview works without the registry being reachable.
+
+Resolution order for the registry URL:
+  --registry flag > powerhouse.config.json > PH_REGISTRY_URL env > default
```
acaldas: Same as above

acaldas (commented):
Looks good, please look into the failing CI job.

clis/ph-cli/src/commands/install.ts
gpuente: added fix in [8463ff1](https://github.com/powerhouse-inc/powerhouse/pull/2476/commits/8463ff18d3234f6c59e685630279ee2049e8975b)

apps/academy/docs/academy/04-APIReferences/00-PowerhouseCLI.md
gpuente: added fix in [8463ff1](https://github.com/powerhouse-inc/powerhouse/pull/2476/commits/8463ff18d3234f6c59e685630279ee2049e8975b)

PR #2489, opened 2026-04-21
feat: tree-shaking improvements for reactor-browser and shared packages

acaldas:
## Summary

- **Split analytics into store + hooks subpath**: Moved `createAnalyticsStore`/`getAnalyticsStore` to `store.ts` with no React/tanstack dependency. The root `reactor-browser` index now exports only the store, eliminating `@tanstack/react-query` from the root bundle. React hooks remain available via `@powerhousedao/reactor-browser/analytics`.
- **Update analytics imports to subpath**: Updated `drive-analytics` consumers to import from `@powerhousedao/reactor-browser/analytics`.
- **Isolate `generateMock` to prevent zocker bundle leakage**: Moved `generateMock` (and its `zocker`/`faker` dependency) to its own `mock.ts` module so it becomes a separate chunk and doesn't contaminate shared bundle chunks.
- **Lazy-load jszip**: `files.ts` had a static `import JSZip from "jszip"` causing ~118KB to load eagerly with every document model package. Switched to a dynamic import inside the async functions that use it.
- **Lazy-load pglite in analytics-engine**: Deferred pglite import in `BrowserAnalyticsStore` creation.
- **Add `sideEffects: false` to codegen boilerplate**: Generated packages now declare themselves fully tree-shakeable.

## Test plan

- [ ] `pnpm build` passes in `reactor-browser`, `shared`, `analytics-engine`
- [ ] Analytics hooks still work via `@powerhousedao/reactor-browser/analytics`
- [ ] File export/import (jszip) still works when triggered
- [ ] `generateMock` still importable from `@powerhousedao/shared/document-model`


ryanwolhuter (approved):
looks good to me

PR #2490, opened 2026-04-21
feat: preserve dist-tags end-to-end and streamline package manager UX

gpuente:
## PR Description

Makes dist-tag handling first-class across the package install flow — from `ph install` down to Connect's package manager UI — and tightens the available-packages listing.

- **Preserve dist-tags end-to-end:**  
  `ph install name@dev` and `powerhouse.config.json` entries now keep the tag string (`dev`, `staging`, …) instead of being silently resolved to a pinned version.  
  `builder-tools` emits `name@version` into `ph-packages.json` so CDN fetches honor the requested release stream.

- **`name@tag` / `name@version` in the search input:**  
  Connect's Package Manager search parses suffixes like `@dev` or `@1.2.3` and installs exactly what the user typed.

- **Scoped registry listing + npm fallback:**  
  The "Available packages" list now shows only packages published to the custom registry (no more npm-uplink cache noise), with a single "Install from npm" card appearing when the typed name isn't local.


acaldas (commented):
Looks good to me but the E2E test is failing

PR #2491, opened 2026-04-21
fix(shared): rewrite bundled React chunks to external ESM imports

gpuente:
### Description

Fixes an issue where React/React-DOM was being unintentionally bundled
into build outputs, creating multiple React instances at runtime.

This happened when a dependency used `require("react")`. The bundler
would inline React into the output, for example:

``` js
// generated chunk (before)
import { t as require_react } from "./react-WlOzh90t.js";

const React = require_react(); // ← bundled copy
```

or even inline inside a larger file:

``` js
//#region node_modules/react/cjs/react.production.js
var require_react_production = __commonJSMin(...);
//#endregion
```

At the same time, Connect loads React from an external ESM source
(`esm.sh`). This meant the app ended up with **two different React
instances**, which caused runtime errors (e.g. hooks failing or version
mismatches).

Existing tooling (`esmExternalRequirePlugin`) only rewrote simple
`require("react")` cases and missed these bundled patterns, so the issue
remained.

This PR adds a post-build rewrite step that:

-   Removes bundled React chunks/regions
-   Rewrites imports to use the external React instead

Example after the fix:

``` js
// rewritten output
import * as __ext0 from "react";
var require_react = () => __ext0;

const React = require_react(); // ← now uses external React
```

This ensures a single React instance is used at runtime and prevents
those inconsistencies.

### DIFF
--- a/packages/shared/clis/build-config.ts
+++ b/packages/shared/clis/build-config.ts
@@ -1,5 +1,6 @@
 import { esmExternalRequirePlugin } from "rolldown/plugins";
 import type { InlineConfig } from "tsdown";
+import { rewriteBundledReactToExternalPlugin } from "./rewrite-bundled-react-plugin.js";
 
 const entry = [
   "index.ts",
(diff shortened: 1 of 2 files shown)


acaldas:
This fix looks scary 🐈‍⬛ 
How can the issue be reproduced?
It's happening with the builder profile package? What dependency is causing it?

gpuente:
this PR is not longer needed

PR #2492, opened 2026-04-22
feat: update migrate command

ryanwolhuter:
Implement the updated migrate command.

In my opinion this handles around 80% of migration issues. I have explicitly not tried to handle every edge case, since this would be as likely to break things as to fix them.

To achieve this, we need to run several generate commands for several module types. Even though we were awaiting these promises, race conditions still arose because ts-morph uses async fs operations which don't always resolve as we might expect.

This was resolved by passing a single instance of ts-morph's `Project` class to all of the functions, who then call its methods. This let's ts-morph and the typescript runtime handle avoiding race conditions, and also makes codegen significantly more efficient (especially in terms of memory use).

I have also added equivalents of the existing `generateAll` codegen function, which only worked on document models, for each of the module types for a given project. There is now a new `generateAll` function which runs each of these functions. The updated `migrate` command calls this `generateAll` function after doing some attempts at fixing problems.

I also noticed that when run via the cli, the codegen functions were not updating the manifest file. This meant that running these would not work to update the manifest when doing the migration. I have instead moved the job of updating the file to the codegen functions themselves, which means that this works for both the cli and vetra. I have moved the types for the `Manifest` to instead be derived from zod schemas so that these can also be used to validate the json code.


packages/codegen/src/name-builders/get-variable-names.ts
```diff
@@ -73,10 +70,10 @@ export function getLatestDocumentModelSpecVersionNumber(
   return getLatestDocumentModelSpec(documentModelState).version;
 }
 
-export function getDocumentModelVariableNames(documentType: string) {
```
ryanwolhuter, line 76: I think this is actually a mistake that I've only noticed now. The parameter for this function that we use for creating all of the names we use when generating a document model says "documentType", but the actual value we pass is the document model state name field. I don't know if it would even work if you passed in something like "@namespace/something". it just assumes its a normal string. But the problem is that then the name field is subject to change and is not unique, so I'm not sure what we should do about it.

packages/shared/clis/constants.ts
```diff
@@ -62,6 +62,7 @@ export const VERSIONED_DEPENDENCIES = [
   "@powerhousedao/reactor-api",
   "@powerhousedao/reactor-browser",
   "@powerhousedao/connect",
+  "@powerhousedao/shared",
```
ryanwolhuter, line 65: I didn't want to include this, but I see that there is subgraph generated code which uses a function `addFile` which only lives in /shared. We should re-export it.

clis/ph-cli/src/services/generate.ts
```diff
@@ -32,20 +31,15 @@ export async function startGenerate(options: GenerateArgs) {
...
-    useVersioning: useVersioningFlag,
     subgraphName,
-    migrateLegacy,
   } = options;
+  const project = buildTsMorphProject(projectDir);
 
   const documentModelFile =
     documentModelFilePositional ?? documentModelFileOption;
 
-  const useVersioning = useVersioningFlag || migrateLegacy;
```
ryanwolhuter, line 48: We no longer need to migrate legacy in the generate command. This is handled by the migrate command, and should not be needed after that

clis/ph-cli/src/services/generate.ts
```diff
@@ -65,48 +59,47 @@ export async function startGenerate(options: GenerateArgs) {
...
-      path: filePath,
-      config,
-      options,
-      useVersioning,
-      migrateLegacy,
-    });
-  } else {
-    await generateCode(config, useVersioning, migrateLegacy);
+    await generateFromFile(filePath, project);
   }
```
ryanwolhuter, line 101: saving the project at the end is much faster and also makes race conditions less likely

packages/codegen/src/templates/subgraphs/document-model-schema.ts
```diff
@@ -61,15 +63,3 @@ ${moduleSchemas}
 `;
   return body;
 };
-
-function pascal(name: string): string {
```
ryanwolhuter, line 65: we should really try to avoid letting agents put stuff like this in the code, I've seen this same implementation done in-place in 3 different files. It's also not consistent with the `change-case` library that we use everywhere else

packages/codegen/src/templates/subgraphs/document-model-schema.ts
```diff
@@ -1,29 +1,31 @@
-export type DocumentModelSubgraphSchemaParams = {
-  pascalCaseDocumentType: string;
-  modules: Array<{
+import { camelCase, pascalCase } from "change-case";
+import { filter, flatMap, isTruthy, join, map, pipe, prop } from "remeda";
+
+export const documentModelSubgraphSchemaTemplate = (v: {
```
acaldas: This template is not needed anymore, as the document model subgraphs get generated at runtime.

PR #2505, opened 2026-04-23
feat: separate generate commands

ryanwolhuter:
Improve the `ph generate` cli command so that it is separated into subcommands, and leverage the new functions which allow re-generating any/all existing modules in a project.

Previously the generate command was a single command, and it required many complicated arguments to achieve a specific result.

The generate command is now used with these subcommands:

```
ph-cli generate <subcommand>
> The generate command creates code from document models. It helps you build editors, processors, and other components based on your document model files.

where <subcommand> can be one of:

- all - Re-generate all modules in the current project
- document-model - Generate a document model [alias: doc]
- editor - Generate a document editor
- app - Generate a drive app
- processor - Generate a processor
- subgraph - Generate a subgraph
- migration-file - Generate a migration file

For more help, try running `ph-cli generate <subcommand> --help`
```

Note that there is now an "all" subcommand, which runs generate for all of the existing generated code in the current project.

I have also as far as possible made it so that the arguments for each of the subcommands are as similar as possible for each.

```
> Generate a document model

OPTIONS:
  --file, -f <file> - Path to the file to generate the document model from [optional]
  --dir, -d <dir>   - Name of the directory of an existing document model to re-generate [optional]

FLAGS:
  --all, -a  - Re-generate all existing document models in the current project [optional]
  --debug    - Log arguments passed to this command [optional]
  --help, -h - show help [optional]
  ```
  
  In general, for cases where code is generated from a file, you pass a `--file` argument with the path.
  
  For convenience, I have also added a `--dir` argument, which is the name or path to the directory where an existing module to be regenerated is found.
  
  Each of these sub-commands also has a `--all` flag, which runs generate on all of the existing generated modules of the type specified by the subcommand.
  
  In the cases where a name is required to generate a new module, like with an editor for example, there is now a `--name` argument. Previously you had to use `--editor-name` for an editor, `--app-name` for an app, and so on. Since we now already know what we are generating from the subcommand, we can use use the simple `--name` arg for all of them.
  
  ```
  ph-cli generate editor
> Generate a document editor

OPTIONS:
  --name, -n <str>          - The name of the document editor to generate [optional]
  --document-type, -t <str> - The document type for the new editor [optional]
  --dir, -d <dir>           - Name of the directory of an existing editor to re-generate [optional]

FLAGS:
  --all, -a  - Re-generate all existing editors in the current project [optional]
  --debug    - Log arguments passed to this command [optional]
  --help, -h - show help [optional]
```

```
ph-cli generate app
> Generate a drive app

OPTIONS:
  --name, -n <str>                 - The name of the drive app to generate [optional]
  --document-types <str>, -t=<str> - The document types allowed by the new app [[...optional]]
  --dir, -d <dir>                  - Name of the directory of an existing app to re-generate [optional]

FLAGS:
  --disable-drag-and-drop - Do not allow drag and drop in this drive app. [default: false]
  --all, -a               - Re-generate all existing apps in the current project [optional]
  --debug                 - Log arguments passed to this command [optional]
  --help, -h              - show help [optional]
```

Finally, you will also see that the `--help` command is significantly improved. It can be run at the top level with `ph generate --help` to get an overview of the subcommands, and then also for each of the subcommands like `ph generate document-model --help` to get help data specific to each of the subcommands.


apps/academy/docs/academy/04-APIReferences/00-PowerhouseCLI.md
```diff
@@ -224,86 +231,152 @@ Use your local `powerhouse` monorepo dependencies the current project.
 
 
 ## Generate
+The generate command creates code from document models. It helps you build editors, processors, and other components based on your document model files.
```
acaldas: I'd remove the "creates code from document models". As that is only true for the document model spec.
It creates the boilerplate of each module

acaldas (approved):

PR #2532, opened 2026-04-30
feat: connect config json

gpuente:
# Connect runtime config: `powerhouse.config.json`

Replaces the legacy `dist/ph-packages.json` runtime artifact with a
versioned, structured `powerhouse.config.json` that the SPA fetches on
boot.

Operators can edit this file in the Docker image (or post-build `dist`)
to customize Connect at runtime—packages, branding, default drives—
without rebuilding the SPA.

Ships with a sibling JSON Schema (`powerhouse.config.schema.json`) so
any schema-aware editor (VS Code, JetBrains) gives autocomplete and
validation when operators hand-edit the file.

## What's included

- New runtime artifact `dist/powerhouse.config.json` with:
  - `schemaVersion: 2`
  - Structured `packages` (`{ packageName, version?, provider?, url? }`)
  - A `connect` customization section
- `connect.branding.appName` and `connect.branding.homeBackground` for
  app identity and the home-screen hero
- `connect.drives.allowAddDrive` and
  `connect.drives.defaultDrives` for:
  - Default-drive auto-connect
  - Locked-down deployments
- Sibling `powerhouse.config.schema.json` emitted by the build,
  referenced via `$schema` in the runtime file

## Sample `powerhouse.config.json`

```json
{
  "$schema": "./powerhouse.config.schema.json",
  "schemaVersion": 2,
  "packages": [
    {
      "packageName": "@powerhousedao/versioned-documents",
      "version": "5.1.0-dev.26",
      "provider": "registry"
    }
  ],
  "localPackage": null,
  "connect": {
    "branding": {
      "appName": "Acme Connect",
      "homeBackground": {
        "avif": "/branding/home.avif",
        "png": "/branding/home.png"
      }
    },
    "drives": {
      "allowAddDrive": true,
      "defaultDrives": [
        {
          "url": "https://switchboard.example.com/d/acme-main",
          "name": "Acme Main",
          "icon": "/branding/main-icon.png"
        },
        {
          "url": "https://switchboard.example.com/d/acme-archive"
        }
      ]
    }
  }
}
```

### Packages

Each entry needs `packageName`.

- `version` pins it
- `provider` picks the source:
  - `"registry"`
  - `"local"`
  - `"npm"`
  - `"github"`

```json
{
  "packageName": "@scope/pkg",
  "version": "1.2.0",
  "provider": "registry"
}
```

### Branding

- `appName` sets the tab title
- `homeBackground` sets the home-screen hero
  - AVIF preferred
  - PNG fallback

```json
{
  "branding": {
    "appName": "Acme Connect"
  }
}
```

### Default drives

Auto-connect on first load.

Only `url` is required.

- `name` is optional
- `icon` is optional

Set `allowAddDrive: false` to hide the add-drive UI.

```json
{
  "drives": {
    "allowAddDrive": false,
    "defaultDrives": [
      {
        "url": "https://switchboard.example.com/d/main",
        "name": "Main"
      }
    ]
  }
}
```

### Editor support

Keep the `$schema` line.

Opening the file in VS Code gives:

- Inline documentation
- Validation
- Autocomplete


apps/connect/src/store/reactor.ts
```diff
@@ -109,16 +112,20 @@ export async function createReactor(localPackage?: DocumentModelLib) {
...
     .build();
 
-  // load packages list from ph-packages.json (replaceable post-build)
-  const packagesConfig = await loadPackagesConfig();
+  // load runtime config from powerhouse.config.json (replaceable post-build)
+  const runtimeConfig = await loadRuntimeConfig();
+
+  // Apply branding from runtime config
+  const appName = runtimeConfig.connect?.branding?.appName;
+  if (appName) document.title = appName;
```
acaldas: Let's move this out of this file, it already does too much.

docker/nginx.conf.template
```diff
@@ -61,6 +61,16 @@ http {
             add_header Expires "0";
         }
 
+        # Operator-mutable runtime config: never cache so post-deploy edits
+        # are visible on the next page load.
+        location = ${PH_CONNECT_BASE_PATH}/powerhouse.config.json {
+            alias /var/www/html/project/powerhouse.config.json;
+            access_log off;
+            log_not_found off;
+            etag off;
```
acaldas: I'm not sure about this but doesn't etag allow caching based on the file contents? If the config file is unchanged it would be good to apply caching so that Connect startup doesn't need to wait for the request

packages/builder-tools/connect-utils/vite-plugins/ph-config.test.ts
```diff
@@ -0,0 +1,275 @@
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
```
acaldas, line 1: Please make sure the test files and vitest don't get added to the `builder-tools` bundle.

packages/builder-tools/connect-utils/runtime-config-schema.ts
```diff
@@ -0,0 +1,149 @@
...
+
+export const RUNTIME_CONFIG_SCHEMA_FILENAME = "powerhouse.config.schema.json";
+export const RUNTIME_CONFIG_SCHEMA_REF = `./${RUNTIME_CONFIG_SCHEMA_FILENAME}`;
+
+export const runtimeConfigSchema = {
+  $schema: "http://json-schema.org/draft-07/schema#",
+  $id: "https://powerhouse.inc/schemas/powerhouse.config.json",
+  title: "Powerhouse Connect runtime configuration",
+  description:
+    "Runtime configuration loaded by Connect at boot from /powerhouse.config.json. Operator-mutable in Docker; emitted by ph connect build.",
```
acaldas: `Operator-mutable in Docker` No need for this reference

packages/builder-tools/connect-utils/runtime-config-schema.ts
```diff
@@ -0,0 +1,149 @@
+// JSON Schema (draft-07) for dist/powerhouse.config.json. Hand-written to
```
acaldas: How does the schema handle the `powerhouse.config.json` fields already in use? 

https://github.com/powerhouse-inc/powerhouse/blob/a16dd97eae8d12f65a7ae7eeb15c6860102b8f29/packages/shared/clis/types.ts#L30-L81

apps/connect/src/runtime-config.ts
```diff
@@ -0,0 +1,200 @@
...
+function assertOptionalString(
+  value: unknown,
+  path: string,
+): asserts value is string | undefined {
+  if (value !== undefined && typeof value !== "string") {
+    throw new Error(`powerhouse.config.json: '${path}' must be a string`);
+  }
+}
+
+function assertRuntimePowerhouseConfig(
```
acaldas, line 22: What is the relationship between https://github.com/powerhouse-inc/powerhouse/blob/main/packages/shared/connect/env-config.ts and this?

We should make it consistent otherwise it's very confusing,

acaldas:
How does the json schema work? Can we point to a URL instead of a relative path? So we don't need to add it to the boilerplate

apps/connect/src/store/reactor.ts
gpuente: done, moved to a new `applyConnectBranding(config)` helper in `runtime-config.ts`, called from `loadComponent` right after `loadRuntimeConfig` resolves. `createReactor` no longer touches the title

docker/nginx.conf.template
gpuente: yeah good catch, `no-store` was overkill... switched to `no-cache` + etag on (default). Operator edits still show up immediately, but unchanged file gets a 304 instead of a full body transfer.

> Note: doesn't speed up startup (round-trip is the same), just saves the bytes when nothing changed

packages/builder-tools/connect-utils/vite-plugins/ph-config.test.ts
gpuente: Already covered:

- `tsdown.config.ts` entry is just `./index.mts`, tests aren't in the import graph
- `package.json` has `"files": ["dist"]`
- vitest is in devDependencies

Verified with `npm pack --dry-run`

packages/builder-tools/connect-utils/runtime-config-schema.ts
gpuente: Removed

packages/builder-tools/connect-utils/runtime-config-schema.ts
gpuente: refactored, now:

- `packages/shared/connect/schema-fragments.ts`: shared shapes (`PowerhousePackage`, `PHConnectRuntimeConfig`) defined once
- `packages/shared/clis/source-config-schema.ts`: full schema for the source `powerhouse.config.json` (new, devs editing the source file now get autocomplete too)
- `runtime-config-schema.ts`: imports the shared fragments, no more duplicated shapes

both schemas reference the same fragments. Runtime stays a strict subset of source by design (server fields like `auth`/`switchboard` are stripped.

Drift test in `source-config-schema.test.ts` keeps the schema in sync with the TS type.

apps/connect/src/runtime-config.ts
gpuente: yep, real overlap, coming in a follow-up commit:
`env vars → powerhouse.config.json → SPA reads`. Env vars become first-time seed values for the config file (set if absent only, never override). Operator edits to the file win and survive restarts. SPA reads only from the file. True build-time things (Sentry, GA, basePath) stay env-vars-only

gpuente:
Hit a couple of bugs while testing the env seeding flow in docker, fixed both here since they're small and in the same area.

Entrypoint now seeds `powerhouse.config.json` from env vars at container start, same set-if-absent rules as the Vite plugin, so operator edits to the file always win.

Also swapped `mv tmp file` for `cat tmp > file`. `mv` replaces the inode, which crashes with `Resource busy` when the config file is bind-mounted as a single file (K8s ConfigMap pattern). Without the swap, anyone overriding via a ConfigMap would get a crashlooping pod.

The nginx change is a pre-existing bug. `location ${BASE}/foo` rendered as `//foo` at the default base path, so requests never matched and fell through to the catch-all, losing the custom headers. Including the new `no-cache`, which means until now that fix wasn't actually doing anything at the default deploy. Fixed via a separate `${PH_CONNECT_BASE_PREFIX}` (trailing slash stripped) for the prefixed locations. Catch-all unchanged so routing is identical.

Tested with `docker build` + `docker run` at both `/` and `/connect`: baseline, env seeding, file wins, ETag → 304, bare-subpath URLs, operator edit surviving a restart on a bind-mounted file. All pass.

apps/connect/src/runtime-config.ts
acaldas: env vars having less precedence than the config.json is a bit counterintuitive, but it's only for runtime vars.
If it's easy to do, let's log a warning if an env var is overriden by a value already set in the config.json.

acaldas:
@gpuente @ryanwolhuter How does the JSON schema work? Can we host it somewhere and use an URL in `powerhouse.config.json`?

acaldas:
@gpuente Can you ensure all the Connect configuration options (env vars + config json) are documented in the Academy?
Split by buildtime and runtime.

ryanwolhuter:
> @gpuente @ryanwolhuter How does the JSON schema work? Can we host it somewhere and use an URL in `powerhouse.config.json`?

afaik any url for a source of a schema.json file works, looks like the current is pointing to a github url? I've seen people just use a github blob url too.

packages/shared/connect/env-to-runtime-config.ts
```diff
@@ -0,0 +1,130 @@
...
+    parse: parseDefaultDrivesUrl,
+  },
+];
+
+/**
+ * Returns true when the dotted path resolves to a defined value inside obj.
+ * `null` counts as defined; `undefined` does not (matches `//=` semantics
+ * we use in the Docker entrypoint via jq).
+ */
+function hasPath(obj: unknown, dotted: string): boolean {
```
ryanwolhuter, line 64: Please can we rather use existing utility libraries instead of adding implementations like this all over the place? it's much harder to maintain and its more than likely that there are edge cases that this does not consider.

https://remedajs.com/docs/#stringToPath

ryanwolhuter:
> @gpuente Can you ensure all the Connect configuration options (env vars + config json) are documented in the Academy? Split by buildtime and runtime.

The last time we discussed this precedence we said it should be:

4th - default config
3rd - config file to override default config
2nd - env vars to allow overriding config file
1st - command line args to allow overriding env vars

To me this is the logical conclusion, since otherwise any of the following ones would be redundant.

IMO there is way too much reversal of logic and overriding of values here. In what context is this even supposed to be used? Why don't we just update the config file directly when the user does stuff like add or remove packages, which is what we already do for example with the manifest?

If this config is truly intended to be runtime only, then surely that means it doesn't need the data to be persisted other than perhaps in local storage? If it needs to be persisted, then doesn't that mean it isn't actually "runtime" at all?

acaldas:
@gpuente Let's add a build time env var to configure the URL to get the config from.
So we can configure it to get the config from a Switchboard subgraph for example.

froid1911:
This entire file thing does not fit into the vetra cloud architecture. Maybe you can do the following (as alternative)

- PH_CONNECT_CONFIG_SWITCHBOARD which points to https://switchboard.staging.vetra.io 
- https://switchboard.staging.vetra.io offers a subgraph query to return env vars for connect which accepts hostname / env name or vetra cloud project name as input

PR #2556, opened 2026-05-04
feat: use generated get document with operations query in switchboard link

### LINKED ISSUE #2516: Switchboard button doesn't give a correct query in graphQL
The wrong query that doesn't return relevant data. 
Image

The query i had to figure out myself 

Image

Also when i hit Switchboard button in localhost Vetra Preview 

Image

We probably want something like this?

Image

query Query($identifier: String!) {
  document(identifier: $identifier) {
    document {
      documentType
      name
      operations {
        items {
          action {
            input
          }
        }
      }
    }
  }
}

ryanwolhuter:
fixes #2516 

Currently we have been using a hardcoded string when generating the query we use to populate the explorer when the user clicks the switchboard button in connect.

This string is long out of date and does not work anymore.

Since we already generate the `GetDocumentWithOperationsDocument` in reactor browser, which we then use elsewhere, we can simply use the string value of the body of this query in this function instead. This will prevent future drift also.

The variables created were for `documentId`, which is no longer the correct name, should now be `identifier`, and `driveId`, which the explorer says is not a valid variable. I have updated the former and removed the latter.

### DIFF
--- a/packages/reactor-browser/src/utils/switchboard.ts
+++ b/packages/reactor-browser/src/utils/switchboard.ts
@@ -1,4 +1,5 @@
 import * as lzString from "lz-string";
+import { GetDocumentWithOperationsDocument } from "../graphql/gen/schema.js";
 
 export async function getDriveIdBySlug(driveUrl: string, slug: string) {
   if (!driveUrl) {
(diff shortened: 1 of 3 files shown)


CallmeT-ty:
Sounds like a better solution to me.

PR #2586, opened 2026-05-08
feat: add download button

ryanwolhuter:
Consolidate the various "export document" logic in one place and rename the logic to "download document".

Add a new `useDownloadDocument` hook which uses the underlying `downloadDocument` function, but accepts an id instead of a whole document. This lets the hook be used in places where we only know a node id.

Use this logic in the existing places like the document toolbar.
Add a "download" option to the file dropdown menu.


apps/connect/src/components/document-editor-container.tsx
```diff
@@ -1,39 +1,15 @@
...
   useSelectedDocument,
-  validateDocument,
 } from "@powerhousedao/reactor-browser";
 import { useCallback, useMemo } from "react";
 
 export function DocumentEditorContainer() {
   const [selectedDocument] = useSelectedDocument();
   const parentFolder = useNodeParentFolderById(selectedDocument.header.id);
 
-  const onExport = useCallback(() => {
```
ryanwolhuter, line 17: I removed these things that were being passed to the DocumentEditor component as props because the actual component was just ignoring them.

acaldas (approved):

PR #2593, opened 2026-05-11
fix(release): resolve release tag by channel, not by HEAD

acaldas:
## Summary
The `publish-ph-binaries` job from #2589 has been silently skipping binary uploads on every release since it merged — that's why `v6.0.0-dev.238` and `v6.0.0-dev.239` have empty asset lists, and why `scripts/install-binary.dev.sh` failed with `release 'v6.0.0-dev.239' has no asset named 'ph-darwin-arm64'`.

**Root cause:** nx-release's GitHub-release-creation step (`createRelease: "github"` in `releases/release.ts`) writes a **lightweight** tag against the branch tip at the moment it runs — which is the commit *before* `release.ts`'s `chore(release)` version-bump commit. So when `publish-ph-binaries` checked out the post-bump HEAD on main, `git tag --sort=-v:refname --points-at HEAD` matched nothing and the job exited with `::notice::No release tag at HEAD`. Confirmed via `git cat-file -t v6.0.0-dev.{237,238,239}` — all three report `commit` (lightweight), not `tag` (annotated), proving nx (not release.ts) is what writes the remote tag.

**Fix:** pass the just-published tag from the `release` job to `publish-ph-binaries` as a job output, instead of having the downstream job guess.

```yaml
release:
  outputs:
    tag: ${{ steps.capture-tag.outputs.tag }}
  steps:
    - id: pre-tag
      run: echo "tag=$(git tag --list 'v*' --sort=-v:refname | head -1)" >> "$GITHUB_OUTPUT"
    - name: Run release
      run: pnpm run run-release -- ...
    - id: capture-tag
      run: |
        git fetch --tags --force
        NEW=$(git tag --list 'v*' --sort=-v:refname | head -1)
        [ -n "$NEW" ] && [ "$NEW" != "$OLD" ] && echo "tag=$NEW" >> "$GITHUB_OUTPUT"

publish-ph-binaries:
  needs: release
  if: ${{ needs.release.outputs.tag != '' }}
  steps:
    - uses: actions/checkout@v6
      with:
        ref: ${{ needs.release.outputs.tag }}
```

This guards against three classes of failure that the previous regex-by-newest approach could hit:

| failure mode | regex-by-newest | output-passing |
|---|---|---|
| release no-op (no version-impacting changes) | picks the *previous* release tag → `--clobber` overwrites that release's assets | empty output → skip |
| two concurrent release runs (concurrency-group failure / manual dispatch) | each run picks whichever tag was pushed last | each run's output is its own tag |
| locally-created tag matching a channel pattern | regex matches it accidentally | only the tag actually produced this run is used |

Also removes the channel-aware regex (`main` → `-dev\.`, `release/staging/*` → `-staging\.`, etc.) since `release` is the source of truth for which tag belongs to which run.

**Subtle but important:** the tag the release job captures points to the commit *before* the `chore(release)` version-bump commit (nx-release's lightweight-tag quirk). So when `publish-ph-binaries` checks out `${{ needs.release.outputs.tag }}`, the `package.json` at that ref is one version behind. That's fine because `CLI_VERSION` is injected at compile time via `--define`, so the binary self-reports the tag's version, not whatever's in package.json. This is documented in a comment in the workflow so future-you doesn't try to "fix" it.

## Test plan
- [ ] Merge and let the next scheduled (or dispatched) release run on `main`; verify the new release has `ph-darwin-arm64`, `ph-darwin-x64`, `ph-linux-arm64`, `ph-linux-x64`, `ph-windows-x64.exe`, `checksums.txt` attached.
- [ ] `curl -fsSL https://raw.githubusercontent.com/powerhouse-inc/powerhouse/main/scripts/install-binary.dev.sh | bash` resolves `dev → v6.0.0-dev.<latest>` and installs a working binary.
- [ ] Trigger a release with `--skip-publish` or against a branch with no changes; verify `publish-ph-binaries` is *not* run (the `outputs.tag` is empty, so `if:` evaluates false).

## Notes
- `v6.0.0-dev.239`'s assets have already been backfilled manually (cross-compiled locally, sha256-verified, `gh release upload`'d). The current dev install line works today; this PR ensures the next release works automatically.
- The upstream lightweight-vs-annotated tag mismatch in `releases/release.ts` is a separate concern (release.ts wants to push annotated tags, but nx beats it to it with a lightweight one). Not addressed here.

### DIFF
--- a/.github/workflows/release-branch.yml
+++ b/.github/workflows/release-branch.yml
@@ -57,6 +57,12 @@ jobs:
     name: Release packages in monorepo
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    outputs:
+      # Tag the release step just created (release.ts writes this directly to
+      # $GITHUB_OUTPUT). Empty when no version bump happened, when --skip-git-tag
+      # is set, or when release.ts exits early — downstream jobs (notably
+      # publish-ph-binaries) gate on this so they only run for real releases.
+      tag: ${{ steps.run-release.outputs.tag }}
     steps:
       - name: Log inputs
         run: |
@@ -98,7 +104,12 @@ jobs:
           echo "//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}" >> ~/.npmrc
           echo "registry=https://registry.npmjs.org/" >> ~/.npmrc
 
+      # release.ts appends `tag=v<workspaceVersion>` to $GITHUB_OUTPUT when it
+      # actually creates a tag (i.e. not a dry-run, not --skip-git-tag, and a
+      # real version bump occurred). That output flows through this step into
+      # the job's `outputs.tag` for downstream consumers.
       - name: Run release
+        id: run-release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_ACCESS_TOKEN }}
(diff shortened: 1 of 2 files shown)


acaldas:
Thanks for the review. Force-pushed `79a876cee` with substantive responses to all three:

**1 + 2: drop the git-tag-sort heuristic entirely; have `release.ts` emit the tag directly.**
Both review comments stem from the same brittleness — using `git tag --sort=-v:refname | head -1` is wrong because (a) stable `vX.Y.Z` sorts above `vX.Y.Z-dev.N` per SemVer precedence (the snapshot/post comparison would silently fail every time `v6.0.2-staging.6` exists alongside a new `v6.0.0-dev.N`), and (b) `git fetch --tags --force` clobbers any local annotated tag, which would in turn break the Sentry sourcemap upload that also does `git tag --points-at HEAD`.

Fix: `releases/release.ts` now appends `tag=v${workspaceVersion}` to `$GITHUB_OUTPUT` after a successful release (guarded by `gitTag && workspaceVersion && process.env.GITHUB_OUTPUT`). The workflow gives `Run release` an `id`, maps the job's `outputs.tag` to that step's output, and removes the snapshot/capture-tag steps entirely. No git tag listing, no fetch, no force.

Bonus: the Sentry "Upload source maps" step had the same `git tag --points-at HEAD` assumption and was silently skipping for the same reason (lightweight tag pre-bump → never matches the post-bump HEAD). Switched it to consume `steps.run-release.outputs.tag` so source maps actually upload again.

**3: soften the checkout comment to acknowledge the conditional case.**
You're right that when `skip_changelog` is set (or nx's `createRelease: "github"` is otherwise bypassed), `release.ts`'s annotated tag IS the only tag and it points at the post-bump commit. Reworded to say the ref "may point at either the pre-bump commit or the chore(release) commit itself" and to call out that we don't actually depend on which one — `CLI_VERSION` is injected via `--define` regardless.

Diff summary:
- `releases/release.ts`: +9 lines (import + emit block)
- `release-branch.yml`: −33 net (the snapshot + capture-tag + git-fetch-force machinery is gone; one Sentry step rewritten)

PR #2596, opened 2026-05-11
feat(ph-lora): Claude Code doc-review skill suite + release-notes generator

CallmeT-ty:
## Summary

- Adds a suite of Claude Code slash commands under `.claude/commands/` for AI-assisted academy documentation review
- Adds `/release-notes` skill for generating human-readable release notes from CHANGELOG.md
- Adds gap reports, mapping, and strategy docs under `test/ph-lora/`
- Fills in the `v6.0.0` RELEASE-NOTES.md entry (236 dev versions, Jan–May 2026)

## Skills added

| Command | What it does |
|---|---|
| `/doc-review <section-id>` | Reviews one academy doc section for drift against monorepo source — exports a structured gap report |
| `/doc-status` | Shows coverage across all 15 mapped sections (reviewed / clean / findings open) |
| `/doc-clarity <section-id>` | Rewrites a doc section for clarity after mechanical drift is fixed |
| `/doc-fix <section-id>` | Applies fixes from a gap report back to the academy markdown |
| `/release-notes [from] [to] [--write]` | Generates human-readable release notes from CHANGELOG.md entries |

## Test plan

- [ ] Run `/doc-status` — should show coverage table for all 15 mapped sections
- [ ] Run `/doc-review get-started` — should produce a structured gap report
- [ ] Run `/release-notes` — should generate notes for the latest CHANGELOG.md version
- [ ] Run `/release-notes 6.0.0-dev.230 6.0.0-dev.239` — should span the version range correctly
- [ ] Verify gap reports under `test/ph-lora/gap-reports/` match the sections they cover
- [ ] Ask devs to validate gap report findings against the actual source packages


.claude/settings.local.json
```diff
@@ -37,6 +37,19 @@
       "Bash(git stash:*)",
       "mcp__hotseat",
       "Bash(npm run *)",
+      "Bash(sed 's/.*\\\\\\(```[^ ]*\\\\\\).*/\\\\1/')",
```
acaldas: These rules should not be committed.

test/ph-lora/ph-lora-mapping.md
```diff
@@ -0,0 +1,151 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$comment": "Maps academy doc sections to the monorepo packages they are responsible for documenting. Used by PH Laura to scope code↔doc consistency checks. Update this file when new packages are added or doc ownership changes.",
```
gpuente, line 3: typo: PH **Laura** (should be "Lora" no?)

.claude/commands/doc-clarity.md
```diff
@@ -0,0 +1,215 @@
+# /doc-clarity
+
+Review one academy documentation section for prose quality, clarity, and appropriate scope. Does not check code correctness — use `/doc-review` for that. Does not edit files — outputs suggestions for a human or doc-writer agent to act on.
+
+**Usage:** `/doc-clarity <section-id>`
+
+**Best run after `/doc-fix`** — reviewing prose on a section with known mechanical gaps is waste.
+
+**Section IDs** (from `test/ph-lora/ph-lora-mapping.json`):
```
gpuente, line 9: Is there a real reason why we need this .json mappings? maybe I'm missing something, but I think the agent working with this command can figure it out the section if you pass the path to the file and the title of the section, this way you save tokens (avoid read an extra file).

test/ph-lora/ph-lora-mapping.json
```diff
@@ -0,0 +1,199 @@
+{
```
gpuente, line 1: I think the purpose of this mappings is to guide the agent into the files to look at... but I see a few problems with this apporach:
1. we have to keep in sync this mappings. doing this manually will be really hard to mantain.
2. **test** directory doesn't seem to be the right place for this metada

I think you can add this info directly into academy or even better, embed this info in the documentation itself:

- Option A, frontmatter on the section's index/intro file (Docusaurus already parses
  frontmatter, so it's free):
```md
  ---
  title: Work With Data
  phLora:
    packages: [packages/reactor, packages/reactor-api, packages/analytics-engine]
    sourceFiles: [packages/reactor/src/processors, ...]
    checkFocus: Processor interface, subgraph API, GraphQL schema shape
  ---
```

- Option B, HTML comment, invisible in the rendered site:

```md
  <!-- ph-lora
  packages: packages/reactor, packages/reactor-api
  sourceFiles: packages/reactor/src/processors
  checkFocus: Processor interface, subgraph API
  -->
```

.claude/commands/doc-clarity.md
```diff
@@ -0,0 +1,215 @@
+# /doc-clarity
+
+Review one academy documentation section for prose quality, clarity, and appropriate scope. Does not check code correctness — use `/doc-review` for that. Does not edit files — outputs suggestions for a human or doc-writer agent to act on.
+
+**Usage:** `/doc-clarity <section-id>`
```
gpuente, line 5: I think passing a section-id adds some extra steps to the usage of the command... I think you can pass the file path and the title of the section... that's way more easy than find the section-id I guess

.claude/commands/release-notes.md
```diff
@@ -0,0 +1,121 @@
...
+
+---
+
+## Steps
+
+### 1. Parse arguments
+
+From `$ARGUMENTS`:
+
+- If empty or `--latest`: use latest version only (first entry in CHANGELOG.md)
```
gpuente, line 20: maybe add the git history as a source is also a good option, it will be expensive in term of tokens, but can lead to better results since it can read the changes directly... maybe just as an alternative to not clear commit messages

test/ph-lora/scripts/check-pr-drift.ts
```diff
@@ -0,0 +1,308 @@
+/**
```
gpuente, line 1: this file should be placed in `<monorepo-root>/scripts`

.github/workflows/ph-lora-tier2.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: pnpm
+
+      - name: Install deps
+        run: pnpm install-ci
+
+      - name: Run drift classifier
+        run: pnpm exec tsx test/ph-lora/scripts/check-pr-drift.ts
```
gpuente, line 88: not sure about this, if you're running this script just to validate if the docs should be updated or not using claude... why not simply fix the docs with claude? I mean, you'll be spending a lot of tokens just to get the changes reviewed against docs and get a boolean result (docs need or not be updated)

test/ph-lora/adr-ph-lora.md
```diff
@@ -0,0 +1,107 @@
+*This document is intended to stir the pot and get feedback from the team through assumptions.*
```
gpuente, line 1: I don't this this should be in **test** folder

test/ph-lora/gap-reports/gap-report-get-started.md
gpuente, line 1: ## Findings

- **Finding 1 — `ph install` / `ph add` missing**: Source claims correct; doc claim wrong — `00-ExploreDemoPackage.mdx:30,34` already uses `ph install`. Review only scanned `.md`, missed `.mdx`.
- **Finding 2 — `ph generate --editor`**: Correct. `05-BuildToDoListEditor.md:24` still says `--editor`; correct form is `ph generate editor`. Not fixed in PR.
- **Finding 3 — `ph generate TodoList.phdm.zip`**: Correct. Actual line is 20. Parent `generate` is a pure subcommand router — correct form is `ph generate document-model --file <path>`. Not fixed in PR.
- **Finding 4 (missed) — Stale `--legacy-peer-deps`**: `01-CreateNewPowerhouseProject.md:23` prescribes the flag; no longer required for ph-cmd ≥ 6.0.0.
- **Finding 5 (missed) — Stale `ph init` terminal output**: `01-CreateNewPowerhouseProject.md:140-144` shows two lines including `The installation is done!`; real output is ~30 lines with step markers ending in `🎉 Successfully created project...`.
- **Finding 6 (missed) — Stale generated project structure**: `02-DefineToDoListDocumentModel.md:158-174` shows a flat `document-models/<name>/` with 5 files in `gen/` and `src/reducers + src/tests`. Real codegen uses a versioned `v<n>/` layout with 11 files in `gen/`, plus `gen/schema/` and `gen/modules/<module>/`. `migrate-legacy.ts` explicitly migrates the doc's layout into `v1/`.

## Verified clean

- **`ph init`**: Confirmed in both `ph-cmd` and `ph-cli` init.ts files.
- **`namePositional`**: Confirmed at `packages/shared/clis/args/init.ts:13`.
- **`ph vetra --watch`**: Confirmed at `packages/shared/clis/args/vetra.ts:40-48`.
- **Package `ph-cmd` + Node 24**: Confirmed in `ph-cmd/package.json` and doc line 23-24.
- **`ph add` alias**: Confirmed at `install.ts:7` and `help.ts:124`.

.claude/commands/doc-clarity.md
CallmeT-ty: So according to Claude the actual mapping would avoid making the agent search for the section and related mono-repo source files & packages but use the mapping to immediately search within the correct files. Hence while the mappings file needs to stay up to date.

test/ph-lora/ph-lora-mapping.json
CallmeT-ty: that's a nice one. Manually maintaining this would be a pain indeed. 
I think we could use one of the scripts to check upon each commit if a package or new code already has documentation and then update the mappings file.

.claude/commands/doc-clarity.md
CallmeT-ty: I could update that indeed. It is kind of a command that the 'documentation maintainer' uses a lot. Maybe not as much regular dev?

test/ph-lora/adr-ph-lora.md
CallmeT-ty: Correct, i'm going to delete this one. It was early exploration. 
Eventually the whole ph-lora thing should probably be somewhere else

PR #2612, opened 2026-05-14
feat: add tailwind eslint plugin

ryanwolhuter:
We used to have a tailwind eslint plugin, but then they failed to update to v4 of tailwind so we removed it.

I have found a [new one](https://www.npmjs.com/package/eslint-plugin-better-tailwindcss) that does what we want.

I did this because when I was trying to find styles and classes to add dark mode variants, there were many places where classes were used wrong.

I have applied the autofixes that the lint tool detected. There were a lot of issues, most of which are stylistic.

There were some more significant issues though, like places where classes which do not correspond to any existing classes custom or otherwise were used. I have added comments to this PR to call those out.


apps/connect/src/components/error-boundary.tsx
```diff
@@ -102,18 +102,18 @@ function DetailedFallback({ error, resetErrorBoundary }: FallbackProps) {
...
-            <pre className="mt-2 max-h-48 overflow-auto rounded bg-gray-100 p-3 text-xs">
+            <pre className="mt-2 max-h-48 overflow-auto rounded-sm bg-gray-100 p-3 text-xs">
               {errorDetails}
             </pre>
           </details>
         )}
         <PowerhouseButton
           type="button"
           onClick={resetErrorBoundary}
-          className="text-md mt-4 px-3 py-1.5 font-medium"
```
ryanwolhuter, line 116: text-md does not exist in tailwind v4

apps/connect/src/components/search-bar.tsx
```diff
@@ -36,7 +36,7 @@ export const SearchBar = () => {
 
   return (
     <ConnectSearchBar
-      className="max-w-searchbar-width m-4 shrink-0 bg-gray-100"
```
ryanwolhuter, line 39: This searchbar-width class must have been removed at some point because its gone now

apps/connect/style.css
```diff
@@ -1,4 +1,5 @@
 @import "tailwindcss";
```
ryanwolhuter, line 1: tailwind was ignoring classes that came from the theme

packages/design-system/src/connect/components/drop-zone/drop-zone.tsx
```diff
@@ -93,13 +93,13 @@ export function DropZone(props: DropZoneProps) {
       {children}
 
       {enable && isDropTarget && (
-        <div className="fixed inset-0 z-[1000] flex min-h-screen w-screen items-center justify-center bg-black/50">
-          <div className="rounded-[24px] bg-white p-6 shadow-[1px_4px_15px_rgba(74,88,115,0.25)]">
-            <div className="relative flex h-[130px] w-[400px] flex-col items-center justify-start overflow-visible rounded-lg border border-dashed border-black px-4 py-6">
-              <div className="text-center text-base leading-5 text-zinc-500">
+        <div className="fixed inset-0 z-1000 flex min-h-screen w-screen items-center justify-center bg-black/50">
+          <div className="rounded-3xl bg-white p-6 shadow-charcoal">
```
ryanwolhuter, line 97: I added this as a class in the theme

packages/design-system/src/connect/components/footer/footer.tsx
```diff
@@ -8,7 +8,7 @@ export const Footer: React.FC<React.HTMLAttributes<HTMLElement>> = ({
     <footer
       {...mergeClassNameProps(
         props,
```
ryanwolhuter, line 10: I fixed these places where arbitrary styles were used with their equivalents from the theme so that we can find them when adding the dark: variants

packages/design-system/src/connect/components/tabs/tabs.tsx
```diff
@@ -22,7 +22,7 @@ export function Tabs({
             const { label, disabled } = child.props as TabContentProps;
             return (
               <Trigger
-                className="data-[state='active']:tab-shadow ata-disabled:cursor-not-allowed data-disabled:text-gray-400 flex h-7 flex-1 items-center justify-center rounded-lg transition duration-300 data-[state='active']:bg-gray-50 data-[state='active']:text-gray-900"
```
ryanwolhuter, line 25: this obviously couldn't have been working because there's a type in "ata-disabled"

packages/design-system/src/powerhouse/components/sidebar/sidebar-panel.tsx
```diff
@@ -43,7 +43,7 @@ export function SidebarPanel({
     <>
       <div
         className={twMerge(
-          "no-scrollbar flex-1 overflow-auto text-gray-900 transition-shadow",
+          "scrollbar-none flex-1 overflow-auto text-gray-900 transition-shadow",
```
ryanwolhuter, line 46: this class also doesn't exist in the tailwind-scrollbar library, so I changed it this one based on the usage context

packages/design-system/src/ui/components/button/button.tsx
```diff
@@ -7,19 +7,18 @@ import { forwardRef } from "react";
...
     "[&_svg]:pointer-events-none [&_svg]:size-4 [&_svg]:shrink-0",
   ),
   {
     variants: {
       variant: {
         default: "",
-        outline:
-          "border border-input bg-background hover:bg-accent hover:text-accent-foreground",
-        ghost: "hover:bg-accent hover:text-accent-foreground",
+        outline: "border",
```
ryanwolhuter, line 20: These are all classes that would have been added by shadcn, which assumes that you have these defined in your theme, which we do not. So these were never working.

packages/design-system/src/ui/components/checkbox/checkbox-base.tsx
```diff
@@ -18,17 +18,17 @@ const CheckboxBase = React.forwardRef<
     ref={ref}
     className={cn(
       // Base styles
-      "peer size-4 shrink-0 rounded",
+      "peer size-4 shrink-0 rounded-sm",
       // Border & Shadow
-      "border-input border shadow-sm shadow-black/[.04]",
```
ryanwolhuter, line 23: same for these, these classes don't exist

packages/design-system/src/ui/components/id-autocomplete/id-autocomplete-list-option.tsx
```diff
@@ -65,7 +65,7 @@ const ReloadButton: React.FC<{
         className={cn(
           "text-gray-500 dark:text-gray-600",
           isLoadingSelectedOption && "animate-spin",
-          isFetchSelectedOptionSync && "animate-in fade-in duration-500",
```
ryanwolhuter, line 68: the animate-in class needs to be animate-fade-in, so this class wasn't doing anything

packages/design-system/src/ui/components/tooltip/tooltip.tsx
```diff
@@ -35,8 +35,8 @@ const TooltipContent = forwardRef<
         // Padding & Shadow
         "px-3 py-1.5 shadow-md",
         // Animations
-        "data-[state=open]:animate-fade-in data-[state=open]:animate-zoom-in",
```
ryanwolhuter, line 38: this was applying two conflicting animation classes, so only the last defined one was actually being applied

packages/design-system/color-map.txt
```diff
@@ -0,0 +1,2 @@
+text-black -> dark:text-slate-50
```
ryanwolhuter, line 1: I'm using this file to keep track of what colors we choose for each light mode color in dark mode

packages/design-system/style.css
```diff
@@ -5,6 +5,11 @@
 @import "./theme.css";
 @plugin "tailwind-scrollbar";
 
+@plugin 'tailwind-scrollbar' {
```
ryanwolhuter, line 8: most of the scrollbar classes we were using from this library weren't actually being applied because if you don't add this config they don't get imported

scripts/find-arbirary-tailwind.sh
```diff
@@ -0,0 +1,46 @@
+rg -n \
```
ryanwolhuter, line 1: we can use this to find arbitrary tailwind classes to replace them

scripts/find-dark-mode-candidates.sh
```diff
@@ -0,0 +1,46 @@
+rg -n \
```
ryanwolhuter, line 1: this script is for finding components that might need dark mode variants added

.prettierrc.json
```diff
@@ -1,4 +1,3 @@
 {
-  "trailingComma": "all",
-  "plugins": ["prettier-plugin-tailwindcss"]
+  "trailingComma": "all"
```
ryanwolhuter, line 2: the eslint plugin replaces this and does much more

acaldas (approved):

PR #2620, opened 2026-05-18
fix: add missing bg and text styles

ryanwolhuter:
We have a lot of components which were relying on fallbacks / html defaults for their bg color and text color. 

Since we know that these fallbacks are white for bg and text-gray-900 for text, I have added these styles explicitly to components which don't have them. From what I can see, it looks exactly the same as before, which is what we want.

With this in place, we can proceed with automatically adding the dark mode variants.

### DIFF
--- a/packages/design-system/src/connect/components/home-screen/home-screen-item.tsx
+++ b/packages/design-system/src/connect/components/home-screen/home-screen-item.tsx
@@ -33,7 +33,7 @@ export const HomeScreenItem = function HomeScreenItem(
         )}
       </div>
       <div className="w-full max-w-full">
-        <h3 className="w-full max-w-full truncate px-2 text-slate-50">
+        <h3 className="w-full max-w-full truncate px-2 text-gray-900">
           {title}
         </h3>
         {description && <p className="text-gray-500">{description}</p>}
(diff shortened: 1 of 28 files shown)


packages/design-system/src/connect/components/home-screen/home-screen-item.tsx
```diff
@@ -33,7 +33,7 @@ export const HomeScreenItem = function HomeScreenItem(
         )}
       </div>
       <div className="w-full max-w-full">
```
ryanwolhuter, line 35: this was a mistake that I accidentally committed — T and I changed it when we were trying out dark mode color options.

packages/design-system/src/connect/components/home-screen/home-screen.tsx
```diff
@@ -15,7 +15,7 @@ export const HomeScreen = function HomeScreen(props: HomeScreenProps) {
         containerClassName,
       )}
     >
-      <div className="m-8 flex flex-wrap justify-center gap-4 pt-12 dark:bg-slate-900">
```
ryanwolhuter, line 18: same here, also added by mistake

acaldas (approved):

PR #2720, opened 2026-06-15
feat: use shadcn type theming class names

ryanwolhuter:
Replaces raw Tailwind palette classes (bg-gray-_, dark:bg-slate-_, etc.) across the UI packages with shadcn-style semantic tokens, and adds a lint rule to prevent people from re-adding them.

- Defined semantic tokens in `packages/design-system/theme.css` (`:root` + `.dark` + `@theme inline`) that reference the existing palette. Light = gray, dark = slate, so the `dark:` variant twins are gone — one class adapts to both modes.

- Tokens: background, foreground, card, popover, primary, secondary, muted, accent, destructive, info, success, warning (+ their `-foreground`), border, input, ring, sidebar-_, chart-_.

- Status colors are mode-independent and use a single `{50, 500, 900}` ladder (50 = soft, e.g. `bg-info/10`). Solids use white foreground; `warning` is orange (yellow failed contrast with white text).

- Added interaction utilities `hover-effect`, `active-effect`, `disabled-effect` (brightness/opacity via `--hover`/`--active`/`--disabled`) to replace ad-hoc per-color hover/active/disabled classes.

- Migrated component classes across design-system, connect, vetra, powerhouse-vetra-packages, and switchboard-gui; wired card/popover/sidebar components to their tokens.

- `reactor-browser` inline styles use `var(--token, #hex)` so the package still renders standalone without the theme.

- `switchboard-gui` now imports the design-system theme (added as a workspace dep) — it was using tokens without defining them.

- Migrated the `@apply` color classes in `style.css`.

## Lint

- Extended the `better-tailwindcss` config to switchboard-gui and to custom `*ClassName` props (headerClassName, containerClassName, etc.) so they're ordered/checked like `className`.

- Added `better-tailwindcss/no-restricted-classes`: a single regex that errors on any raw palette class and points to the tokens. Matches class names by shade pattern, so it's independent of the theme's actual color values.


package.json
```diff
@@ -34,7 +34,7 @@
     "knip": "knip",
     "postinstall": "pnpm update-ts-references",
     "add-dark-mode": "tsx scripts/dark-mode/main.ts",
-    "count-classes": "tsx scripts/dark-mode/count-classes.ts",
+    "count-classes": "tsx scripts/dark-mode/get-all-classes.ts",
     "add-classes": "tsx scripts/dark-mode/add-classes.ts",
     "replace-classes": "tsx scripts/dark-mode/replace-classes.ts",
     "remove-classes": "tsx scripts/dark-mode/remove-classes.ts",
```
YasielCabrera: If we plan to keep these scripts around for future use, could we add a bit of documentation on when and how to use them? If they were only needed for this migration, it might be better to remove them now to avoid confusion later

packages/switchboard-gui/package.json
```diff
@@ -17,14 +17,16 @@
...
+    "@powerhousedao/design-system": "workspace:*",
     "@apollo/client": "^3.11.8",
     "@heroicons/react": "^2.1.5",
     "ethers": "^6.13.3",
     "jwt-decode": "^4.0.0",
     "preact": "^10.24.1",
     "preact-router": "^4.1.2",
     "tailwindcss": "catalog:",
-    "zustand": "5.0.0-rc.2"
+    "zustand": "5.0.0-rc.2",
```
YasielCabrera, line 28: We’re currently using an `rc` version, but the stable version has already been released. It would be worth upgrading to the stable version

apps/connect/src/components/button.tsx
```diff
@@ -7,7 +7,7 @@ const Button: React.FC<React.ButtonHTMLAttributes<HTMLButtonElement>> = ({
   return (
     <button
       className={twMerge(
```
YasielCabrera, line 9: Now that we're speaking "shadcn dialect", what if we use the classic `cn` function over raw `twMerge` ? For most cases, twMerge is enough but it is better to have `clsx` to filter falsy values and to support objects/cva for complex tw class styling

package.json
ryanwolhuter: These were a one-off, they are deleted now

packages/switchboard-gui/package.json
ryanwolhuter: I didn't change this, it's only in the diff because of some kind of merge artifact. We can absolutely update it but that's not part of this PR

apps/connect/src/components/button.tsx
ryanwolhuter: Sure I have no problem with that, but this PR is already quite large so it would be better to do it separately

packages/reactor-browser/src/renown/components/RenownAuthButton.tsx
```diff
@@ -40,7 +40,7 @@ export function RenownAuthButton({
             gap: "8px",
             padding: "6px 12px",
             borderRadius: "8px",
-            border: "1px solid #e5e7eb",
+            border: "1px solid var(--border)",
```
acaldas: This one is missing a default value. These Renown components are meant to be used outside of Connect too.

packages/reactor-browser/src/renown/components/RenownAuthButton.tsx
```diff
@@ -49,15 +49,15 @@ export function RenownAuthButton({
               width: "28px",
               height: "28px",
               borderRadius: "50%",
-              backgroundColor: "#e5e7eb",
+              backgroundColor: "var(--secondary)",
```
acaldas: Same as above.

packages/reactor-browser/src/renown/components/RenownAuthButton.tsx
```diff
@@ -49,15 +49,15 @@ export function RenownAuthButton({
...
             }}
           />
           <div
             style={{
               width: "80px",
               height: "14px",
               borderRadius: "4px",
-              backgroundColor: "#e5e7eb",
+              backgroundColor: "var(--secondary)",
```
acaldas: Same as above.

acaldas (approved):

packages/reactor-browser/src/renown/components/RenownAuthButton.tsx
ryanwolhuter: weird, I did have the default values but I might have lost them in all the undo redo this PR needed

YasielCabrera (approved):

PR #2747, opened 2026-06-19
fix(connect): gate drive login prompt on genuine auth rejections

acaldas:
## Problem

Connect's drive login prompt (the full-page `DriveAuthGate` and the `driveAuthRequired` modal) fired on **any** drive sync error — a network failure, a 5xx, a "drive not found", or a switchboard being down all surfaced a "Log in to access this drive" prompt, even though logging in wouldn't fix any of them.

## Fix

The login prompt now fires **only for a genuine auth rejection**:

- **`reactor`** — `ConnectionStateSnapshot` carries a `requiresAuth` flag. A shared `isDriveAuthError(error)` + a `DRIVE_AUTH_ERROR_MESSAGES` constant replace **two duplicated message regexes** (previously copy-pasted in `reactor-browser` and `connect`). Detection: HTTP `401`/`403`, or a GraphQL error whose message matches the shared auth-message constants.
- **`reactor-api`** — the auth throws in `base-subgraph.ts` now use dedicated `ForbiddenError` / `AuthenticationRequiredError` classes carrying a stable `extensions.code` (`FORBIDDEN` / `UNAUTHENTICATED`). Messages are unchanged.
- **`connect`** — the gate fires only when an errored remote is an auth rejection (`state === "error" && requiresAuth`); the add-drive modal uses the same shared helper.

## Back-compat

The client matches the **message constant**, not the new code, so detection keeps working against already-deployed switchboards (which emit the message but not the code). The `extensions.code` is additive — a future client can prefer it with the message as fallback.

## How it was verified

- **Unit tests** — `isDriveAuthError` (401/403, Forbidden/auth-required GraphQL, non-auth, network/parse), channel `requiresAuth` transitions (401/403/Forbidden → true; 500/validation → false), the error classes (code + message), and the gate decision.
- **End-to-end (Playwright + real Switchboard + Connect)**:
  - Protected drive, anonymous → login prompt shown; `touchChannel` returns HTTP 200 + `Forbidden` message + `code: FORBIDDEN`.
  - Same drive, auth off (`OPEN`) → no prompt, drive syncs.
  - Mock switchboard returning HTTP 500 / connection-reset on poll → **before** the fix the prompt showed (reproduced the bug), **after** it does not.
- Reactor sync suite (469) + reactor-api error/permission tests pass; eslint + `tsc` clean across `reactor`, `reactor-api`, `reactor-browser`, `connect`.

## Notes

- The `requiresAuth` flag is set at the unrecoverable-error sites and cleared only on a successful connect, so a transient recoverable error can't wipe a real auth flag.
- `DRIVE_AUTH_ERROR_MESSAGES` is a named object (`{ forbidden, authenticationRequired }`) to avoid positional coupling between message and error code.

### DIFF
--- a/apps/connect/src/components/use-drive-auth-gate.test.ts
+++ b/apps/connect/src/components/use-drive-auth-gate.test.ts
@@ -2,9 +2,10 @@ import { describe, expect, it } from "vitest";
 import type { ConnectionStateSnapshot } from "@powerhousedao/reactor-browser";
 import { computeNeedsLogin } from "./use-drive-auth-gate.js";
 
-// Only `state` drives the decision; the rest are filled with neutral values.
+// `state` + `requiresAuth` drive the decision; the rest are neutral.
 function snap(
   state: ConnectionStateSnapshot["state"],
+  requiresAuth = false,
 ): ConnectionStateSnapshot {
   return {
     state,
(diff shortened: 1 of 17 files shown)


gpuente (approved):
LGTM

PR #2887, opened 2026-07-27
fix(connect): re-import a deleted document as a copy

liberuum:
Deleting a document in Connect hides it but keeps its history, so its id stays taken. Re-importing the same file reuses that id and fails with "Revision mismatch: expected 3, got 0".

Since a deleted id is indistinguishable from an unused one, the importer cannot check up front. It now catches the reactor's rejection, generates a new id, and imports the file again under it, so the document comes back as a copy with its contents intact. Both import paths behave this way, and unrelated failures are still surfaced as errors.

### DIFF
--- a/packages/reactor-browser/src/actions/document.ts
+++ b/packages/reactor-browser/src/actions/document.ts
@@ -108,6 +108,56 @@ async function isDocumentInLocation(
   return { isDuplicate: false };
 }
 
+/**
+ * Recognizes the failure raised when CREATE_DOCUMENT targets a document id the
+ * reactor still owns.
+ *
+ * Deleting a document appends a tombstone but retains its operation history, so
+ * the id stays owned. No read path reveals this before the write: the document
+ * view, `resolveIdOrSlug` and `getOperations` all hide soft-deleted documents,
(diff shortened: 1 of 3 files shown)


thegoldenmule (changes requested):
I think we need to approach this problem differently (and I'm happy to take it if you'd like!):

1. Provide an explicit error when a document already exists, we can't piggyback on mismatch.
2. Provide the option to return deleted documents in the `ViewFilter`.
3. Simply check for existence on import and dupe if needed. That is, keep it in the import level, not deeper down.
4. The way it's duping here is not going to work once auth is in due to how signing works, but... maybe we cross that bridge when we get to it.

thegoldenmule:
Superseded by main (2026-08-28). Closing.

1. Explicit error: a taken id now fails with `DocumentAlreadyExistsError` (be6db7b118, b64079a42d).
2. Deleted docs: added `IReactorClient.isDocumentIdTaken` instead of a `ViewFilter` option (2b92450caf).
3. Import-level check: `createImportedDocument` checks the id first and copies under a new id if it's taken, retrying on a create race (f324f42ad3).
4. Signing under auth is still open and applies to the new path.

PR #2901, opened 2026-08-06
Fix connect document upgrades

liberuum:
### Problem
Documents created with a newer document-model version could not be opened or
upgraded in a Connect running an older package, and nothing in the platform
let publishers release or manage model versions.

### What's included
**Upgrade engine (reactor/shared)** — version stamping on create/import,
versioned replay segmentation, UPGRADE_DOCUMENT execution with cross-scope
state handling, write-cache boundary handling, `UnsupportedDocumentModelVersionError`.

**Connect UX** — warning toast + toolbar "Update available" button on outdated
documents; confirmation modal with a dry-run field diff (added/removed/changed);
"version not supported" card instead of a crash when a document is newer than
the installed package; version-aware create flow.

**Document model editor (Vetra)** — version badge, read-only version switcher,
"Release new version" button with confirmation, and a plain-language advisory
that intercepts breaking edits (schema shape changes, deletions, renames,
operation-input changes) offering to release a new version first. Fixed the
editor editing `specifications[0]` while reducers write to the last spec.

**Codegen** — version-aware generated validators: documents are validated
against the schema of the version they are stamped with, so a new version may
add non-nullable fields without breaking not-yet-upgraded documents (runtime
regression test included).

**Create dialog** — outside studio mode only the latest version of each type
is offered (builders keep the full per-version list in studio mode).

**Fixtures/infra** — `test/versioned-documents` todo v1/v2 fixture with real
editor + migration; vetra-e2e now exports `upgradeManifests` (upgrade
notifications were silently disabled for local project packages);
powerhouse-vetra-packages tests wired into `test:ci`.

**Document Editor Toggle, Toolbar and Modal prompt**
connect-doc-update-promp

**Document Model Editor version display, selector and new release button**
doc-model-editor-versions-display

**Modal prompt**
nodal-prompt-after-change


froid1911 (approved):

thegoldenmule (changes requested):
Hmm, a claude review found a number of issues in the core reactor changes. Please hold off on merging this until I can review them and work on it on Monday. Posting here for an artifact (again -- I haven't reviewed these all yet but there are enough issues that I'd like to hold off on merging):

Rebuild/replay divergence (the serious cluster):

1. packages/reactor/src/read-models/document-view.ts:151 — UPGRADE_DOCUMENT indexing now overwrites every scope's snapshot unconditionally. Replaying upgrade operations persisted by the pre-PR executor (whose resultingState carries stale sibling-scope echoes) clobbers good snapshots — the read model serves reverted content.
2. packages/reactor/src/client/reactor-client.ts:629 — upgradeDocument records fromVersion = version || 1, but the write cache resolves version 0 to the latest module. Upgrading a legacy version-0 document retroactively re-pins all its pre-upgrade operations to the v1 reducer on rebuild — rebuilt state and hashes diverge from live execution.
3. packages/reactor/src/client/reactor-client.ts:653 — the upgrade's revision boundary is snapshotted client-side and never re-derived by the executor. An edit that lands between the client's read and the upgrade job executing gets misclassified as post-upgrade on rebuild, replaying with the wrong reducer version.
5. packages/reactor/src/client/drive-client.ts:126 — addFile now dereferences document.state.document.version (previously hardcoded toVersion: 1), and nothing backfills the document scope. Importing a legacy .phd file that predates the document scope throws a TypeError instead of importing.
6. packages/codegen/src/templates/document-model/gen/document-schema.ts:74 — generated stampedDocumentModelVersion passes version 0 through unchanged, so SchemasByVersion[0] is undefined and legacy documents fall back to the latest schema. Once v2 adds a required field, isXDocument fails / assertIsXDocument throws for exactly the documents this feature is meant to protect. Everywhere else in the PR, version 0 means "treat as v1."
7. packages/reactor/src/cache/kysely-write-cache.ts:746 — held-back pending upgrades invert document-scope ordering: a DELETE_DOCUMENT is applied inline while a preceding upgrade is deferred, and in the manifest-less fallback the upgrade's initialState then applies after the delete — a rebuilt deleted document comes back as live.
8. packages/reactor/src/client/reactor-client.ts:138 — getDocumentModelModule resolves latest-wins, but version-blind consumers like signAction (reactor-browser actions/sign.ts:18) feed that reducer to buildSignedAction regardless of the document's stamped version. Signed operations on not-yet-upgraded documents are built with the wrong reducer, diverging from replay.
9. packages/reactor/src/executor/document-action-handler.ts:740 — executeUpgrade invalidates the write cache for sibling scopes before the upgrade operation commits, with no re-invalidation after. A concurrent read in that window repopulates the cache with pre-upgrade state that then sticks.
Connect UI:

9. packages/reactor-browser/src/utils/upgrade-preview.ts:135 — third-party upgrade reducers run unguarded during React render (unmemoized, after a full structuredClone). A throwing migration reducer silently kills the confirm modal; large documents pay clone-plus-migrate every render.
10. apps/connect/src/components/editors.tsx:103 — the upgrade toast effect has no cleanup and only dismisses the toast for the currently open documentId, so an autoClose:false toast for document A persists forever after navigating to B.
11. apps/connect/src/components/modal/modals/ConfirmDocumentUpgradeModal.tsx:53 — the modal returns null when the document or preview is missing while phModal stays set to confirmDocumentUpgrade, leaving global modal state stuck with nothing visible to dismiss.

Other:

12. packages/reactor/src/executor/document-action-handler.ts:614 — executeUpgrade sequentially fetches every sibling scope before the no-op check, and runs for the seed 0→N upgrade in every create/addFile batch — pure overhead, plus a new failure mode where a transient sibling-scope error fails the whole creation.
13. packages/reactor/src/cache/kysely-write-cache.ts:921 — positional reads (targetRevision below the upgrade boundary) still apply the pending upgrade at the tail, returning migrated v2 state labeled with a pre-upgrade revision — which can be cached as a keyframe and poison later rebuilds.
14. packages/powerhouse-vetra-packages/vitest.config.ts:5 — the new config narrows include to editors/**/*.test.ts, silently orphaning the seven pre-existing document-model test suites (reducer, versioning, validation, zip, prune, skip-operations ×2) while the PR wires the package into root test:ci — CI reports green while running zero of those tests.
15. packages/powerhouse-vetra-packages/editors/document-model-editor/hooks/useVersionAdvisory.ts:69 — guardedDispatch unconditionally overwrites pendingRef.current, so a second version-relevant dispatch while the advisory prompt is open (e.g. a blur-committed edit as the modal steals focus) silently discards the first held-back edit.

PR #2916, opened 2026-08-19
feat(reactor-browser): sign action batches in the GraphQL light client

gpuente:
## Summary

The GraphQL light client signed only a single action. With an ambient signer and a batch it warned and pushed the whole batch **unsigned**. It now signs every action of a same-scope batch and sends them in the one existing mutation — no schema, resolver, or executor change.

The light read path fetches state and per-scope revisions but **no operation history**, so the chain is predicted client-side: each action is stamped at a virtual revision, signed, then reduced with the document's own reducer to get the state the next signature hashes. The virtual revision is tracked separately because the reducer derives operation indices from the last stored operation, which counts from zero on a history-free snapshot.

## Changes

- `static-package-manager.ts` — new pure `resolveDocumentModelModule(modules, type, version?)`: exact match when a version is given, latest when not. `StaticPackageManager.load` delegates to it.
- `signing.ts` — new `prepareSignedActions`; `stampAction` takes an optional explicit revision, `signStampedAction` takes an `AbortSignal`.
- `graphql-reactor-client.ts` — new `documentModels` and `signer` options; the module is resolved with `normalizeDocumentModelVersion`, the same rule `SimpleJobExecutor` applies, so client and server cannot pick different reducers.
- `graphql-reactor-provider.tsx` — its existing `documentModels` prop now also reaches the client constructor.

A batch that cannot be predicted — no matching module version, mixed scopes, or an action needing history (`UNDO`/`REDO`/`PRUNE`/`NOOP`, document-scope actions) — fails before the mutation instead of downgrading to unsigned. Single actions still sign without any models, and unsigned pushes are unchanged.

### DIFF
--- a/packages/reactor-browser/src/graphql-client/graphql-reactor-client.ts
+++ b/packages/reactor-browser/src/graphql-client/graphql-reactor-client.ts
@@ -8,8 +8,17 @@ import type {
   SearchFilter,
   ViewFilter,
 } from "@powerhousedao/reactor";
-import type { ISigner } from "@powerhousedao/shared/document-model";
-import type { Action, Operation, PHDocument } from "document-model";
+import type {
+  ISigner,
+  PHDocumentState,
+} from "@powerhousedao/shared/document-model";
+import { normalizeDocumentModelVersion } from "@powerhousedao/shared/document-model";
+import type {
+  Action,
+  DocumentModelModule,
+  Operation,
+  PHDocument,
+} from "document-model";
 import { logger } from "document-model";
 import type { Variables } from "graphql-request";
 import { createClient } from "../graphql/client.js";
@@ -38,7 +47,8 @@ import {
   type MutateDocumentWithOperationsResult,
   type MutateDocumentWithOperationsVariables,
 } from "./operations.js";
-import { signStampedAction, stampAction } from "./signing.js";
+import { prepareSignedActions } from "./signing.js";
+import { resolveDocumentModelModule } from "./static-package-manager.js";
 import {
   describeGraphQLDocument,
   SubgraphSdkRegistry,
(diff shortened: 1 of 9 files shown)


gpuente:
Re: the CI duration warning — this is not caused by this branch. The whole regression sits in one step.

Per-step timings for `Run Package E2E Tests`, this run vs the three most recent green `main` runs:

| Step | Baseline (3 runs) | This run |
| --- | --- | --- |
| Build | 1m48s / 1m52s / 1m49s | 1m50s |
| Install Playwright Browsers | 54s / 26s / 22s | **24m18s** |
| Run package E2E flow | 6m29s / 6m48s / 6m33s | 6m39s |

The job went 10m46s → 34m43s, a delta of 23m57s; the install step alone accounts for 23m48s of it. The tests themselves are normal — 6m39s is mid-range against the baseline — and the build is identical to the second.

`pnpm exec playwright install --with-deps chromium` spent that time in `apt-get`, pulling font packages from `azure.archive.ubuntu.com`:

| Package | Size | Time |
| --- | --- | --- |
| fonts-wqy-zenhei | 7.5 MB | 8m55s |
| fonts-freefont-ttf | 5.6 MB | 5m45s |
| fonts-unifont | 3.0 MB | 3m54s |
| fonts-ipafont-gothic | 3.5 MB | 3m30s |

~20 MB in ~22 minutes, about 15 KB/s. The apt metadata fetch earlier in the same step managed 11.4 MB in 26s (432 kB/s), so the mirror degraded partway through. A re-run should land back around 11m.

One structural note while the evidence is here: the Playwright cache hit (`Cache hit occurred on the primary key playwright-chromium-Linux-1.60.0`), but `--with-deps` runs apt unconditionally regardless. The cache covers the browser binary, not the system libraries, so this job is exposed to apt-mirror weather on every run. Worth a separate look if it recurs.

This branch changes no dependency, no Playwright config, and nothing under `test/package-e2e`.

gpuente:
> ### ⚠️ CI duration regression
> Compared with the median of the last green `main` runs of `e2e-tests.yml`:
> 
> Job	This run	Baseline median	Δ
> Run Package E2E Tests	34m43s	10m46s	+222%
> This is a warning, not a failure — but a slow creep here is how jobs walk into their timeouts. If the slowdown is expected, say so in the PR; if not, it is worth a look before merging.

The CI duration warning, this is not caused by this branch. The whole regression sits in one step.

Per-step timings for `Run Package E2E Tests`, this run vs the three most recent green `main` runs:

| Step | Baseline (3 runs) | This run |
| --- | --- | --- |
| Build | 1m48s / 1m52s / 1m49s | 1m50s |
| Install Playwright Browsers | 54s / 26s / 22s | **24m18s** |
| Run package E2E flow | 6m29s / 6m48s / 6m33s | 6m39s |

The job went 10m46s → 34m43s, a delta of 23m57s; the install step alone accounts for 23m48s of it. The tests themselves are normal — 6m39s is mid-range against the baseline — and the build is identical to the second.

`pnpm exec playwright install --with-deps chromium` spent that time in `apt-get`, pulling font packages from `azure.archive.ubuntu.com`:

| Package | Size | Time |
| --- | --- | --- |
| fonts-wqy-zenhei | 7.5 MB | 8m55s |
| fonts-freefont-ttf | 5.6 MB | 5m45s |
| fonts-unifont | 3.0 MB | 3m54s |
| fonts-ipafont-gothic | 3.5 MB | 3m30s |

~20 MB in ~22 minutes, about 15 KB/s. The apt metadata fetch earlier in the same step managed 11.4 MB in 26s (432 kB/s), so the mirror degraded partway through. A re-run should land back around 11m.

One structural note while the evidence is here: the Playwright cache hit (`Cache hit occurred on the primary key playwright-chromium-Linux-1.60.0`), but `--with-deps` runs apt unconditionally regardless. The cache covers the browser binary, not the system libraries, so this job is exposed to apt-mirror weather on every run. Worth a separate look if it recurs.

This branch changes no dependency, no Playwright config, and nothing under `test/package-e2e`.

PR #2922, opened 2026-08-20
Revert the updateOutbox scope-ordering repro off main

thegoldenmule:
`Test Reactor` is red on `main`, and this is the whole of it: two deliberately
failing test cases.

`aee8e6d09` is a **repro** commit. Its own message says two of its three cases
"fail on main, deliberately — they assert the causally correct behaviour, not the
current one". It landed on main by accident on the back of #2920, and main has
been red since.

A repro belongs on the branch that fixes what it reproduces, and it already is
there: **#2921** carries this same file alongside the fix to `updateOutbox`'s
page ordering. Removing it here loses nothing — #2921 reintroduces it green.

**My fault it was here.** #2920 was branched from a local `main` that carried
this commit while `origin/main` did not, and I pushed without checking the branch
point, so that PR contained two commits instead of one.

Verified: reactor suite 2666/2666 with it out.

**Why not just merge #2921 instead:** its `Test Reactor` passes, but `Run Package
E2E Tests` and `Run Vetra E2E Tests (SharedWorker reactor)` fail on it, and both
of those *pass* on main. Merging it as-is would trade one red job for two. This
revert gets main green now and leaves #2921 free to land on its own merits.


thegoldenmule:
Superseded — skipped the two cases on main directly (78e18948d) instead of removing the file, so the repro stays readable and #2921 just un-skips it.

PR #2925, opened 2026-08-24
Auth-scope cost matrix: micro and meso tiers

thegoldenmule:
Measures what the auth enforcement flags cost, decomposed so the number is attributable. Adds no new benchmark harness: the repo already had three, each with an unused A/B mechanism, and this puts the flag axis on two of them.

No production code paths change. `reactor-direct.ts` defaults to `L0_CLEAN` with no policy, so existing runs are unaffected.

---

## 1. Most of what "turning auth on" costs is not authorization

ladder

+31% end to end on Postgres 16, and **+23 of those points are `documentDecisions`** — the prerequisite. `authEnforcement`, the flag that actually evaluates the policy, adds 6%. Both steps survive pairing: all five interleaved repetitions positive.

An auth-off/auth-on comparison yields one number and hands all of it to authorization. Three quarters belongs to the admission mechanism underneath — different work, different owner, different fix.

PGlite and Postgres agree to within a thousandth (1.232×/1.232×, 1.060×/1.059×). **That agreement is the finding's limit, not reassurance:** this workload keeps one write in flight, so the advisory lock is never contended and both engines skip the same path.

## 2. On the CPU, cost is set by policy shape — not by which flag is on

drivers

- **Grant ordering spans 905×.** Retention on an auth-scope write is 0.49 µs with the administration grant first, **445 µs with it last**, same 100 grants, both installable. One-line policy change, three orders of magnitude.
- **Conditions cost 1622×.** Worst legal policy (100 grants × 100-node `where`) is **206 µs/decision** against 0.13 µs for the same policy below the flag.
- **No early exit.** Moving the deciding grant to the front changes cost by +0.8% — nothing. Last-applicable-grant-wins forbids stopping; an early-exit implementation would have been ~50× faster there.
- **Rosters allocate per member.** 1000 members costs 34× ten, ~9.2 ns each. The *cheapest* outcome is a group the map doesn't hold — which is why a harness missing the group model reports groups as nearly free.

The first two are policy-authoring choices, invisible from reading the source, and not tunable from outside.

## 3. The read path is the one on the saturated loop

gates

Write-path auth lands in the executor's worker pool. Read-path auth runs on the `ReactorClient` in the host process — the loop `BASELINE.md` Run 10 found pinned at 99.3%.

A policied read is 3.11 µs. Reading a roster **100 documents reference is 209 µs**, because deciding whether to serve it rebuilds each referencing document's decision model. The gate runs **once per result**, so a 50-result listing is ~10 ms of loop time, with no cache on that path.

The uninitialized fast path (0.18 µs, 17× cheaper) is also the measurement trap: forget `INITIALIZE_AUTH` and you measure it and conclude the gate is free.

## 4. Inside the prerequisite, it's a round trip

attribution

The flag is named `documentDecisions`, so the natural reading is that deciding is expensive. **Deciding is 0.15%.**

| component | ms | share |
|---|---:|---:|
| Postgres executing SQL | 171 | **8.8%** |
| round trips, +1 statement per operation | ~600 | **~31%** |
| building + serialising the larger statement | ~1160 | **~60%** |

**The number that mattered was a count.** L1 issues **+5135 statements for 5000 operations** — one per operation — because `acquireStreamLocks` must be separate from the guarded insert: the locks have to be held *before* the insert takes its snapshot. That lock costs **2 µs to execute** across 5006 acquisitions and a **full round trip to issue**. Measured round trip here: 0.120 ms.

All of it is multiplied by ~100, because `apply` runs per operation (909 inserts for 900 operations). Batching a hundred actions into one `execute` does not batch the storage writes underneath.

**This ranks two fixes that were tied.** Batching applies attacks all three rows; preparing the statement attacks part of one. Batch first — and it's falsifiable: statement count should fall towards one per batch and the wall delta should collapse with it. If the count falls and the delta doesn't, the ranking is wrong.

---

## What's here

| | |
|---|---|
| `bench/fixtures/auth-policies.ts` | the six-rung ladder and policy builders, imported by every tier so the numbers relate |
| `test/bench/auth-policies.test.ts` | 36 assertions that the fixtures install and reach the code under test |
| `bench/auth-scope.bench.ts` | 33 micro cells; `bench:auth` / `:record` / `:compare` |
| `scripts/profiling/reactor-direct.ts` | flag axis, policy seeding with read-back, signer context |
| `scripts/profiling/pg-statement-diff.ts` | attributes a wall delta to SQL, or rules SQL out |
| `bench/AUTH-FINDINGS.md` | the go/no-go read, charts embedded |
| `bench/test/integration/AUTH-MATRIX.md` | the run log, in `BASELINE.md`'s shape |

The ladder is cumulative because `validateFeatureFlags` throws on a skipped prerequisite — 11 of 16 flag subsets are unconstructible. Its bottom rung splits in two: below `authEnforcement` the executor still scans grants in memory, so a policied L0 against L2 separates auth's reads from the scan they share.

## Measurement traps, kept in the record

Six, written up rather than quietly fixed, because each produced or would produce a confident wrong answer.

| trap | what it did |
|---|---|
| uninitialized policy | fast-paths to allow; measures auth as free |
| anonymous subject | `{address}`/`{group}` grants can never match |
| unregistered group model | group principals fail closed — the cheapest path, dressed as the dearest |
| 10 ms poll quantum | hid every effect at batch size 1; the ladder read flat |
| blocked run ordering | manufactured a confident +5% for L3 that interleaving erased |
| statement logging | taxes the busier arm 1.84× and inflates its own numerator: gave 20–54% where `pg_stat_statements` gives 8.8% |

The last one is the one to dwell on — two instruments disagreed **by a factor of two, and the cheaper-looking one was wrong.** The fixture guard exists for the same reason, and caught three real bugs in the condition-tree builder before they became numbers.

## What this does not claim

- **Concurrency is untested.** One write in flight throughout, so the advisory lock is never contended and the retry-exempt append conflicts never fire. Largest remaining gap, and a driver limitation rather than a storage one. Next probe: N writers over documents naming one shared group, against the same documents naming disjoint groups.
- **No backdated arm**, so `foldEvaluatedScope` — the most expensive construct in the feature — never runs. `authConditions` keeps measuring as a tie until it does.
- **Macro tier untouched.** The bench host sets no flags, installs no policy, registers no group model, stubs verification to `() => true`, and builds its signer with no user, so its subject address is `""`.
- **+31% is not a production number.** Four reasons it's an underestimate are named in `AUTH-FINDINGS.md`.

## Verification

```sh
pnpm --filter @powerhousedao/reactor exec vitest run test/bench/auth-policies.test.ts   # 36 pass
pnpm --filter @powerhousedao/reactor bench:auth
pnpm --filter @powerhousedao/reactor lint     # 0 errors (54 pre-existing warnings)
pnpm --filter @powerhousedao/reactor tsc --build
```

Charts are generated from the recorded JSON by `plot-auth-matrix.py` rather than retyped, which caught two prose numbers that had already drifted.


thegoldenmule:
## Findings, one at a time

Eight runs. Two writeups: [`AUTH-FINDINGS.md`](packages/reactor/bench/AUTH-FINDINGS.md) is the go/no-go read, [`AUTH-MATRIX.md`](packages/reactor/bench/test/integration/AUTH-MATRIX.md) is the append-only log. Every chart is generated from the recorded JSON by `plot-auth-matrix.py`, never retyped — which is how two drifted prose numbers got caught.

---

### 1. Most of what "turning auth on" costs is not authorization

ladder

+31% end to end on Postgres 16, and **+23 of those points are `documentDecisions`** — the prerequisite. `authEnforcement`, the flag that evaluates the policy, adds 6%. Both steps survive pairing: every one of five interleaved repetitions is positive.

A single auth-off/auth-on number hands all 31% to authorization. Three quarters of it belongs to the admission mechanism underneath, which is different work with a different fix.

The third panel is the cross-engine check: PGlite and Postgres agree to within a thousandth (1.232×/1.232×, 1.060×/1.059×). **That agreement is the finding's limit, not reassurance** — this workload keeps one write in flight, so the advisory lock is never contended.

---

### 2. On the CPU, cost is set by policy shape — not by which flag is on

drivers

Four drivers, all properties of the policy someone writes rather than of the code:

- **The grant stack has no early exit.** Moving the deciding grant to the front changes cost by +0.8%, i.e. nothing. Last-applicable-grant-wins forbids stopping; an early-exit implementation would have been ~50× faster on that cell.
- **Group membership allocates per member.** A 1000-member roster costs 34× a ten-member one, ~9.2 ns/member. The *cheapest* outcome is a group the map doesn't hold — which is exactly why a harness missing the group model reports groups as nearly free.
- **Conditions are 1622×.** The worst legal policy (100 grants × 100-node `where`) is **206 µs/decision** against 0.13 µs for the identical policy below the flag.
- **Retention is 905× on ordering alone.** 0.49 µs with the administration grant first, **445 µs with it last**, same 100 grants, both installable.

The last two are actionable and neither is tunable from outside. Putting the administration grant first is a one-line policy change worth three orders of magnitude.

---

### 3. The read path is the one on the saturated loop

gates

Write-path auth lands in the executor's worker pool. Read-path auth runs on the `ReactorClient` in the host process — the loop `BASELINE.md` Run 10 found pinned at 99.3%.

An ordinary policied read is 3.11 µs. Reading a roster **100 documents reference is 209 µs**, because deciding whether to serve it rebuilds each referencing document's decision model. The gate runs **once per result**, so a 50-result listing is ~10 ms of loop time, with no cache anywhere on that path.

The uninitialized fast path (0.18 µs, 17× cheaper) is also the measurement trap: forget `INITIALIZE_AUTH` and you measure that and conclude the gate is free.

---

### 4. Inside the prerequisite, it's a round trip — not the decision, and mostly not the database

attribution

The flag is named `documentDecisions`, so the obvious reading is that deciding is expensive. It is 0.15%.

| component | ms | share |
|---|---:|---:|
| Postgres executing SQL | 171 | **8.8%** |
| round trips, +1 statement per operation | ~600 | **~31%** |
| building + serialising the larger statement | ~1160 | **~60%** |

**The number that mattered was a count.** L1 issues **+5135 statements for 5000 operations** — one per operation — because `acquireStreamLocks` must be separate from the guarded insert: the locks have to be held *before* the insert takes its snapshot. That lock costs **2 µs to execute** across 5006 acquisitions and a **full round trip to issue**. Measured round trip against this Postgres: 0.120 ms.

All of it is multiplied by ~100, because `apply` runs per operation — 909 inserts for 900 operations. Batching a hundred actions into one `execute` does not batch the storage writes underneath.

**This ranks two fixes that were previously tied.** Batching applies attacks all three rows at once; preparing the statement attacks part of one. Batch first — and the prediction is falsifiable: statement count should fall towards one per batch and the wall delta should collapse with it. If the count falls and the delta doesn't, the ranking is wrong.

---

### Measurement traps, kept in the record

Six, all written up rather than quietly fixed, because each produced or would produce a confident wrong answer:

| trap | what it did |
|---|---|
| uninitialized policy | fast-paths to allow; measures auth as free |
| anonymous subject | `{address}`/`{group}` grants can never match |
| unregistered group model | group principals fail closed — the *cheapest* path, dressed as the dearest |
| 10 ms poll quantum | hid every effect at batch size 1; the ladder read flat |
| blocked run ordering | manufactured a confident +5% for L3 that interleaving erased |
| statement logging | taxes the busier arm 1.84× and inflates its own numerator — gave 20–54% where `pg_stat_statements` gives 8.8% |

The last one is worth dwelling on: the first instrument disagreed with the second **by a factor of two, and the cheap-looking one was wrong**. The fixture guard (36 assertions) exists for the same reason, and caught three real bugs in the condition-tree builder before they became numbers.

---

### What this does not claim

- **Concurrency is untested.** One write in flight throughout, so the advisory lock is never contended and the retry-exempt append conflicts never fire. This is the largest remaining gap, and it is a driver limitation rather than a storage one.
- **No backdated arm**, so `foldEvaluatedScope` — the most expensive construct in the feature — never runs. `authConditions` will keep measuring as a tie until it does.
- **Macro tier untouched.** The bench host sets no flags, installs no policy, registers no group model, stubs verification to `() => true`, and builds its signer with no user.
- **+31% is not a production number.** Four reasons it is an underestimate are named in `AUTH-FINDINGS.md`.

No production code paths change in this PR. `reactor-direct.ts` defaults to `L0_CLEAN` with no policy, so existing runs are unaffected.

PR #2927, opened 2026-08-25
test(reactor): pin processor delivery under concurrent read-model batches

acaldas:
Two `it.fails` tests documenting a delivery bug in `ProcessorManager`; they flip to passing once it is fixed.

`ReadModelCoordinator` serializes projection per `documentId:scope:branch` and runs other keys in parallel, so a drive's creation (`document` scope) and its first edit (`global` scope) reach `ProcessorManager.indexOperations` concurrently. The edit's batch runs before the creation's batch has registered the drive's processors, so the edit is routed to nobody and is never delivered live. Cursors are then written with the batch's max ordinal unconditionally, so both the processor cursor and the read model's `ViewState` regress to the older batch.

- Integration: `create` then `execute(setDriveName)` back-to-back; both jobs reach `READ_READY` but the processor never receives `SET_DRIVE_NAME`.
- Standalone: two overlapping `indexOperations` calls; the processor cursor and `ViewState` end at 2 instead of 3.

### DIFF
--- a/packages/reactor/test/processors/processor-manager.test.ts
+++ b/packages/reactor/test/processors/processor-manager.test.ts
@@ -1,5 +1,8 @@
 import { PGlite } from "@electric-sql/pglite";
-import { driveDocumentModelModule } from "@powerhousedao/shared/document-drive";
+import {
+  driveDocumentModelModule,
+  setDriveName,
+} from "@powerhousedao/shared/document-drive";
 import {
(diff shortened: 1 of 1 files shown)


thegoldenmule:
This has some issues, working on it in a branch.

thegoldenmule:
I have a comprehensive attack plan for this.

